PR #565, opened 2020-07-24
[EGD-3307] Developed Calendar database

tomaszrogala:
-Implement calendar events database interface
-create events table
-create database connection on service level
-create basic unit tests for events table
-implement send data and data receive methods on application level


p-sims:
This tread appears to imply that Calendar does/should support events being added and viewed, rather than just displaying the days of the week. The below photo is from v1.2.1:
IMG_1352
I've tried pressing all the buttons and there is no way to add events or look at a particular day. Is this known deficit that will be rectified in due course, or bug I should raise or ?

Is the plan to add calander syncing support or just to allow a stand alone calander?

PR #1770, opened 2021-03-12
[EGD-3646] Add starting and switching back apps via harness

asleszyn:


pholat:

PR #1786, opened 2021-03-17
Calculator crash when comma is decimal separator.

kolegs:
Removed replacing full stop to comma when comma is decimal separator.
Since text field is empty when starting the application it is not possible to have full stop inside it, so replacing is not required. 
Anyway it is crashing the system since find return string::npos and passes it to replace function.

We can use while to find all full stops and check what find returns or just simply remove this section.

Tested on ubuntu 20.04.


GravisZro:
This pull request is obsolete and can be safely closed.

PR #2019, opened 2021-04-28
[EGD-7257] Czech localization 

lukascalda:
Hi, I have just finished 1st version of czech translation. Some lines are not translated yet (189-226), some will probably need review once I get the emulator running (need to see the translation in context.)


lukascalda (commented):
Review done.

PR #2146, opened 2021-05-26
Add Finnish localization

hihaiu:
Finnish localization tested on PurePhone Linux build. All menus and their sub menus have been checked that all the text will fit the context and on the screen properly.

One text couldn't fit to the menus "app_settings_network_operator_auto_select". This really can't be shortened. I don't know how an right to left scrolling animation for the text would work on eInk display. User might though understand what this text is about even if the isn't showing fully.
Screenshot from 2021-05-27 _00-17-20_
_There should read "Operaattorin automaattinen valinta"_

Also word wrapping was not correct in PhoneModes descriptions. I didn't correct it with whitespace etc., because I saw there was possibility to change the font size and my fixes would not work then. I think there should be better dynamic solution for this. You might not understand the text, but there are few words that are cut to next line:
Screenshot from 2021-05-27 00-17-43

There were some missing translations in the English.json, but I will list those to the issue board.


hihaiu:
> Is there anything I could do to help this PR get accepted?
I believe we need someone to review the Finnish localizations.

PR #2187, opened 2021-06-03
[EGD-6746] Cellular static API

wsovalle:
API for storing and quick access to Service's runtime state


pholat:

PR #2575, opened 2021-08-09
[BH-728] Product staging overhaul

asleszyn:
* move product file staging to a single place
* allow products do specify assets


pholat:
outdated

PR #2985, opened 2021-10-14
[BH-728] Define assets per product support

asleszyn:
We want to be able to define different sets of assets for each product.


pholat:
outdated

PR #3079, opened 2021-10-26
[BH-1092] Snooze Chime Playback

tomaszrybarski-digica:
Snooze Chime Sound is played only once (without loop)


mjan-dev:
Closing as it was already fixed properly when adding new playback types.

PR #3182, opened 2021-11-08
[EGD-2489] Enable full newlib

alekrudnik:
Newlib nano is an optimized for size version of newlib libc library.
While newlib nano is smaller in size, it is also noticeably slower than
the full version - https://interrupt.memfault.com/blog/memcpy-newlib-nano

Newlib full is more consistent w c11 stanard, including
* full newlib provides native possiblity to printf 64-variables,
including double and float.
* calling destructors of global objectes ate system exit
* probaly many more...

There was hard fault while trying to enable full newlib whithout
recompilling memcpy implementation with -mno-unaligned-access compile
option.
It was failing for missalinged memory access during memcpy.
Similar issue is described here -
https://bugs.launchpad.net/gcc-arm-embedded/+bug/1857469

### DIFF
--- a/board/rt1051/ldscripts/libs.ld
+++ b/board/rt1051/ldscripts/libs.ld
@@ -10,7 +10,7 @@
 
 GROUP (
   "libgcc.a"
-  "libc_nano.a"
+  "libc.a"
   "libstdc++.a"
   "libm.a"
   "libnosys.a"
--- a/board/rt1051/memcpy-armv7m.S
+++ b/board/rt1051/memcpy-armv7m.S
@@ -1,0 +1,329 @@
+/*
+ * Copyright (c) 2013 ARM Ltd
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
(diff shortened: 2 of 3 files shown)


SP2FET (approved):

mpsm (approved):

lucckb (approved):

niepiekm (approved):

alekrudnik:
Please do not merge until confirmed without approval !

board/rt1051/ldscripts/libs.ld
```diff
@@ -10,7 +10,7 @@
 
 GROUP (
   "libgcc.a"
-  "libc_nano.a"
+  "libc.a"
```
alekrudnik, line 13: comment just to block auto merge

board/rt1051/memcpy-armv7m.S
```diff
@@ -0,0 +1,329 @@
+/*
+ * Copyright (c) 2013 ARM Ltd
```
niepiekm, line 2: Instead of this assembly version, I'd like to suggest using a plain C version from here: 
https://github.com/travisg/newos/blob/master/lib/libc/string/memcpy.c
It's also used in [Android](https://android.googlesource.com/kernel/lk/+/master/lib/libc/string/memcpy.c).
It's short, readable and optimized for handling aligned and unaligned addresses.
And it compiles to [fewer instructions](https://gcc.godbolt.org/z/n4r1jdc46) than this one.

PR #3209, opened 2021-11-10
[BH-1164[ Remember value for Chime Interval

tomaszrybarski-digica:
Display last setting of Meditation Timer - Chime Interval

### DIFF
--- a/products/BellHybrid/apps/application-bell-meditation-timer/presenter/IntervalChimePresenter.cpp
+++ b/products/BellHybrid/apps/application-bell-meditation-timer/presenter/IntervalChimePresenter.cpp
@@ -29,12 +29,13 @@ namespace app::meditation
 
     std::string IntervalChimePresenter::getCurrentInterval() const
     {
-        // Always set interval as none
+        const auto value = settings->getValue(intervalDBRecordName, settings::SettingsScope::AppLocal);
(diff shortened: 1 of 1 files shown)


products/BellHybrid/apps/application-bell-meditation-timer/presenter/IntervalChimePresenter.cpp
```diff
@@ -27,12 +27,13 @@ namespace app::meditation
...
     {
-        // Always set interval as none
+        const auto value = settings->getValue(intervalDBRecordName, settings::SettingsScope::AppLocal);
         for (auto const &option : intervalOptions) {
-            if (option.first == std::chrono::minutes{0}) {
+            if (utils::to_string(option.first.count()) == value) {
                 return option.second;
             }
         }
+        // If not found, return "None"
```
swichu91, line 36: How about using `std::find_if`? It is more idiomatic:)

tomaszrybarski-digica: I would leave it as it is for now as the majority of functions in IntervalChimePresenter class use ranged for loops anyway :)

swichu91 (approved):

mjan-dev (approved):

PR #3211, opened 2021-11-10
[BH-1167] USB async receive

wojtekrzepecki:
Polling routine replaced with event driven implementation for USB receive

### DIFF
--- a/module-bsp/board/linux/usb_cdc/usb_cdc.cpp
+++ b/module-bsp/board/linux/usb_cdc/usb_cdc.cpp
@@ -220,4 +220,7 @@ namespace bsp
 
         return 0;
     }
+
+    void usbHandleDataReceived()
+    {}
 } // namespace bsp
--- a/module-services/service-desktop/WorkerDesktop.cpp
+++ b/module-services/service-desktop/WorkerDesktop.cpp
@@ -148,6 +148,7 @@ bool WorkerDesktop::handleMessage(uint32_t queueID)
             usbSuspendTimer.stop();
             ownerService->bus.sendUnicast(std::make_shared<sdesktop::usb::USBConnected>(),
                                           service::name::service_desktop);
+            usbStatus = bsp::USBDeviceStatus::Connected;
         }
         else if (notification == bsp::USBDeviceStatus::Configured) {
             cpuSentinel->HoldMinimumFrequency(bsp::CpuFrequencyHz::Level_4);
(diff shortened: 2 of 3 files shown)


module-services/service-desktop/WorkerDesktop.cpp
```diff
@@ -161,14 +162,18 @@ bool WorkerDesktop::handleMessage(uint32_t queueID)
                                                   sdesktop::usb::USBConfigurationType::reconfiguration),
                                               service::name::service_desktop);
             }
+            usbStatus = bsp::USBDeviceStatus::Configured;
```
pholat, line 165: check if it doesn't break shutdown behaviour :)

module-bsp/board/linux/usb_cdc/usb_cdc.cpp
```diff
@@ -220,4 +220,7 @@ namespace bsp
 
         return 0;
     }
+
+    void usbHandleDataReceived()
```
alekrudnik, line 224: means it will no longer work on linux?
I believe it was working previously with virtual tty

module-services/service-desktop/WorkerDesktop.cpp
alekrudnik: good point, please also check if CDC is still working on simualtor

module-bsp/board/linux/usb_cdc/usb_cdc.cpp
wojtekrzepecki: It will still work there. I haven't changed the implementation there, only changed the API for the stack. Here I added only blank implementation to satisfy the compiler.

module-services/service-desktop/WorkerDesktop.cpp
wojtekrzepecki: @pholat Phone turning off when USB connected work flawlessly. 
@alekrudnik  USB for simulator wokrs fine as proven with harness test
sim_usb
:

alekrudnik (approved):

mjan-dev (approved):

pholat (approved):

wojtekrzepecki:
Conducted tests: 
- update via script and Mudita Center
- various harness testsboth on Phone and simulator
- Phone on/off with/without USB
- correctness of detected USB source

pawel-mudita (approved):

PR #3223, opened 2021-11-15
[BH-1166] Disable temperature feature

swichu91:
Added possibility to enable/disable
displaying of temperature and corresponding
settings/onboarding views.

### DIFF
--- a/products/BellHybrid/config/ProductConfig.in.hpp
+++ b/products/BellHybrid/config/ProductConfig.in.hpp
@@ -3,4 +3,4 @@
 
 #pragma once
 
-#define ENABLE_GSM (0)
+#cmakedefine01 CONFIG_ENABLE_TEMP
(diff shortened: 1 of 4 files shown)


mjan-dev (approved):

products/BellHybrid/config/ProductConfig.in.hpp
```diff
@@ -3,4 +3,4 @@
 
 #pragma once
 
-#define ENABLE_GSM (0)
+#cmakedefine01 CONFIG_ENABLE_TEMP
```
wojtekrzepecki, line 6: Is it a proper way to pass cmake-defined options?

wojtekrzepecki (approved):

products/BellHybrid/config/ProductConfig.in.hpp
swichu91: According to https://cmake.org/cmake/help/latest/command/configure_file.html, it is the preferred way of doing it instead of passing raw defines.

PR #3231, opened 2021-11-16
[EGD-7973] Fixed luts file problem

pholat:
Luts file was to big to download via api which limit is 1MB

### DIFF
--- a/products/PurePhone/assets.json
+++ b/products/PurePhone/assets.json
@@ -67,8 +67,9 @@
             "output": "image/assets/fonts/gt_pressura_light_90.mpf"
         },
         {
-            "name": "./image/Luts.bin",
-            "output": "./image/Luts.bin"
+            "name": "image/Luts.bin.tgz",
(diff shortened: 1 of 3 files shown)


alekrudnik (approved):

products/PurePhone/assets.json
```diff
@@ -67,8 +67,9 @@
             "output": "image/assets/fonts/gt_pressura_light_90.mpf"
         },
         {
-            "name": "./image/Luts.bin",
-            "output": "./image/Luts.bin"
+            "name": "image/Luts.bin.tgz",
+            "output": "./image/Luts.bin",
+            "unpack": true
```
SP2FET, line 72: please add the same for bell. Tested with pure - works

SP2FET (approved):

products/PurePhone/assets.json
pholat: fixed tx :)

niepiekm (approved):

PR #3251, opened 2021-11-17
[BH-1195] Fix audio playback while volume change

tigransoghbatyan-digica:
Prevent audio restart on volume setting change

### DIFF
--- a/products/BellHybrid/apps/application-bell-settings/presenter/alarm_settings/PrewakeUpPresenter.cpp
+++ b/products/BellHybrid/apps/application-bell-settings/presenter/alarm_settings/PrewakeUpPresenter.cpp
@@ -8,16 +8,16 @@ namespace app::bell_settings
 {
     PrewakeUpWindowPresenter::PrewakeUpWindowPresenter(std::shared_ptr<PrewakeUpListItemProvider> provider,
                                                        std::unique_ptr<AbstractPrewakeUpSettingsModel> model,
-                                                       std::unique_ptr<AbstractAudioModel> audioModel,
+                                                       AbstractAudioModel &audioModel,
                                                        std::unique_ptr<AbstractSoundsRepository> soundsRepository)
         : provider(std::move(provider)),
-          model(std::move(model)), audioModel{std::move(audioModel)}, soundsRepository{std::move(soundsRepository)}
+          model(std::move(model)), audioModel{audioModel}, soundsRepository{std::move(soundsRepository)}
     {
         auto playSound = [this](const UTF8 &val) {
             currentSoundPath = val;
-            this->audioModel->play(this->soundsRepository->titleToPath(currentSoundPath).value_or(""),
-                                   AbstractAudioModel::PlaybackType::PreWakeup,
-                                   {});
+            this->audioModel.play(this->soundsRepository->titleToPath(currentSoundPath).value_or(""),
+                                  AbstractAudioModel::PlaybackType::PreWakeup,
+                                  {});
         };
 
         this->provider->onExit = [this]() { getView()->exit(); };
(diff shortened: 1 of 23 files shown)


products/BellHybrid/apps/application-bell-settings/presenter/alarm_settings/PrewakeUpPresenter.cpp
```diff
@@ -30,7 +30,9 @@ namespace app::bell_settings
         this->provider->onVolumeExit   = [this](const auto &) { this->stopSound(); };
         this->provider->onVolumeChange = [this, playSound](const auto &val) {
             this->audioModel->setVolume(val, AbstractAudioModel::PlaybackType::PreWakeup, {});
-            playSound(currentSoundPath);
+            if (this->audioModel->hasPlaybackFinished()) {
```
mjan-dev, line 33: why only the prewakeup presenter checks if the playback has finished?

tigransoghbatyan-digica: fixed

products/BellHybrid/apps/include/AudioApplication.hpp
```diff
@@ -0,0 +1,30 @@
...
+                         StartInBackground startInBackground = {false},
+                         uint32_t stackDepth                 = 4096,
+                         sys::ServicePriority priority       = sys::ServicePriority::Idle)
+            : Application(std::move(name), std::move(parent), statusIndicators, startInBackground, stackDepth),
+              audioModel{std::make_shared<AudioModel>(this)}
+        {
+            bus.channels.push_back(sys::BusChannel::ServiceAudioNotifications);
+        }
+    };
+} // namespace app
```
swichu91, line 30: I get you are trying to remove unnecessary code duplication, which is fine but I'm not sure, if this is the correct solution. I strongly believe, the composition should be used in favor of inheritance. After all, it's just two lines more that needs to be added to applications:
```
std::unique_ptr<AbstractAudioModel> audioModel;
bus.channels.push_back(sys::BusChannel::ServiceAudioNotifications);
```

products/BellHybrid/apps/include/AudioApplication.hpp
```diff
@@ -0,0 +1,30 @@
...
+#include <Application.hpp>
+#include <common/models/AbstractAudioModel.hpp>
+#include <common/models/AudioModel.hpp>
+
+namespace app
+{
+    class AudioApplication : public Application
+    {
+      protected:
+        std::shared_ptr<AbstractAudioModel> audioModel;
```
swichu91, line 15: Do we really need a shared pointer here?

products/BellHybrid/apps/common/include/common/models/BedtimeModel.hpp
```diff
@@ -65,7 +65,7 @@ namespace app::bell_bedtime
       public:
         BedtimeModel() = delete;
 
-        explicit BedtimeModel(ApplicationCommon *app, std::unique_ptr<AudioModel> &&audioModel)
+        explicit BedtimeModel(ApplicationCommon *app, std::shared_ptr<AbstractAudioModel> audioModel)
```
swichu91, line 68: Does `BedTimeModel` really need to have a shared pointer to `audioModel`? Shared pointers are much slower than unique pointers or references. Moreover, the cases where you need them are rather rare. In this case, a simple reference will be sufficient. This also applies to other models/presenters.

swichu91 (approved):

mjan-dev (approved):

tigransoghbatyan-digica:
Merged in aggregated PR

PR #3253, opened 2021-11-17
[BH-1178] Fix onboarding guidelines

paweljonskim:
Fix onboarding guidelines:
- exiting info screen only on specific keypresses
- deep press up also activatess info screen
- deep press info screen also works on other info screens

### DIFF
--- a/products/BellHybrid/apps/application-bell-onboarding/windows/OnBoardingInstructionPromptWindow.cpp
+++ b/products/BellHybrid/apps/application-bell-onboarding/windows/OnBoardingInstructionPromptWindow.cpp
@@ -27,10 +27,7 @@ namespace gui
 
     bool OnBoardingInstructionPromptWindow::onInput(const InputEvent &inputEvent)
     {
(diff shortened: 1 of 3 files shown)


mjan-dev (approved):

products/BellHybrid/apps/application-bell-onboarding/windows/OnBoardingInstructionPromptWindow.cpp
```diff
@@ -27,10 +27,7 @@ namespace gui
 
     bool OnBoardingInstructionPromptWindow::onInput(const InputEvent &inputEvent)
```
PrzeBrudny, line 28: hmm so remove it if it is not used?

paweljonskim: we need to override it so it does nothing

PrzeBrudny (approved):

PR #3256, opened 2021-11-17
Bh 1192 powernap ends with zero

paweljonskim:


module-apps/apps-common/widgets/ProgressTimer.cpp
```diff
@@ -55,6 +55,7 @@ namespace app
     auto ProgressTimer::onTimerTimeout(sys::Timer &task) -> bool
     {
         ++elapsed;
+        update();
```
piotr-tanski, line 58: Doesn't it change the behavior of meditation timer?

paweljonskim: I think that logically this is correct - last tick should also cause to update progress. In this case to `-0:00`. I checked the Meditation app on simulator and everything seems to work correctly.

mjan-dev (approved):

piotr-tanski (approved):

PR #3265, opened 2021-11-18
[BH-1211] Fix bell audio use cpu sentinel

wojtekrzepecki:
hold of CPU freq when audio is active

### DIFF
--- a/products/BellHybrid/services/audio/ServiceAudio.cpp
+++ b/products/BellHybrid/services/audio/ServiceAudio.cpp
@@ -144,6 +144,7 @@ namespace service
         };
         auto input = audioMux.GetPlaybackInput(playbackType);
         AudioStart(input);
+        manageCpuSentinel();
 
         return std::make_unique<AudioStartPlaybackResponse>(retCode, retToken);
     }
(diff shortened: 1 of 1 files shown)


mjan-dev (approved):

pawel-mudita (approved):

swichu91 (approved):

products/BellHybrid/services/audio/ServiceAudio.cpp
```diff
@@ -284,6 +287,7 @@ namespace service
             activeInput && activeInput.value()->audio->GetCurrentOperationState() == audio::Operation::State::Paused) {
             retCode = activeInput.value()->audio->Resume();
         }
+        manageCpuSentinel();
```
pholat, line 290: tbh - cant this be somehow done other way round? `if bussy -> manage sentinel` not `manage sentinel -> if bussy` ? :)

wojtekrzepecki: The status is taken unfortunately from the lowest layers of the audio stack and the callback would need to be installed there.

PR #3268, opened 2021-11-18
[EGD-7988] Fix 'forget' HF in BT devices window

SP2FET:
Fixes HF when forgetting more than available devices

### DIFF
--- a/module-apps/application-settings/models/bluetooth/BluetoothSettingsModel.cpp
+++ b/module-apps/application-settings/models/bluetooth/BluetoothSettingsModel.cpp
@@ -144,3 +144,7 @@ auto BluetoothSettingsModel::getStatus() const -> const BluetoothStatus
 {
     return status;
 }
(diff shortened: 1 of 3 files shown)


module-apps/application-settings/models/bluetooth/BluetoothSettingsModel.cpp
```diff
@@ -144,3 +144,7 @@ auto BluetoothSettingsModel::getStatus() const -> const BluetoothStatus
 {
     return status;
 }
+auto BluetoothSettingsModel::isDeviceListEmpty() const -> bool
+{
+    return devices.size() == 0;
```
pholat, line 149: just return `devices.empty()` ? :)

SP2FET: true :D

pholat (approved):

wojtekrzepecki (approved):

PR #3271, opened 2021-11-18
[CP-813] Fix issues with adding contacts via Mudita Center

pawel-mudita:
There was an issue, when contact with a specific number was added,
then removed and added again via Mudita Center. Then
it was displayed without a name. This issue was caused by
another known issue with temporary contacts handling (EGD-79473).
This fix is a workaround to make adding contacts via MC work,
without the need to refactor temporary contacts handling.


niepiekm (approved):

piotr-tanski (approved):
LGTM

PR #3282, opened 2021-11-19
[BH-955] Fix onClose when popup displayed

wojtekrzepecki:
no action on onClose when switched to popup window

### DIFF
--- a/module-apps/apps-common/ApplicationCommon.cpp
+++ b/module-apps/apps-common/ApplicationCommon.cpp
@@ -540,9 +540,17 @@ namespace app
             }
             if (!isCurrentWindow(msg->getWindowName())) {
                 if (!windowsStack.isEmpty()) {
-                    const auto closeReason = msg->getReason() == SwitchReason::PhoneLock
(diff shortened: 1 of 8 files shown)


PrzeBrudny (approved):

module-apps/apps-common/ApplicationCommon.cpp
```diff
@@ -540,9 +540,17 @@ namespace app
             }
             if (!isCurrentWindow(msg->getWindowName())) {
                 if (!windowsStack.isEmpty()) {
-                    const auto closeReason = msg->getReason() == SwitchReason::PhoneLock
-                                                 ? gui::Window::CloseReason::PhoneLock
-                                                 : gui::Window::CloseReason::WindowSwitch;
+                    auto closeReason = gui::Window::CloseReason::WindowSwitch;
```
pholat, line 543: just saying this code is messed up. This callback is called on switch on previous window, not exactly on close (and we can miss on close too - I've added issue for that today :) ) 
saying that - it's ok.

pholat (approved):

mjan-dev (approved):

PR #3289, opened 2021-11-22
[EGD-7993] Fixed contacts matching algorithm

piotr-tanski:
number.isValid() always returns false if there is no national prefix.

### DIFF
--- a/module-db/Interface/ContactRecord.cpp
+++ b/module-db/Interface/ContactRecord.cpp
@@ -1159,10 +1159,11 @@ auto ContactRecordInterface::buildNumberMatcher(unsigned int maxPageSize)
 {
     return utils::NumberHolderMatcher<std::vector, ContactNumberHolder>(
         [this](const utils::PhoneNumber &number, auto offset, auto limit) {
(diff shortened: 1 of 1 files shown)


pawel-mudita (approved):

module-db/Interface/ContactRecord.cpp
```diff
@@ -1159,10 +1159,11 @@ auto ContactRecordInterface::buildNumberMatcher(unsigned int maxPageSize)
 {
     return utils::NumberHolderMatcher<std::vector, ContactNumberHolder>(
         [this](const utils::PhoneNumber &number, auto offset, auto limit) {
+            const auto numbers = !number.get().empty() ? contactDB->number.getLimitOffset(number.get(), offset, limit)
```
alekrudnik, line 1162: is isValid our own method? We can consider fixing it....

alekrudnik (approved):

module-db/Interface/ContactRecord.cpp
piotr-tanski: It seems not. It's using `IsValidNumber` from the libphonenumber, we only have a wrapper for that.

PR #3299, opened 2021-11-22
[BH-1240] Prewakeup frontlight when alarm light off

paweljonskim:


products/BellHybrid/alarms/src/actions/FrontlightAction.cpp
```diff
@@ -54,17 +54,31 @@ namespace alarms
...
+            }
+            break;
+        case SettingsDependency::Prewakeup:
+            settingString =
+                settings.getValue(bell::settings::PrewakeUp::lightDuration, settings::SettingsScope::Global);
+            if (settingString == std::string(prewakeupFrontlightOFF)) {
+                return true;
+            }
+            break;
+        default:
```
alekrudnik, line 80: how many cases are there?
Can you cover all of them without using default?

products/BellHybrid/alarms/src/actions/FrontlightAction.cpp
```diff
@@ -54,17 +54,31 @@ namespace alarms
...
-            settings.getValue(bell::settings::Alarm::lightActive, settings::SettingsScope::Global);
-        if (flontlightEnabled == std::string(frontlightOFF)) {
-            return true;
+        std::string settingString;
+
+        switch (settingsDependency) {
+        case SettingsDependency::AlarmClock:
+            settingString = settings.getValue(bell::settings::Alarm::lightActive, settings::SettingsScope::Global);
+            if (settingString == std::string(alarmFrontlightOFF)) {
+                return true;
```
piotr-tanski, line 70: Why not use different implementations (reuse the pimpl idiom which is already implemented) instead of introducing such conditions?
Is AlarmClock using both Linear and Manual modes, depending on the situation??

paweljonskim: @piotr-tanski 
I did not want to bind two different traits of an alarm:
- value changing characteristics (Manual/Linear)
and 
- configuration dependency
into one place. I did not want to create multiple frontligt actions differing only on condition check.
Im open to change, not sure if you wanted the explaination or require change here, please specify :)

products/BellHybrid/alarms/src/actions/FrontlightAction.hpp
```diff
@@ -19,13 +19,24 @@ namespace alarms
             LinearProgress
         };
 
-        FrontlightAction(sys::Service &service, Mode mode);
+        enum class SettingsDependency
+        {
+            None,
```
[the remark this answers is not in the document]
paweljonskim: No, but I think I should implement ready "plug-n-play" way to use frontlight actions without settings condition checking for future developers

products/BellHybrid/alarms/src/actions/FrontlightAction.cpp
```diff
@@ -54,17 +54,31 @@ namespace alarms
...
-        const std::string flontlightEnabled =
-            settings.getValue(bell::settings::Alarm::lightActive, settings::SettingsScope::Global);
-        if (flontlightEnabled == std::string(frontlightOFF)) {
-            return true;
+        std::string settingString;
+
+        switch (settingsDependency) {
+        case SettingsDependency::AlarmClock:
+            settingString = settings.getValue(bell::settings::Alarm::lightActive, settings::SettingsScope::Global);
+            if (settingString == std::string(alarmFrontlightOFF)) {
```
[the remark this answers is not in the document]
paweljonskim: thanks for the tip! done

products/BellHybrid/alarms/src/actions/FrontlightAction.cpp
paweljonskim: done

products/BellHybrid/alarms/src/actions/FrontlightAction.hpp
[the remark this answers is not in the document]
piotr-tanski: TBH, I'm always against introducing the mechanisms that are not used and not needed.
Once it's really needed, you'll know more and the implementation will be more meaningful. But this is a minor in that case then.

products/BellHybrid/alarms/src/actions/FrontlightAction.cpp
piotr-tanski: Please, keep in mind that the LinearProgressFrontlightAction is actually pre-wake up only. It uses pre-wake up settings like pre wake up light duration.
So actually LinearProgressFrontlightAction is pre wake up only, and the Manual one is alarm clock only in the current implementation. Probably this code is not needed in the common part then.

tigransoghbatyan-digica:
I just tested this PR. I set alarm FL off and prewakeup FL on. When the time for prewakeup came there was no light both during prewakeup and alarm

tomaszrybarski-digica (approved):

tigransoghbatyan-digica (approved):

paweljonskim:
Closing since:
1. one more change is needed, which require further review
2. merge will be done to branch `stable` not release
3. all active people are either on vacation or not working on the project anymore

PR #3306, opened 2021-11-23
[BH-1243] Fix arrows behaviour in spinners

tigransoghbatyan-digica:
Refactor and fix arrows visibility in spinners

### DIFF
--- a/products/BellHybrid/apps/application-bell-settings/models/alarm_settings/SnoozeListItemProvider.cpp
+++ b/products/BellHybrid/apps/application-bell-settings/models/alarm_settings/SnoozeListItemProvider.cpp
@@ -11,15 +11,7 @@ namespace app::bell_settings
 {
     using namespace gui;
 
-    enum Intervals
-    {
-        Interval_1  = 1,
-        Interval_2  = 2,
-        Interval_5  = 5,
-        Interval_10 = 10,
-        Interval_15 = 15,
-        Interval_30 = 30
-    };
+    const std::set<unsigned int> allowedIntervals{1, 2, 5, 10, 15, 30};
 
     SnoozeListItemProvider::SnoozeListItemProvider(AbstractSnoozeSettingsModel &model,
                                                    std::vector<UTF8> chimeTonesRange)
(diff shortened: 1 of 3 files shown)


products/BellHybrid/apps/application-bell-settings/models/alarm_settings/SnoozeListItemProvider.cpp
```diff
@@ -83,16 +75,11 @@ namespace app::bell_settings
...
                 chimeRange.push_back(NumWithStringListItem::Value{utils::translate("app_alarm_clock_no_snooze")});
-                for (unsigned int i = 1; i <= val; i++) {
-                    if ((i != Intervals::Interval_1) && (i != Intervals::Interval_2) && (i != Intervals::Interval_5) &&
-                        (i != Intervals::Interval_10) && (i != Intervals::Interval_15) &&
-                        (i != Intervals::Interval_30)) {
-                        continue;
-                    }
-                    chimeRange.push_back(NumWithStringListItem::Value{i, minStr});
-                }
+                std::for_each(allowedIntervals.begin(), allowedIntervals.lower_bound(val), [&](auto val) {
```
tomaszrybarski-digica, line 78: What happens when, for example: val == 12 (snooze length set in previous window), won't std::lower_bound point to next greater than 12 in the set, which is 15 and add it to the allowable range for chime interval? i think it is not supposed to work this way, for a val of 12 the allowable max. limit for interval should be 10 (always lower than chime length)?

swichu91 (commented):
I'm afraid it does not solve the problem entirely. Please see the discussion here https://appnroll.slack.com/archives/CCMBKV2NQ/p1637666690009300 .

products/BellHybrid/apps/application-bell-settings/models/alarm_settings/SnoozeListItemProvider.cpp
tigransoghbatyan-digica: No, the lower bound is choosing a lower value between the two closest in the set

tomaszrybarski-digica: In that case it's fine, i was looking at https://www.geeksforgeeks.org/set-lower_bound-function-in-c-stl/ - maybe there is an error there

tigransoghbatyan-digica: can we resolve this conversation?

tigransoghbatyan-digica:
> I'm afraid it does not solve the problem entirely. Please see the discussion here https://appnroll.slack.com/archives/CCMBKV2NQ/p1637666690009300 .

I have no access to this conversation

mjan-dev (changes requested):
Blocking until new branch for patch 0 will be created. It is not of highest priority.

mjan-dev (approved):
As @swichu91 mentioned, it does not solve the problem, so the task cannot be set to done. However, it should improve the behavior nevertheless. I see no harm in merging it.

mjan-dev:
This behavior was already fixed in other PRs. Closing.

PR #3318, opened 2021-11-24
[BH-1254] Charging battery UI fix

paweljonskim:
Charging battery UI fix:
- bigger font
- center widget


paweljonskim:
Green - Figma
Red - simulator
Im aware that there is 1px vertical offset, but it is visible on all elements, so changing whole screen might be neccessary. I suggest another task for this
Selection_201
Selection_200

mjan-dev (approved):

PR #3320, opened 2021-11-24
[EGD-7875] Fixed bt not ringing

jimmorrisson:
Fixed bluetooth not ringing after bt call hang up.

### DIFF
--- a/module-bluetooth/Bluetooth/interface/profiles/HFP/HFP.cpp
+++ b/module-bluetooth/Bluetooth/interface/profiles/HFP/HFP.cpp
@@ -276,7 +276,7 @@ namespace bluetooth
             scoHandle = HCI_CON_HANDLE_INVALID;
             isAudioRouted                = false;
             isAudioConnectionEstablished = false;
-            sendAudioEvent(audio::EventType::BlutoothHFPDeviceState, audio::Event::DeviceState::Disconnected);
+            audioDevice.reset();
             break;
         case HFP_SUBEVENT_START_RINGING:
             LOG_DEBUG("Start Ringing\n");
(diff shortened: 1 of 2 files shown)


SP2FET (approved):

module-bluetooth/Bluetooth/interface/profiles/HFP/HFP.cpp
```diff
@@ -276,7 +276,7 @@ namespace bluetooth
             scoHandle = HCI_CON_HANDLE_INVALID;
             isAudioRouted                = false;
             isAudioConnectionEstablished = false;
-            sendAudioEvent(audio::EventType::BlutoothHFPDeviceState, audio::Event::DeviceState::Disconnected);
+            audioDevice.reset();
```
jimmorrisson, line 279: It is needed otherwise this **https://appnroll.atlassian.net/browse/EGD-7299** bug occures.

pholat, line 279: can't it be somehow handled better in module-audio?

tomaszkrosnowski (approved):

module-bluetooth/Bluetooth/interface/profiles/HFP/HFP.cpp
alekrudnik: but why in BT thread?
there is no race?

jimmorrisson: Module audio doesn't get messages from bt stack. Sure it could be done in module audio, but to my mind we would need to have another message from module bt to module audio - that approach is error prone cause of the extra time that is needed to deliver a message plus we would need to handle a case in which message can't be delivered for whatever reason.

jimmorrisson: @alekrudnik audiodevice is part of hfp pimpl

pholat: ok, and why does it crash exactly? right me if I'm wrong but it's not like calling perfectly normal action should crash system.

pholat: sorry but I'm not convinced. `would need to handle a case in which message can't be delivered for whatever reason.` It looks like a very poor design here. Do we have any documentation for that?

PR #3321, opened 2021-11-24
[EGD-7697] Fix BT names in all files

SP2FET:
Some fallback names were incorrect

### DIFF
--- a/.idea/modules.xml
+++ b/.idea/modules.xml
@@ -2,6 +2,7 @@
 <project version="4">
   <component name="ProjectModuleManager">
     <modules>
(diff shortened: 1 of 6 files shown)


.idea/modules.xml
```diff
@@ -2,6 +2,7 @@
 <project version="4">
   <component name="ProjectModuleManager">
     <modules>
+      <module fileurl="file://$PROJECT_DIR$/.idea/PureOS.iml" filepath="$PROJECT_DIR$/.idea/PureOS.iml" />
```
SP2FET, line 5: This is just annoying fix for broken CLion config (broken during product split months ago)

pawel-mudita (approved):

pholat (approved):

PR #3350, opened 2021-11-30
[BH-1273] Fix audio asset playback stop

tigransoghbatyan-digica:
Make sure apps are only stopping their own audio playbacks

### DIFF
--- a/products/BellHybrid/apps/include/Application.hpp
+++ b/products/BellHybrid/apps/include/Application.hpp
@@ -21,6 +21,7 @@ namespace app
         void restartIdleTimer();
         void stopIdleTimer();
         void stopAllAudio();
+        virtual void onStop();
 
       private:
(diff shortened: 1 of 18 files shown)


products/BellHybrid/apps/include/Application.hpp
```diff
@@ -29,6 +30,6 @@ namespace app
 
         virtual void onKeyPressed();
         virtual void onStart();
-        virtual void onStop();
+        bool idleTimerActiveFlag = true;
```
mjan-dev, line 33: is this change connected with this task?

mjan-dev (approved):
Approved with minor comment.

products/BellHybrid/apps/include/Application.hpp
tigransoghbatyan-digica: no, it sneaked in accidentally. Removed

wojtekrzepecki (approved):

PR #3376, opened 2021-12-03
[EGD-7922] Permanent hold cpu frequency setting window

PrzeBrudny:
image

### DIFF
--- a/module-apps/application-settings/windows/advanced/CPUModeTestWindow.cpp
+++ b/module-apps/application-settings/windows/advanced/CPUModeTestWindow.cpp
@@ -1,0 +1,193 @@
+// Copyright (c) 2017-2021, Mudita Sp. z.o.o. All rights reserved.
+// For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
+
+#include "CPUModeTestWindow.hpp"
+
+#include <system/messages/SentinelRegistrationMessage.hpp>
+#include <application-settings/windows/WindowNames.hpp>
+#include <Application.hpp>
+#include <TextFixedSize.hpp>
+#include <BoxLayout.hpp>
+
+#include <magic_enum.hpp>
+
+namespace magic_enum
+{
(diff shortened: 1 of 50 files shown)


module-apps/application-settings/windows/advanced/CPUModeTestWindow.cpp
```diff
@@ -0,0 +1,193 @@
...
+    template <> struct enum_range<bsp::CpuFrequencyHz>
+    {
+        static constexpr int min = 0;
+        static constexpr int max = 528;
+    };
+} // namespace magic_enum
+
+namespace gui
+{
+    CPUModeTestWindow::CPUModeTestWindow(app::ApplicationCommon *app) : AppWindow(app, window::name::cpu_test_window)
```
PrzeBrudny, line 25: That is only Dev Advanced window -> it has no localization and proper style variables as it is not for clients. It is not attached in Release builds.

module-bsp/bsp/common.hpp
```diff
@@ -15,13 +15,13 @@ namespace bsp
     /// Only a few thresholds are available in the current configuration
     enum class CpuFrequencyHz
```
PrzeBrudny, line 16: I will change name to MHz after approval (it touches a lot of files so do not want to introduce it now to clear up review).

wojtekrzepecki: Wouldn't it be just easier to name it like:
`
level_4MHZ,
level_12MHz,
...
` ?
Probably the values are not use anywhere

wojtekrzepecki (approved):

mjan-dev (approved):

module-bsp/bsp/common.hpp
PrzeBrudny: Well they are used now to print in that new window :D

pawel-mudita (approved):

Maciej-Mudita (approved):

PR #3380, opened 2021-12-07
[CP-831] Move and repair test_device_diagnostics.py

niepiekm:
Moved test script out of pytest/service-desktop
Cleaned up and fixed test code for multiple log/crash dump files

### DIFF
--- a/test/pytest/test_device_diagnostics.py
+++ b/test/pytest/test_device_diagnostics.py
@@ -6,22 +6,20 @@ from harness import log
 from harness.api.device_info import GetDiagnosticFilesList, DiagnosticsFileList
 from harness.request import TransactionError
 
+class DiagnosticsError(RuntimeError):
+    pass
+
 @pytest.mark.service_desktop_test
 @pytest.mark.usefixtures("phone_unlocked")
 def test_device_list_logs(harness):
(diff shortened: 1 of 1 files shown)


pawel-mudita (approved):

test/pytest/test_device_diagnostics.py
```diff
@@ -10,18 +10,17 @@
...
+        if not ret.files:
+            raise RuntimeError("No log files")
 
-    if not ret.files:
-        log.info("No log files")
-        assert True
-        return
+    except RuntimeError as err:
+        log.error(err)
+        pytest.exit(err)
```
pholat, line 19: this seems excessive. First call runtime error, to just catch it in next line? :) 

just do not catch the error and pytest will exit with full trace.

at best create own error after RuntimeError - i.e. 
```
class DiagnosticsError(RuntimeError)
    pass
```

niepiekm: Well, my intention was to avoid abrupt termination and just log the error and exit.

niepiekm: Changed as requested.

pholat (approved):

PR #3381, opened 2021-12-07
[BH-1236] Add temperature compensation

tigransoghbatyan-digica:
Add temperature internal heating compensation

### DIFF
--- a/products/BellHybrid/services/audio/ServiceAudio.cpp
+++ b/products/BellHybrid/services/audio/ServiceAudio.cpp
@@ -7,6 +7,7 @@
 #include <audio/AudioMessage.hpp>
 #include <service-db/Settings.hpp>
 #include <system/messages/SentinelRegistrationMessage.hpp>
+#include <service-evtmgr/service-evtmgr/Constants.hpp>
 
 namespace
 {
@@ -148,6 +149,7 @@ namespace service
         AudioStart(input);
         manageCpuSentinel();
 
+        bus.sendUnicast(std::make_shared<service::AudioPlaybackStartNotification>(), service::name::evt_manager);
         return std::make_unique<AudioStartPlaybackResponse>(retCode, retToken);
     }
     auto Audio::handleStop(const std::vector<audio::PlaybackType> &stopTypes, const audio::Token &token)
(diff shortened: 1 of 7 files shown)


products/BellHybrid/services/evtmgr/EventManager.cpp
```diff
@@ -119,6 +108,16 @@ void EventManager::initProductEvents()
...
+
+    connect(service::AudioPlaybackStartNotification(), [&](sys::Message *msgl) {
+        temperatureProvider.audioStarted();
+        return sys::msgHandled();
+    });
+
+    connect(service::AudioPlaybackStopNotification(), [&](sys::Message *msgl) {
+        temperatureProvider.audioStoped();
+        return sys::msgHandled();
+    });
```
wojtekrzepecki, line 120: In my opinion dependency to ServiceAudio is wrong here. More general solution will be to use our PowerManager to obtain current cpu load and cpu frequency to provide useful input to temperature compensation. Was the concept for this audio playback compensation discussed before in Harmony team?

swichu91: I agree with Wojtek. The same goes for the audio service which now requires event manager dependency. As far as I know, it was found audio service is the main source of the temperature rise. Of course, the solution proposed by Wojtek is much better.

products/BellHybrid/services/audio/ServiceAudio.cpp
```diff
@@ -171,7 +173,7 @@ namespace service
         if (it != retCodes.end()) {
             return std::make_unique<AudioStopResponse>(it->second, it->first);
         }
-
+        bus.sendUnicast(std::make_shared<service::AudioPlaybackStopNotification>(), service::name::evt_manager);
```
alekrudnik, line 176: why do you send it explicitly to event manager?

tigransoghbatyan-digica: according to current implementation, the temperature compensation algorithm is implemented in EventManager Service

products/BellHybrid/services/evtmgr/temperature-compensation/TemperatureProvider.cpp
```diff
@@ -0,0 +1,82 @@
...
+{
+    auto diff =
+        std::chrono::duration_cast<std::chrono::seconds>(std::chrono::system_clock::now() - audioLastTimeStateChange);
+    auto secondsElapsed = diff.count();
+    auto decay          = std::exp(-secondsElapsed / audioCompensationTimeCoefficient);
+    if (audioPlaying) {
+        audioCompensationValue += (audioCompensationAmplitude - audioCompensationValue) * (1 - decay);
+    }
+    else {
+        audioCompensationValue = (audioCompensationValue)*decay;
```
mjan-dev, line 48:
```suggestion
        audioCompensationValue *= decay;
```

mjan-dev:
Closing as this solution was tested and seems to be not effective.

PR #3382, opened 2021-12-07
[EGD-7993] Fix memory leak after import contact via MC

Maciej-Mudita:
Reducing the memory allocation for re2,
which results in a change of the algorithm from DFA to NFA,
which consumes less resources.


lucckb (approved):

kkleczkowski (approved):

alekrudnik (approved):

pawel-mudita:
Tested on >2k contact records - issue no longer observed.
Heap stats are fine.

pawel-mudita (approved):

PR #3385, opened 2021-12-08
Bluetooth SSP minimal changes to work

pholat:
EGD-8002

There are now the following options to pair with our bt device:
1. legacy pairing
2. ssp phone -> host
3. ssp host -> phone


.gdbinit-1051
```diff
@@ -13,4 +13,4 @@ info threads
 thread 2
 tb main
 b HardFault_Handler
-
+b _exit
```
pholat, line 16: convenience only - if not added, then you need to add it by hand :)

module-bluetooth/Bluetooth/error_bluetooth.cpp
```diff
@@ -0,0 +1,138 @@
+#include "error_bluetooth.hpp"
```
pholat, line 1: this can be guarded after some Bluetooth only debug flag :)

module-utils/log/Logger.cpp
```diff
@@ -19,7 +19,6 @@ namespace Log
 #endif
                                                             {"ServiceAntenna", logger_level::LOGERROR},
                                                             {"ServiceAudio", logger_level::LOGINFO},
-                                                            {"ServiceBluetooth", logger_level::LOGINFO},
```
pholat, line 22: this can be reverted

module-bluetooth/Bluetooth/error_bluetooth.cpp
```diff
@@ -0,0 +1,138 @@
...
+#include <module-bluetooth/lib/btstack/src/bluetooth.h>
+
+namespace bluetooth {
+const char *error_cstr(int err)
```
niepiekm, line 5: Nice candidate to replace with [X-Macro](https://quuxplusone.github.io/blog/2021/02/01/x-macros/) technique

module-bluetooth/Bluetooth/error_bluetooth.cpp
```diff
@@ -0,0 +1,138 @@
...
+        return "ERROR_CODE_DIRECTED_ADVERTISING_TIMEOUT";
+    case ERROR_CODE_CONNECTION_TERMINATED_DUE_TO_MIC_FAILURE:
+        return "ERROR_CODE_CONNECTION_TERMINATED_DUE_TO_MIC_FAILURE";
+    case ERROR_CODE_CONNECTION_FAILED_TO_BE_ESTABLISHED:
+        return "ERROR_CODE_CONNECTION_FAILED_TO_BE_ESTABLISHED";
+    case ERROR_CODE_MAC_CONNECTION_FAILED:
+        return "ERROR_CODE_MAC_CONNECTION_FAILED";
+    case ERROR_CODE_COARSE_CLOCK_ADJUSTMENT_REJECTED_BUT_WILL_TRY_TO_ADJUST_USING_CLOCK_DRAGGING:
+        return "ERROR_CODE_COARSE_CLOCK_ADJUSTMENT_REJECTED_BUT_WILL_TRY_TO_ADJUST_USING_CLOCK_DRAGGING";
+    default:
```
alekrudnik, line 134: do you need this default? consider simply covering all cases

alekrudnik: or maybe use magic enum

module-bluetooth/Bluetooth/error_bluetooth.cpp
pholat: I even use string expansion macro in updater for the very same case, but for c++ it's considered code smell, therefore not added here.
Saying that, the creation of more elaborate Xmacros ( how author called them) is a horrible mess to debug for more complex solutions :)
Here it would be pretty simple if in Bluetooth library this would be an enum - magic enum would suffice.

module-bluetooth/Bluetooth/error_bluetooth.cpp
pholat: @alekrudnik this is just a subset of codes + library has its own defined ones. This should suffice for now for our needs in terms - something is better than nothing.
We cant use magic enum - as these are not enums. These are defines :)

module-apps/application-settings/windows/bluetooth/BluetoothCheckPasskeyWindow.hpp
```diff
@@ -13,7 +13,7 @@ namespace gui
     class BluetoothCheckPasskeyWindow : public AppWindow
     {
       public:
-        explicit BluetoothCheckPasskeyWindow(app::ApplicationCommon *app);
+        explicit BluetoothCheckPasskeyWindow(app::ApplicationCommon *app, std::shared_ptr<BluetoothSettingsModel> bluetoothSettingsModel);
```
SP2FET, line 16: explicit is not needed anymore

module-apps/application-settings/windows/bluetooth/BluetoothCheckPasskeyWindow.hpp
```diff
@@ -22,6 +22,6 @@ namespace gui
         Image *image = nullptr;
         Label *label = nullptr;
         Text *text   = nullptr;
-        std::unique_ptr<BluetoothSettingsModel> bluetoothSettingsModel;
+        std::shared_ptr<BluetoothSettingsModel> bluetoothSettingsModel;
```
SP2FET, line 25: :+1:

module-bluetooth/Bluetooth/Device.cpp
```diff
@@ -0,0 +1,78 @@
...
+}
+
+Devicei::Devicei(Devicei &&d) noexcept : Devicei(d.name.data())
+{
+    operator=(d);
+}
+
+void Devicei::setAddress(bd_addr_t *addr)
+{
+    memcpy(&address, addr, sizeof address);
```
SP2FET, line 62: I think that there's a function in BT stack to copy the address, but I won't be surprised if it was just a wrapped `memcpy`

module-bluetooth/Bluetooth/error_bluetooth.cpp
SP2FET: like @pholat said, those codes are defines in stack header - we are not able to convert it to enums/strings automatically.

SP2FET (approved):

module-apps/application-settings/windows/bluetooth/BluetoothCheckPasskeyWindow.hpp
pholat: this was a funny one - it was defined in the application and here, under the same name and I spend quite a bit of time wondering - why the set device doesn't match the device received in application :)

module-bluetooth/Bluetooth/Device.cpp
pholat: I just moved the header code to cpp in this case :)

module-apps/application-settings/models/bluetooth/BluetoothSettingsModel.hpp
```diff
@@ -40,6 +40,7 @@ class BluetoothSettingsModel
     auto isDeviceConnecting() -> bool;
     auto getStatus() const -> const BluetoothStatus;
     auto isDeviceListEmpty() const -> bool;
+    Devicei pinRequestor;
```
alekrudnik, line 43: why need another one as there is devicei vector below?

module-bluetooth/Bluetooth/Device.hpp
```diff
@@ -96,37 +100,25 @@ enum class DeviceState
...
-    bd_addr_t address;
-    uint8_t pageScanRepetitionMode;
-    uint16_t clockOffset;
-    uint32_t classOfDevice; ///> Class of Device/Service
+    mutable bd_addr_t address{};
+    uint8_t pageScanRepetitionMode{};
+    uint16_t clockOffset{};
+    uint32_t classOfDevice{}; ///> Class of Device/Service
     DEVICE_STATE state;
     DeviceState deviceState;
```
alekrudnik, line 108: no default assignments?

module-bluetooth/Bluetooth/interface/profiles/GAP/GAP.cpp
```diff
@@ -94,48 +109,28 @@ namespace bluetooth
...
-            if (packet[2] == 0) {
-                devices[index].state = REMOTE_NAME_FETCHED;
-                strcpy(devices[index].name.data(), reinterpret_cast<const char *>(&packet[9]));
-                return true;
-            }
-            else {
-                LOG_INFO("Failed to get name: page timeout");
+        if (auto it = devices().find(addr); it != devices().end()) {
+            it->state = packet[2] ? REMOTE_NAME_FAILURE : REMOTE_NAME_FETCHED;
+            if ( it->state != REMOTE_NAME_FAILURE ) {
```
alekrudnik, line 130: while at it, consider naming 3 and 2.

module-bluetooth/Bluetooth/interface/profiles/GAP/GAP.hpp
```diff
@@ -12,35 +12,40 @@ extern "C"
...
-        done
-    };
+        enum state
+        {
+            off,
+            init,
+            working,
+            halting,
+            sleeping,
+            falling_asleep
```
alekrudnik, line 24:
```suggestion
        enum class state
        {
            off,
            init,
            working,
            halting,
            sleeping,
            falling_asleep
```

module-utils/log/Logger.cpp
```diff
@@ -20,6 +20,7 @@ namespace Log
                                                             {"ServiceAntenna", logger_level::LOGERROR},
                                                             {"ServiceAudio", logger_level::LOGINFO},
                                                             {"ServiceBluetooth", logger_level::LOGINFO},
+                                                            {"ServiceBluetooth_w1", logger_level::LOGINFO},
```
alekrudnik, line 23: needed?

alekrudnik (approved):

module-apps/application-settings/models/bluetooth/BluetoothSettingsModel.hpp
pholat: This are two different elements:
- this is the device which requested for pin
- the vector is list of elements in scan and is a mirror of data in Bluetooth service

module-bluetooth/Bluetooth/interface/profiles/GAP/GAP.cpp
pholat: tbh this is taken straight from packet definition - this should be rather addressed in Bluetooth stack, not in our code. in most scenarios error code is on 2'nd field, bt for some it's not :)

module-utils/log/Logger.cpp
pholat: covinience to not log debug information added while using the Bluetooth.
Tbh it would be great to have window to set these in i.e. settings and be able to change it in runtime

module-bluetooth/Bluetooth/error_bluetooth.cpp
niepiekm: You're right, `magic_enum` would be a better choice.

module-utils/log/Logger.cpp
niepiekm: We have a Jira for that, @pholat :)

niepiekm (approved):

module-bluetooth/Bluetooth/interface/profiles/GAP/GAP.cpp
```diff
@@ -300,8 +364,45 @@ namespace bluetooth
...
+        }
+
+        unsigned int passkey = 0;
+        try {
+            passkey = stoi(pin);
+            LOG_DEBUG("Sending %06u as a passkey", passkey);
+        }
+        catch (const std::invalid_argument &e) {
+            LOG_ERROR("STOI error: %s", e.what());
+        }
```
swichu91, line 386: I believe we have read-to-use functions for such operations(utils module) :)

swichu91 (approved):
LGTM, please check minor comments.

module-apps/application-settings/windows/bluetooth/BluetoothCheckPasskeyWindow.hpp
pholat: tbh - it's already shared in ApplicationCommon

PR #3386, opened 2021-12-08
[EGD-8083] Add wallpaper option

wojtekrzepecki:
Added connection of wallpaper window to settings

### DIFF
--- a/module-apps/application-settings/ApplicationSettings.cpp
+++ b/module-apps/application-settings/ApplicationSettings.cpp
@@ -441,7 +441,7 @@ namespace app
             return std::make_unique<gui::FontSizeWindow>(app);
         });
         windowsFactory.attach(gui::window::name::wallpaper, [](ApplicationCommon *app, const std::string &name) {
-            return std::make_unique<gui::WallpaperWindow>(app);
+            return std::make_unique<gui::WallpaperWindow>(app, static_cast<ApplicationSettings *>(app));
         });
         windowsFactory.attach(gui::window::name::quotes, [](ApplicationCommon *app, const std::string &name) {
             return std::make_unique<gui::QuotesMainWindow>(app);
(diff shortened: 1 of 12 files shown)


module-apps/application-settings/ApplicationSettings.cpp
```diff
@@ -692,6 +692,18 @@ namespace app
                            ::settings::SettingsScope::Global);
     }
 
+    auto ApplicationSettings::getWallpaperOption() -> gui::WallpaperOption
+    {
+        return static_cast<gui::WallpaperOption>(utils::getNumericValue<int>(
+            settings->getValue(::settings::Wallpaper::option, ::settings::SettingsScope::Global)));
```
PrzeBrudny, line 698: Well accordingly to settings guidelines it should be Local and probably handled by APM in that case (I do not agree with that -> for me it is ok).

PrzeBrudny (approved):

module-apps/application-settings/windows/display-keypad/WallpaperWindow.cpp
```diff
@@ -32,14 +44,11 @@ namespace gui
...
-
-        addCheckOption(utils::translate("app_settings_display_wallpaper_logo"), isWallpaperLogoSwitchOn);
-        addCheckOption(utils::translate("app_settings_display_wallpaper_clock"), isWallpaperClockSwitchOn);
-        addCheckOption(utils::translate("app_settings_display_wallpaper_quotes"), isWallpaperQuotesSwitchOn);
+                wallpaperOption == option ? gui::option::SettingRightItem::Checked
+                                          : gui::option::SettingRightItem::Disabled));
+        }
 
-        if (isWallpaperQuotesSwitchOn) {
+        if (wallpaperOption == WallpaperOption::Quote) {
```
pholat, line 51: this seems like part of model we are missing :)
generally, it might be nice to have one common model for lock windows and use its derivatives (or even itself) with two different windows.
Or even - by different settings, by model choose different views.
Saying that - I understand why it's not like that :)

pholat (approved):

module-apps/application-settings/windows/display-keypad/WallpaperWindow.cpp
wojtekrzepecki: Model added

PR #3388, opened 2021-12-08
[CP-830] Merge Backup and Restore tests

niepiekm:
Merged Backup and Restore tests


test/pytest/service-desktop/test_user_data_backup_restore.py
```diff
@@ -0,0 +1,268 @@
+# Copyright (c) 2017-2021, Mudita Sp. z.o.o. All rights reserved.
```
pholat, line 1: it's a valid code, could you please move it to repository https://github.com/mudita/QAMuditaOS and ask @filkamola for adoption there? otherwise, our QA won't run it, and after moving - there is a chance :)

pholat (approved):

test/pytest/service-desktop/test_user_data_backup_restore.py
```diff
@@ -0,0 +1,268 @@
...
+
+    log.debug("Requesting restore")
+    perform_restore(harness)
+
+    harness.connection.watch_port_reboot(20)
+    harness = get_harness_automatic(20)
+    harness.unlock_phone()
+
+    log.debug("Confirm user data on device match saved data")
+    compare_user_data(harness, totalContacts)
```
pawel-mudita, line 268: What about restoring the initial state of the DUT?
We want to leave DUT in the same state as it was before each test (eg. no extra user data should be left on the DUT).

niepiekm: Good point, fixing.

pawel-mudita (approved):

PR #3397, opened 2021-12-09
[BH-1264] Implement alarm fade in

tigransoghbatyan-digica:
Add fixed alarm volume fade in

### DIFF
--- a/products/BellHybrid/services/audio/ServiceAudio.cpp
+++ b/products/BellHybrid/services/audio/ServiceAudio.cpp
@@ -148,8 +148,34 @@ namespace service
         AudioStart(input);
         manageCpuSentinel();
 
+        if (playbackType == audio::PlaybackType::Alarm) {
+            startVolumeRamp(playbackType);
+        }
         return std::make_unique<AudioStartPlaybackResponse>(retCode, retToken);
     }
+
+    auto Audio::stopVolumeRamp()
+    {
+        if (volumeRampTimerHandle.isValid()) {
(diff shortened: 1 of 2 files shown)


products/BellHybrid/services/audio/ServiceAudio.cpp
```diff
@@ -243,20 +269,38 @@ namespace service
...
+
+        if (retCode == audio::RetCode::Success) {
+            settingsProvider->setValue(dbPath(setting, playbackType, profileType), std::to_string(clampedValue));
+        }
+        return std::make_unique<AudioResponseMessage>(retCode);
+    }
+
+    auto Audio::setVolume(audio::Volume volume) -> audio::RetCode
+    {
+        auto retCode = audio::RetCode::Success;
```
tomaszrybarski-digica, line 284: Shouldn't return code be initialized to some RetCode::Failed at first (and overwritten later by RetCode::Success when setting output volume succeeds)?

tigransoghbatyan-digica: true. Fixed

tomaszrybarski-digica (approved):

wojtekrzepecki (approved):

mjan-dev:
Closing this PR as the proposed solution has several flaws:
1. The ramp is updated every 1 second, therefore not providing acceptable fluency. 
2. It should be done on module-audio level to avoid operations associated with changing volume through service-audio (eg. database writes).

PR #3411, opened 2021-12-14
[EGD-1295] Fix deep press with light click

wojtekrzepecki:
light click should not be detected while deep press


PrzeBrudny (approved):

wojtekrzepecki:
Only drawback: 2x very fast light clicks (less than 500ms difference) will be not detected. Anyway the gui reacts a bit slower than this.

mjan-dev (approved):

PR #3419, opened 2021-12-15
[BH-1285] Onboarding tips language

paweljonskim:
Tips are displayed in selected (focused) language


tomaszrybarski-digica (approved):

mjan-dev (approved):
This solution doesn't cause lag when user scrolls the list?

paweljonskim:
> This solution doesn't cause lag when user scrolls the list?

no :) the only thing added to scrolling is to store the new value in `OnBoardingLanguageWindow` class. Language change (partial!) is done when the information popup is being created.

PR #3429, opened 2021-12-20
[CP-858] Fix test security

pawel-mudita:
Add security EP API to QAMuditaHarness repo.
Refactor security EP test to make it work.

### DIFF
--- a/test/pytest/service-desktop/test_security.py
+++ b/test/pytest/service-desktop/test_security.py
@@ -1,33 +1,82 @@
 # Copyright (c) 2017-2021, Mudita Sp. z.o.o. All rights reserved.
 # For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
+import logging
 
 import pytest
 import time
 
 from harness.interface.defs import status
+from harness.api.developermode import PhoneModeLock
+from harness.api.security import GetPhoneLockStatus, GetPhoneLockTime, SetPhoneLockOff
+from harness.request import TransactionError
+
+
+class SecurityTester:
+    def __init__(self, harness):
+        self.harness = harness
+
+    def confirm_phone_is_unlocked(self):
+        try:
+            GetPhoneLockStatus().run(self.harness)
+        except TransactionError:
+            return False
+        else:
+            return True
+
+    def confirm_phone_is_locked(self):
+        try:
+            GetPhoneLockStatus().run(self.harness)
+        except TransactionError as error:
+            return error.status == status["Forbidden"]
+        else:
+            return False
+
+    def check_if_passcode_format_validation_works(self, passcode: list):
+        try:
+            SetPhoneLockOff(passcode).run(self.harness)
+        except TransactionError as error:
+            return error.status == status["BadRequest"]
+        else:
+            return False
+
+    def confirm_no_time_lock(self):
+        try:
+            GetPhoneLockTime().run(self.harness)
+        except TransactionError as error:
+            return error.status == status["UnprocessableEntity"]
+        else:
+            return False
+
+    def unlock_phone(self, passcode: list = None):
+        try:
+            SetPhoneLockOff(passcode).run(self.harness)
+            time.sleep(0.1)
+        except TransactionError:
+            return False
+        else:
+            return True
 
 
 @pytest.mark.service_desktop_test
 @pytest.mark.usefixtures("phone_unlocked")
 def test_security_phone_unlocked(harness):
-    body = {"category": "phoneLockStatus"}
-
-    ret = harness.endpoint_request("usbSecurity", "get", body)
-    assert ret["status"] == status["NoContent"]
+    security_tester = SecurityTester(harness)
+    assert security_tester.confirm_phone_is_unlocked(), "Phone is locked, but should be unlocked!"
 
 
 @pytest.mark.service_desktop_test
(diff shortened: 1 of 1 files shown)


test/pytest/service-desktop/test_security.py
```diff
@@ -1,33 +1,73 @@
...
+        self.harness = harness
+
+    def assert_phone_unlocked(self):
+        assert GetPhoneLockStatus().run(self.harness).response.status == status["NoContent"]
+
+    def assert_phone_locked(self):
+        try:
+            GetPhoneLockStatus().run(self.harness).response.status
+        except TransactionError as error:
+            assert error.status == status["Forbidden"]
```
pholat, line 25: consider logging the error explaining why does it happen

test/pytest/service-desktop/test_security.py
```diff
@@ -1,33 +1,73 @@
...
+class SecurityTester:
+    def __init__(self, harness):
+        self.harness = harness
+
+    def assert_phone_unlocked(self):
+        assert GetPhoneLockStatus().run(self.harness).response.status == status["NoContent"]
+
+    def assert_phone_locked(self):
+        try:
+            GetPhoneLockStatus().run(self.harness).response.status
```
pholat, line 23: there are general fixtures for QAMuditaOS pytest- please change the usage there too.

test/pytest/service-desktop/test_security.py
```diff
@@ -1,33 +1,73 @@
...
 import time
 
 from harness.interface.defs import status
-
+from harness.api.developermode import PhoneModeLock
+from harness.api.security import GetPhoneLockStatus, GetPhoneLockTime, SetPhoneLockOff
+from harness.request import TransactionError
+
+
+class SecurityTester:
```
pholat, line 14: please either:
1. move asserts to test
2. make this class a test class

test/pytest/service-desktop/test_security.py
```diff
@@ -1,33 +1,73 @@
...
+            GetPhoneLockTime().run(self.harness)
+        except TransactionError as error:
+            assert error.status == status["UnprocessableEntity"]
+        else:
+            logging.error("Phone is time locked unexpectedly!")
+            assert False
+
+    def unlock_phone(self, passcode: list = None):
+        assert SetPhoneLockOff(passcode).run(self.harness).response.status == status["NoContent"]
+        time.sleep(0.5)
```
pholat, line 50: why is this sleep required?

pawel-mudita: On a physical device unlocking the phone takes (a short) time.
Without any sleep in this place, this operation could not be finished in the OS, before check after it.
0.1 or even less would be enough probably, but a bit more is reserved just in case.
Probably it could be combined with await and _assert_phone_locked_, but not sure if it's needed.
This solution is simple and reliable enough. The state of the lock is always checked after this operation, and if sleeping for 0.5 is not enough, this means that something is bad.
tl;dr
ihmo its reliable enough for this test

test/pytest/service-desktop/test_security.py
pawel-mudita: Do you mean fixtures from conftest.py, like _@pytest.mark.usefixtures("phone_locked")_?
They are using developer mode EP API, not security EP API -> no changes are needed

test/pytest/service-desktop/test_security.py
pawel-mudita: Actually, this is the expected status.
Just need to catch it here because some external class raises an exception on any status but _OK_

test/pytest/service-desktop/test_security.py
pholat: we have a lot of tests like that and in my opinion its a bad practice.
1. we are waiting for something which is a side-effect but user doesn't know what it is - we should wait for event explicitly
2. it can be slightly delayed and it will be issue, when it should be. Or it would be frequently used, but will happen much faster - but code will be super slow as there would be lots of sleep code in it.

changing this to something we could trust should be fairly easy. Unfortunatelly this is quite common pattern in our tests. I could agree to ie. helping you out changing it to something better (few hour top)?

test/pytest/service-desktop/test_security.py
pawel-mudita: Asserts moved to the test

test/pytest/service-desktop/test_security.py
pawel-mudita: Will be changed in the separate task: https://appnroll.atlassian.net/browse/CP-953

test/pytest/service-desktop/test_security.py
pholat: but we assert it here. so it's an expected failure? :)

pholat: please double check and close this comment :manbow:

pholat (approved):

test/pytest/service-desktop/test_security.py
pawel-mudita: Yes.
It is a failure just from the perspective of the external class, which raises an exception if the status is different than `OK`.
Here we expect the status to be `Forbidden`, since the phone should be locked when we use this method.
See updated code, now assertion is moved to test. It should be clear now.

wojtekrzepecki (approved):

PR #3440, opened 2021-12-21
EGD-8066 assets split

pholat:
Minimal changes to load assets from external repository and build images for different targets independently
With these:
- we are able to have different assets for tests and actual image
- we are able o have different assets for Bell and Pure - as these have own listings
- we have better dependencies on packages

We do not yet have:
- listings from release branches

Saying that:
PR is ready for review, just listings in assets.json will change

comparison of mounted images for rt1051 for current master and the branch for Pure - please see image assets for bell missing in image (as desired)
image


cmake/modules/AddPackage.cmake
```diff
@@ -65,16 +66,15 @@ function(add_update_package SOURCE_TARGET)
...
-        DEPENDS ecoboot.bin-target
-        DEPENDS updater.bin-target
-        DEPENDS assets
-        COMMAND ${CMAKE_SOURCE_DIR}/tools/generate_update_image.sh ${SOURCE_TARGET} ${PROJECT_VERSION} ${CPACK_SYSTEM_NAME}
+                ${SOURCE_TARGET}-boot.bin
+                ${SOURCE_TARGET}-version.json-target
+                ecoboot.bin-target
+                updater.bin-target
+                assets
+                COMMAND ${CMAKE_SOURCE_DIR}/tools/generate_update_image.sh ${SOURCE_TARGET} ${PROJECT_VERSION} ${CPACK_SYSTEM_NAME}
```
SP2FET, line 75: wrong indentation

cmake/modules/Assets.cmake
```diff
@@ -4,33 +4,36 @@ function(add_assets_target)
...
         ${_ASSETS_TARGET}
-        DEPENDS json-target
+        DEPENDS ${_ASSETS_DEPENDS}
         COMMAND mkdir -p ${_ASSETS_DEST_DIR}
         COMMAND rsync -qravu --delete
             ${_ASSETS_SOURCE_DIR}/.boot.json*
             ${_ASSETS_SOURCE_DIR}/personalization.json
             ${_ASSETS_DEST_DIR}
-        COMMAND rsync -qravu --delete
+        COMMAND rsync -qravu
```
SP2FET, line 27: why without deleting?

products/PurePhone/assets.json
```diff
@@ -2,73 +2,74 @@
...
             "name": "./fonts/common/gt_pressura_light_46.mpf",
             "ref": "10c74fcb09c2022325767cad735c0183b6f5393a",
-            "output": "image/assets/fonts/gt_pressura_light_46.mpf"
+            "output": "assets/fonts/gt_pressura_light_46.mpf"
         },
         {
             "name": "./fonts/common/gt_pressura_light_90.mpf",
             "ref": "10c74fcb09c2022325767cad735c0183b6f5393a",
-            "output": "image/assets/fonts/gt_pressura_light_90.mpf"
+            "output": "assets/fonts/gt_pressura_light_90.mpf"
```
SP2FET, line 67: did you ensure yourself that the path is correct? :P

SP2FET (approved):

cmake/modules/Assets.cmake
pholat: this is because of what this target actually does is copying stuff from `image` catalog in `proper` places
now we have target which copies downloaded elements directly to build catalog and its set as a dependency to this `asset` target to keep it like that:
```
download & generation of assets  <- assets <- creation of image
```

but if `assets` target removes `assets` directory where we already put data so that it would end up in image we won't have it in it - without any error risen... 
now you would probably as "Ok, then why don't we copy assets to `image` catalog as we did.
This is because we can't do it. If we did - we won't have properly built packages, as we are copying whole directories here - we will copy all files: then image generated for tests will have both assets for tests and assets for i.e. bell and we do not want that.

At best I would rather remove all of that rsync stuff, then generate different `asset` targes (one for tests, one for bell, one for pure one for pure with developer data) just aggregating its dependencies. This way we wouldn't have to i.e. have hacks in line 35 or weird `if's` for LUTS - as different configurations would result in a different set of data copied.. That would be out of scope here as I do not want to refactor our cmake code too much.

products/PurePhone/assets.json
pholat: yes, but I'll double-check it with the updated assets list for release images.
The best way to verify is to compare build images structure (mounting similar packages with losetup and mount) - then if it's matching, check on hardware :)

PrzeBrudny (approved):

SP2FET:
BTW is the harness update on purpose?

image/CMakeLists.txt
```diff
@@ -19,10 +20,10 @@ multicomp_install(
...
-    DESTINATION ./
-    COMPONENTS Standalone Update
-)
+# multicomp_install(
+#     DIRECTORY
+#         ${ASSETS_DEST_DIR}/current/
+#         ${ASSETS_DEST_DIR}/user
+#     DESTINATION ./
+#     COMPONENTS Standalone Update
+# )
```
swichu91, line 29: Why was it commented out? Remove it completely or at least leave a comment explaining why.

products/BellHybrid/CMakeLists.txt
```diff
@@ -77,20 +77,37 @@ strip_executable(BellHybrid)
...
 
 include(BinaryAssetsVersions.cmake)
 include(DownloadAsset)
-download_asset_json(${CMAKE_CURRENT_SOURCE_DIR}/assets.json ${PROJECT_SOURCE_DIR} ${MUDITA_CACHE_DIR})
-download_asset_release(PureUpdater_RT.bin updater.bin PureUpdater ${UPDATER_BIN_VERSION})
-download_asset_release(ecoboot.bin ecoboot.bin ecoboot ${ECOBOOT_BIN_VERSION})
+# copy all assets required to build catalog under current folder as in json recipe
+download_asset_json( json-target
+                     ${CMAKE_CURRENT_SOURCE_DIR}/assets.json
+                     # ${CMAKE_SOURCE_DIR}
```
swichu91, line 96: Remove it?

products/BellHybrid/CMakeLists.txt
```diff
@@ -77,20 +77,37 @@ strip_executable(BellHybrid)
...
+                     ${CMAKE_BINARY_DIR}/sysroot/sys/current/
+                     MuditaOSAssets
+                     ${MUDITA_CACHE_DIR}
+    )
+# copy all assets required to build catalog under current folder as in json recipe
+download_asset_release_json(    json-images-target
```
swichu91, line 102: Can it be possibly merged with the `download_asset_json`?

products/BellHybrid/assets_images.json
```diff
@@ -0,0 +1,216 @@
+{
+    "assets": [
+        {"name": "release2.tgz", "tarfile" :"image/assets/images/music-player/mp_note_W_G.vpi", "output": "assets/images/music-player/mp_note_W_G.vpi"},
```
swichu91, line 3: Is it a tarfile or just an input file?

tools/download_asset.py
```diff
@@ -73,105 +75,255 @@ def getToken():
...
         '''
-        path = self.api.repos.get_content(**data)
+        # TODO add exception here if file not found!!!!
+        try:
+            path = self.api.repos.get_content(**data)
+        except HTTP404NotFoundError:
+            raise RuntimeError(f'file {data["path"]} under id {data["ref"]} not found')
         with where.open("wb") as f:
-            # TODO in generall - content says what encoding is used, this is simplification
+            # NOTE in generall - content says what encoding is used, this is simplification
```
swichu91, line 287:
```suggestion
            # NOTE in general - content says what encoding is used, this is a simplification
```

tools/download_asset.py
```diff
@@ -257,14 +402,32 @@ def download_release(self, repo, name_in, name_out, version, product):
...
+        then installs it under `name_out`
+        '''
+        release = self._get_lousy_match_release(repo, product, version)
+        var = Verse(name_in, name_out, sha_gen=lambda x: self._get_file_in_release(release.assets, x.name))
+        self.download_file_from_release(var)
+
+    def download_release_json(self, repo, j, version, product):
+        # [x] download to cache - ready (in _get_file_in_release)
+        # [-] unpack to cache
+        # [x] copy/unpack where needed
```
swichu91, line 426: Are these some kind of developer's notes? If yes, how about removing it?

swichu91 (approved):

tools/download_asset.py
pholat: tx

products/BellHybrid/assets_images.json
pholat: it's the filename in the tar file. As this script uses `name` as input name - which is `release2.tgz` here, and output would be `mp_note_W_G.vpi` we have to know what's the path of this file in the `name` file :)

products/BellHybrid/CMakeLists.txt
pholat: it could be possibly done, but would require yet some more changes in the `json` description - as right now we pass repository as an argument - and we would need it in the `json` file. Not too hard to do - not required for these changes.
Here we have 2 repositories:
1. public one, with images etc
2. non public one: with luts.bin and fonts.
and it represents this split

image/CMakeLists.txt
pholat: ti was temporarily commented out, it's removed :) similarly in the comment below ;) tx

products/BellHybrid/assets_images.json
swichu91: That does make sense ;) How about this? 
```suggestion
        {"tarfile": "release2.tgz", "input" :"image/assets/images/music-player/mp_note_W_G.vpi", "output": "assets/images/music-player/mp_note_W_G.vpi"},
```

swichu91: Feel free to close this conversation ;)

pholat: sounds leigt, I would have to have more complicated json parsing and it was meant to be simple :) ( btw. if we drop `tarfile` which is a bad name, then have `name` + `input` -> `output` this should be ok? )

PR #3441, opened 2021-12-21
[CP-856] Refactor test contacts

pawel-mudita:
Refactor contacts EP test to make it work again.
Rewritten with use of contacts EP API moved to harness submodule.

### DIFF
--- a/test/pytest/service-desktop/test_contacts.py
+++ b/test/pytest/service-desktop/test_contacts.py
@@ -1,146 +1,357 @@
-# Copyright (c) 2017-2021, Mudita Sp. z.o.o. All rights reserved.
+# Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
 # For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 import pytest
 from harness.interface.defs import status
+from harness.request import TransactionError
+from harness.api.contacts import GetContactsCount, GetContactsWithOffsetAndLimit, GetContactById, AddContact, \
+    UpdateContact, DeleteContactById
+
+
+class ContactsTester:
+    def __init__(self, harness):
+        self.harness = harness
+
+    def get_contacts_count(self):
+        try:
+            count = GetContactsCount().run(self.harness).count
+        except TransactionError:
+            return False
+        else:
+            return True, count
+
+    def get_contacts_with_offset_and_limit(self, offset, limit):
+        try:
+            result = GetContactsWithOffsetAndLimit(offset, limit).run(self.harness)
+        except TransactionError:
+            return False
+        else:
+            return True, result.entries, result.totalCount
+
+    def get_contact_by_id(self, contact_record_id):
+        try:
+            result = GetContactById(contact_record_id).run(self.harness)
+        except TransactionError:
+            return False
+        else:
+            return True, result.entry
+
+    def add_contact(self, contact_record):
+        try:
+            contact_id = AddContact(contact_record).run(self.harness).id
+        except TransactionError:
+            return False
+        else:
+            return True, contact_id
+
+    def detect_adding_duplicated_contact(self, contact_record):
+        try:
+            AddContact(contact_record).run(self.harness)
+        except TransactionError as error:
+            if error.status == status["Conflict"]:
+                return True
+        else:
+            return False
+
+    def update_contact(self, contact_record):
+        try:
+            UpdateContact(contact_record).run(self.harness)
+        except TransactionError:
+            return False
+        else:
+            return True
+
+    def delete_contact_by_id(self, contact_record_id):
+        try:
+            DeleteContactById(contact_record_id).run(self.harness)
+        except TransactionError:
+            return False
+        else:
+            return True
 
 
 @pytest.mark.service_desktop_test
 @pytest.mark.usefixtures("phone_unlocked")
-def test_contacts(harness):
-    # getting the contacts count
-    body = {"count": True}
-    ret = harness.endpoint_request("contacts", "get", body)
-    assert ret["status"] == status["OK"]
+def test_add_single_contact(harness):
+    # Check initial state
+    contacts_tester = ContactsTester(harness)
+    result, initial_number_of_contact_records = contacts_tester.get_contacts_count()
+    assert result, "Failed to get contacts count!"
 
-    count = ret["body"]["count"]
-    if count == 0:
-        pytest.skip("No contacts entries, skipping")
+    # Add contact
+    contact_record = {
+        "address": "6 Czeczota St.\n02600 Warsaw",
+        "altName": "Smith",
+        "email": "john.smith@mudita.com",
+        "blocked": False,
+        "favourite": False,
+        "ice": False,
+        "numbers": [
+            "123456789"
+        ],
+        "speedDial": "1",
+        "priName": "John",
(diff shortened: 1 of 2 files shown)


niepiekm (approved):

mjan-dev (commented):
Maybe if we are doing refactor of those tests, we could separate them into more test cases? Or is it justified decision to keep it as one test for contacts?

pholat (approved):

test/pytest/service-desktop/test_contacts.py
```diff
@@ -1,146 +1,357 @@
...
+
+    result, contact_id = contacts_tester.add_contact(contact_record)
+    assert result, "Failed to add contact!"
+    result, retrieved_contact_record = contacts_tester.get_contact_by_id(contact_id)
+    assert result, "Failed to get contact by ID!"
+    del retrieved_contact_record["id"]
+    assert contact_record == retrieved_contact_record
+
+    # Clean up
+    assert contacts_tester.delete_contact_by_id(contact_id), "Failed to delete a contact!"
```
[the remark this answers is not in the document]
filkamola: In pytest, if you need some actions to happen before or after (i.e. cleanup) the execution of the test case, you can use fixtures and reuse them in multiple test cases later on.

test/pytest/service-desktop/test_contacts.py
```diff
@@ -1,146 +1,357 @@
...
+            "123456799"
+        ],
+        "speedDial": "1",
+        "priName": "John",
+        "note": "Some note"
+    }
+    assert contacts_tester.update_contact(edited_contact_record), "Failed to update contact!"
+    result, retrieved_contact_record = contacts_tester.get_contact_by_id(contact_id)
+    assert result, "Failed to get contact by ID!"
+    assert retrieved_contact_record == edited_contact_record
```
filkamola, line 193: I really like that you add failure messages to your asserts, but why stop here? It would be best to give this one a message too so that you can easily see what went wrong on the report.

test/pytest/service-desktop/test_contacts.py
```diff
@@ -1,146 +1,357 @@
...
+    assert returned_contacts[1] == second_contact_record
+    assert returned_contacts[2] == first_contact_record
+    assert returned_contacts[3] == third_contact_record
+    assert current_number_of_contact_records == total_count
+
+    # Get previously added contacts without 2 of them (test offset)
+    result, returned_contacts, total_count = contacts_tester.get_contacts_with_offset_and_limit(
+        initial_number_of_contact_records + 2, current_number_of_contact_records - initial_number_of_contact_records)
+    assert result, "Failed to get contacts with offset and limit!"
+
```
filkamola, line 328: You are reusing this loop and asserts so many times that you might as well write a method for it and use it in the test case. This way, this code will be much more readable and you will be able to change the number of contacts if necessary instead of hard coding it in the scenario.

test/pytest/service-desktop/test_contacts.py
```diff
@@ -1,146 +1,357 @@
...
+    # Clean up
+    assert contacts_tester.delete_contact_by_id(contact_id), "Failed to delete a contact!"
+    result, received_contact_records_count = contacts_tester.get_contacts_count()
+    assert result, "Failed to get contacts count!"
+    assert initial_number_of_contact_records == received_contact_records_count
+
+
+@pytest.mark.service_desktop_test
+@pytest.mark.usefixtures("phone_unlocked")
+def test_add_and_get_multiple_contacts(harness):
```
filkamola, line 204: This test is very long, which in itself is nothing wrong but the amount of possible failure points (see the amount of asserts for example) is so big, that maintenance of this test will be a nightmare. Imagine it fails, and you see it on the report. What information are you getting at this point? Not very informative since any of those multiple asserts could've been the reason.

test/pytest/service-desktop/test_contacts.py
filkamola, line 193: I really like that you add failure messages to your asserts, but why stop here? It would be best to give this one a message too so that you can easily see what went wrong on the report.

filkamola:
I added my 2 cents to this. I only commented on a given issue once even though it may occur many times in the code. I think it will be obvious to fix the other occurrences as well :)

test/pytest/service-desktop/test_contacts.py
pawel-mudita: I agree that this test is very long.
Unfortunately, it seems it has to be like this at this moment.
I could change this test to 4 separate tests: 
1. test adding multiple contacts
2. test getting multiple contacts
3. test limit
4. test offset.
But then, to test 4, I need to copy almost all code from the previous test to it.
Same for 3, almost all code from tests 1 and 2 should be copied to it.
Overall, we end up with 4 tests instead of one, but two of them are almost that big, like the one we divided.

However, I see one solution for this.
If we could have prepared DB files with proper test data, we could simplify this (and not only) test by a lot.
We could utilize a backup and restore mechanism for this. We could also address the clean-up problem with this method.

This potential improvement should be considered in the next iteration of improving this, and other tests.

test/pytest/service-desktop/test_contacts.py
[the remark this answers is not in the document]
pawel-mudita: I see, I wish I know this before I started to work on these tests.
This cleanup really depends on context, so it is not so obvious to make it a fixture.
However, I have an idea how to improve this in the next iteration of reworking these scripts - I've described it in another comment.

test/pytest/service-desktop/test_contacts.py
pawel-mudita: Added missing failure messages.

test/pytest/service-desktop/test_contacts.py
```diff
@@ -1,146 +1,357 @@
...
+@pytest.mark.service_desktop_test
+@pytest.mark.usefixtures("phone_unlocked")
+def test_add_single_contact(harness):
+    # Check initial state
+    contacts_tester = ContactsTester(harness)
+    result, initial_number_of_contact_records = contacts_tester.get_contacts_count()
+    assert result, "Failed to get contacts count!"
+
+    # Add contact
+    contact_record = {
```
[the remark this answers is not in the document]
pawel-mudita: IMO the way we store and use test data should be rethought and then we should apply the solution to all these kinds of tests. This is a nice idea, we should keep it in mind during the next iteration of tests improvement (maybe during moving them to QA repo?).

test/pytest/service-desktop/test_contacts.py
pawel-mudita: Actually, the lines you are pointing are not repeated, they just look similar in two other places.
But of course, I see some space for improvements and optimization here in upcoming iterations.

pawel-mudita:
> I added my 2 cents to this. I only commented on a given issue once even though it may occur many times in the code. I think it will be obvious to fix the other occurrences as well :)

Thank you for your input, it's very valuable.
Too bad I didn't know all of this a few weeks ago before I started work on these tests.
I've made some small improvements, but the bigger ones I leave for the next iteration of improvements.
The main purpose of this PR was to fix the non-working test of contact EP API, and update it with the use of API moved to harness repository. Tests are working, we need them to everyday work, so they need to be merged.
Nevertheless, further improvements should be done, and all that tips will help to make these tests better.

PR #3444, opened 2021-12-22
[BH-1340] Fix audio asset tags

swichu91:
Fixed audio asset tags


module-services/service-cellular/ServiceCellular.cpp
```diff
@@ -1960,22 +1960,22 @@ auto ServiceCellular::handleCellularGetFirmwareVersionMessage(sys::Message *msg)
...
-    auto message = static_cast<sevm::StatusStateMessage *>(msg);
-        auto status_pin = message->state;
-        if (priv->modemResetHandler->handleStatusPinEvent(status_pin == value::ACTIVE)) {
-            return std::make_shared<CellularResponseMessage>(true);
-        }
+    auto message    = static_cast<sevm::StatusStateMessage *>(msg);
+    auto status_pin = message->state;
+    if (priv->modemResetHandler->handleStatusPinEvent(status_pin == value::ACTIVE)) {
+        return std::make_shared<CellularResponseMessage>(true);
+    }
```
pholat, line 1967: why was it miss formatted? is it not checked with formatting git hook? or smbd forced it?

pholat (approved):
+1

module-services/service-cellular/ServiceCellular.cpp
swichu91: Good question. I suppose it was simply pushed without running a check (is it even possible?).

SP2FET (approved):

PR #3451, opened 2021-12-27
[BH-1343] Fix frontlight idle exit

paweljonskim:
Fix frontlight idle exit
The frontlight value will be reverted

### DIFF
--- a/products/BellHybrid/apps/application-bell-powernap/presenter/PowerNapProgressPresenter.cpp
+++ b/products/BellHybrid/apps/application-bell-powernap/presenter/PowerNapProgressPresenter.cpp
@@ -76,7 +76,7 @@ namespace app::powernap
     void PowerNapProgressPresenter::onNapFinished()
     {
         if (alarmLightOnOffModel->getValue()) {
-            frontLightModel.restorePreviousState();
+            frontLightModel.revertUnsavedChanges();
             frontLightModel.setStatus(true);
         }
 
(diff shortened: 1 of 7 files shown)


SP2FET (approved):

pawel-mudita (approved):

products/BellHybrid/apps/application-bell-settings/presenter/FrontlightPresenter.cpp
```diff
@@ -16,6 +16,11 @@ namespace app::bell_settings
         };
     }
 
+    FrontlightPresenter::~FrontlightPresenter()
+    {
+        exitWithoutSave();
+    }
+
```
swichu91, line 23: Are you sure this will work? Imho, the destructor will be invoked even if a user performed full procedure(thus accepted to save the new value) and set the just set value to the default(old) one.

swichu91 (commented):
Please check the comment.

products/BellHybrid/apps/application-bell-settings/presenter/FrontlightPresenter.cpp
paweljonskim: you are right! check now

products/BellHybrid/apps/application-bell-powernap/presenter/PowerNapProgressPresenter.cpp
```diff
@@ -76,7 +76,7 @@ namespace app::powernap
     void PowerNapProgressPresenter::onNapFinished()
     {
         if (alarmLightOnOffModel->getValue()) {
-            frontLightModel.restorePreviousState();
+            frontLightModel.revertUnsavedChanges();
```
tomaszrybarski-digica, line 79: Worth testing if frontlight will still be turned on after power nap ends - i cannot do it on my Dev Kit P1 since the switches do not work

PR #3456, opened 2022-01-04
[BH-1353] Invalid snooze msg

swichu91:
Fixed invalid snooze msg.

### DIFF
--- a/products/BellHybrid/apps/common/src/AlarmModel.cpp
+++ b/products/BellHybrid/apps/common/src/AlarmModel.cpp
@@ -110,13 +110,13 @@ namespace app
         }
         return alarmEventPtr->enabled;
     }
-    std::uint32_t AlarmModel::getSnoozeDuration()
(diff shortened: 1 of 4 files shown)


products/BellHybrid/apps/common/src/AlarmModel.cpp
```diff
@@ -110,13 +110,13 @@ namespace app
         }
         return alarmEventPtr->enabled;
     }
-    std::uint32_t AlarmModel::getSnoozeDuration()
+    std::chrono::seconds AlarmModel::getSnoozeDuration() const
```
tomaszrybarski-digica, line 113: Shouldn't std::chrono::minutes be the return value here?

swichu91: Actually, it was the source of the bug. Snooze duration is stored using minutes but layers that use this method require std::time_t which is in seconds.

tomaszrybarski-digica (approved):

mjan-dev (approved):

PR #3458, opened 2022-01-04
[EGD-8129] Add atexit functionality

alekrudnik:
Added atexit functionality. All global destructors and functions
registered with atexit() should be called at program exit.
It was possbile to reuse original newlib implementation with 
MALLOC_PROVIDED flag set.

Made sure logger destructructor is called as last one.

Due to mudita OS legacy, it was not possible to simply fix dependencies
in global objectes and hence there are neither global objects destructors
nor functions registered with atexit() called when exit() is called.
Possibly it will be changed later.

### DIFF
--- a/module-sys/SystemManager/SystemManagerCommon.cpp
+++ b/module-sys/SystemManager/SystemManagerCommon.cpp
@@ -1,4 +1,4 @@
-﻿// Copyright (c) 2017-2021, Mudita Sp. z.o.o. All rights reserved.
+﻿// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
 // For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 #include <SystemManager/SystemManagerCommon.hpp>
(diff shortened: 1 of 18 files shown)


lucckb (approved):

wojtekrzepecki (approved):

module-sys/SystemManager/SystemManagerCommon.cpp
```diff
@@ -138,7 +138,7 @@ namespace sys
         case SystemManagerCommon::State::Running:
         case SystemManagerCommon::State::Suspend:
         case SystemManagerCommon::State::Shutdown:
-            exit(1);
+            abort();
```
pholat, line 141: why is it abort now?

pholat (approved):
cool :+1:

module-sys/SystemManager/SystemManagerCommon.cpp
alekrudnik: After fixing of calling global destructors at exit(), each exit() call was finished with HF.
Previously I tried to remove all call to exit, but later decided to fix exit() no to clean up.

I will revert it back as at this moment it doesnt matter.

PR #3459, opened 2022-01-04
[CP-855] Refactor test_threads

pawel-mudita:
Refactored threads part of messages endpoint tests.
Rewritten with use of threads EP API from harness.


test/pytest/service-desktop/test_threads.py
```diff
@@ -1,45 +1,130 @@
-# Copyright (c) 2017-2021, Mudita Sp. z.o.o. All rights reserved.
+# Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
```
pholat, line 1: consider renaming to message threads from threads, title might be misleading - as one can think i.e. about software threads

test/pytest/service-desktop/test_threads.py
```diff
@@ -1,45 +1,130 @@
...
+    message_1_body = "Test message nr 1"
+    message_2_body = "Test message nr 2"
+    message_3_body = "Test message nr 3"
+    message_4_body = "Test message nr 4"
+    message_5_body = "Test message nr 5"
+    message_6_body = "Test message nr 6"
+
+    threads_tester = ThreadsTester(harness)
+
+    result, message_1_record = threads_tester.add_message(message_number_1, message_1_body)
```
pholat, line 90: why copy instead of for loop?

test/pytest/service-desktop/test_threads.py
wojtekrzepecki: :+1:

wojtekrzepecki (approved):

test/pytest/service-desktop/test_threads.py
SP2FET: as @pholat mentioned, this could be wrapped in a loop

test/pytest/service-desktop/test_threads.py
```diff
@@ -1,45 +1,130 @@
...
+    assert result, "Failed to add message!"
+
+    result, response = threads_tester.get_threads_with_offset_and_limit(
+        0, 0)
+    assert result, "Failed to get all threads!"
+    total_count_of_threads = response.totalCount
+    result, response = threads_tester.get_threads_with_offset_and_limit(
+        total_count_of_threads - 6, 6)
+    assert result, "Failed to get threads with offset and limit!"
+    THREADS_PAGE_SIZE = 4
```
SP2FET, line 110: this one could be moved a little bit higher in the code, to be more readable

SP2FET (approved):

test/pytest/service-desktop/test_threads.py
pawel-mudita: Renamed

test/pytest/service-desktop/test_threads.py
pawel-mudita: Loop added

PR #3466, opened 2022-01-10
[EGD-8080] Added 4 hour deep refresh on locked screen

PrzeBrudny:
Added 4 hour deep refresh on locked screen.

### DIFF
--- a/module-apps/apps-common/ApplicationCommon.cpp
+++ b/module-apps/apps-common/ApplicationCommon.cpp
@@ -927,8 +927,8 @@ namespace app
 
     bool ApplicationCommon::isCurrentWindow(const std::string &windowName) const noexcept
     {
-        if (const auto &window = windowsStack().get(windowName); window != nullptr) {
-            return window->getName() == windowName;
+        if (const auto &window = windowsStack().get(topWindow); window != std::nullopt) {
+            return window == windowName;
         }
         LOG_ERROR("no window: %s", windowName.c_str());
         return false;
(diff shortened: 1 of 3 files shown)


mjan-dev (approved):

module-apps/apps-common/ApplicationCommon.cpp
```diff
@@ -927,7 +927,8 @@ namespace app
 
     bool ApplicationCommon::isCurrentWindow(const std::string &windowName) const noexcept
     {
-        if (const auto &window = windowsStack().get(windowName); window != nullptr) {
+        auto topWindowName = windowsStack().get(topWindow);
+        if (const auto &window = windowsStack().get(*topWindowName); window != nullptr) {
             return window->getName() == windowName;
```
pholat, line 932:
```   
 bool ApplicationCommon::isCurrentWindow(const std::string &windowName) const noexcept
    {
        if (const auto &window = windowsStack().get(topWindow); window != std::nullopt) {
            return window == windowName;
        }
        LOG_ERROR("no window: %s", windowName.c_str());
        return false;
    }
```
get(...str) in this case returns std::nullopt if there is no window. Therefore: 1. you get name of the top window 2. you check if it's current window. 
I know it's my mistake in this place, but it would be best to change it for good :)

pholat (approved):
plz change the isCurrentWindow and then it's ok :)

module-apps/apps-common/ApplicationCommon.cpp
PrzeBrudny: sure :)

PR #3467, opened 2022-01-11
[EGD-8141] Fix brownout handling while charging

Maciej-Mudita:
When the voltage dropped below the brownout threshold,
the system did not detect the battery charging
and displayed a window with the low battery icon all the time.

### DIFF
--- a/module-bsp/board/rt1051/puretx/hal/battery_charger/BatteryCharger.cpp
+++ b/module-bsp/board/rt1051/puretx/hal/battery_charger/BatteryCharger.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2021, Mudita Sp. z.o.o. All rights reserved.
+// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
 // For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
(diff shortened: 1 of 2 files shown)


kkleczkowski (approved):

module-bsp/board/rt1051/puretx/hal/battery_charger/BatteryCharger.cpp
```diff
@@ -73,7 +81,7 @@ namespace hal::battery
                 bsp::battery_charger::clearFuelGuageIRQ(
                     static_cast<std::uint16_t>(bsp::battery_charger::batteryINTBSource::minVAlert));
 
-                eventsHandler.onBrownout();
+                eventsHandler.onBrownout(bsp::battery_charger::getChargeStatus());
```
wojtekrzepecki, line 84: Wouldn't be better to include all the implemented logic here? (BatteryCharger.cpp for pure). Brownout detector and interface of BatteryCharger does not have to know about existence of fuel gauge IRQs - this is more low level concept. Also Bell have different logic for different chips.

Maciej-Mudita: done

wojtekrzepecki (approved):

PR #3476, opened 2022-01-13
[EGD-6824] Simplified notifications on quote wallpaper

wojtekrzepecki:
simplified home screen notifications as tiles on phone locked wallpaper

### DIFF
--- a/module-apps/apps-common/popups/presenter/WallpaperPresenter.cpp
+++ b/module-apps/apps-common/popups/presenter/WallpaperPresenter.cpp
@@ -2,58 +2,91 @@
 // For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 #include "WallpaperPresenter.hpp"
-#include <service-appmgr/messages/GetWallpaperOptionRequest.hpp>
 #include <service-appmgr/Constants.hpp>
 
 namespace gui
 {
-    WallpaperPresenter::WallpaperPresenter(app::ApplicationCommon *app) : AsyncCallbackReceiver(app), application(app)
+    WallpaperPresenter::WallpaperPresenter(app::ApplicationCommon *app)
+        : AsyncCallbackReceiver(app), application(app),
+          notificationsModel(std::make_shared<NotificationsModel>(NotificationsListPlacement::LockedScreen))
     {}
 
-    void WallpaperPresenter::attachWallpapers(std::shared_ptr<WallpaperBase> clockWallpaper,
-                                              std::shared_ptr<WallpaperBase> quoteWallpaper,
-                                              std::shared_ptr<WallpaperBase> logoWallpaper)
+    void WallpaperPresenter::setupWallpaper(Item *parent)
     {
-        this->clockWallpaper = clockWallpaper;
-        this->quoteWallpaper = quoteWallpaper;
-        this->logoWallpaper  = logoWallpaper;
-    }
-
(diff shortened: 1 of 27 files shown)


wojtekrzepecki:
obraz
obraz

module-apps/apps-common/notifications/NotificationTilesModel.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+#include "NotificationTilesModel.hpp"
+#include <Image.hpp>
+#include <Style.hpp>
+
+using namespace gui;
+
+namespace
+{
+    std::map<notifications::NotificationType, UTF8> typeToIcon{
```
pholat, line 12: 1. isn't pretty
2. shouldn't be const?
3. if it would be const? can it be constexpr - there should be nothing dynamic in known int range mapping to known string?

module-apps/apps-common/notifications/NotificationTilesModel.cpp
```diff
@@ -0,0 +1,79 @@
...
+{
+    if (tilesBox == nullptr) {
+        LOG_ERROR("TilesBox not attached");
+        return;
+    }
+
+    tilesBox->erase();
+
+    for (const auto &notification : params->getNotifications()) {
+        if (typeid(*notification) == typeid(notifications::NotSeenSMSNotification) && callAndSMSVisibility) {
```
pholat, line 49: I don't think it's MVP (or MVC to be honest :) ) now 
good crib: [link](http://1.bp.blogspot.com/-8751o9FiHdE/Vp1V7xzOz7I/AAAAAAAAAtk/GAfrW9lvT2U/s1600/mvp.png)

module-apps/apps-common/notifications/NotificationTilesModel.cpp
wojtekrzepecki: Changed to constexprs

module-apps/apps-common/notifications/NotificationTilesModel.cpp
wojtekrzepecki: I know that it may not be clear at first glance but I had to make some tradeoffs to achieve:
- one notifications model to support multiple views (both listView and tiles)
- one tiles notification model to drive multiple instances of NotificationTilesBox (HBox) - switchable dynamically

wojtekrzepecki: Renamed to provide MVP structure:
model (NotificationsModel) <> presenter (Notifications tiles/list presenter) <> View (ListView, NotificationTiles)

pholat (approved):
good enough :)

module-apps/apps-common/popups/presenter/WallpaperPresenter.cpp
```diff
@@ -2,58 +2,91 @@
...
+            break;
+        }
+    }
+
+    void WallpaperPresenter::buildWallpapers(Item *parent)
     {
-        clockWallpaper->build();
-        quoteWallpaper->build();
-        logoWallpaper->build();
+        clockWallpaper = std::make_shared<WallpaperClock>(parent);
```
PrzeBrudny, line 57: i would recommend moving that to WallpaperModel -> and just access here. Presenter is not best place to hold widgets in my opinion. It should only control them.

PrzeBrudny (approved):

PR #3481, opened 2022-01-13
[EGD-7359] Remove 0 connection frequency for message only mode

pawel-mudita:
When Offline/Message only mode is set, Pure should log into
the network to send and download messages based on
connection frequency. The default option was set to 0, which
meant that the message-only mode was not working at all.

image

### DIFF
--- a/image/user/db/settings_bell_002.sql
+++ b/image/user/db/settings_bell_002.sql
@@ -1,4 +1,4 @@
--- Copyright (c) 2017-2021, Mudita Sp. z.o.o. All rights reserved.
+-- Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
 -- For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 -- ----------- insert default values ----------------------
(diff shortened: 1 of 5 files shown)


pholat:
this is a setting value and question is: will it be updated in the setting (settings_v2 changed) during the update?

pawel-mudita:
> this is a setting value and question is: will it be updated in the setting (settings_v2 changed) during the update?

I have no idea, how to check it?

wojtekrzepecki:
It will not change it. It will create the key with default value if it not exists yet. You can check it just by doing update with newly created software.

wojtekrzepecki:
@mjan-dev  Maybe It is good moment to start discussion about database migration strategies

PrzeBrudny (approved):

niepiekm (approved):

niepiekm (approved):

pholat:
https://github.com/mudita/MuditaOS/pull/3481#issuecomment-1012927123
https://github.com/mudita/MuditaOS/pull/3481#issuecomment-1013158316

This is exactly why I'm asking - we change this DB-related code but values already exist. This way - it won't be fixed for people who do not fully override memory, which should be the majority of people using phones I suppose?

pholat (changes requested):
In my opinion, it should be double-checked

mjan-dev (commented):
As mentioned in comments. These changes doesn't fix the issue for already shipped phones, especially that version for continued produced has already been released. A basic policy for database migration will be planned shortly. To address this PRs issue, maybe we could add a new value to db. It is not pretty at all, however I can't think of less harmful solution in this situation. @wojtekrzepecki @pholat @pawel-mudita Please comment.

wojtekrzepecki:
Yes, it is as option to mitigate the issue.

pawel-mudita:
Changed key name in DB, which should help to resolve this issue also for phones already shipped.

mjan-dev (approved):
Accepting the solution as other solution would require mayor changes.

PR #3502, opened 2022-01-20
EGD-8170

pholat:
crashes fix


.gdbinit
```diff
@@ -1,4 +1,3 @@
 source .gdb_macros
 
-handle all nostop pass
```
mjan-dev, line 3: why this is removed?

module-apps/apps-common/windows/OptionWindow.hpp
```diff
@@ -31,5 +31,11 @@ namespace gui
         void onClose(CloseReason reason) override;
         void rebuild() override;
         void buildInterface() override;
+
+      protected:
+        /// this have to be called in inheriting window if someone passes callback to it
+        /// othervise destruction order will cause call to non existing parent element
```
SP2FET, line 37:
```suggestion
        /// otherwise destruction order will cause call to non existing parent element
```

kkleczkowski (approved):

module-apps/apps-common/windows/OptionWindow.hpp
```diff
@@ -31,5 +31,11 @@ namespace gui
...
+        /// this is because on how OptionWindows are designed
+        void destroyForTheFuture();
```
SP2FET, line 39: almost like back to the future :D

SP2FET (approved):

mjan-dev (approved):
Approved, but please wait for response in the jira issue if there are other places in the system with this issue before merging.

.gdbinit
pholat: 1. it's not needed
2. with it we ignore sigtrap in GDB by default which is terribly confusing

module-apps/apps-common/windows/OptionWindow.hpp
pholat: feel free to provide me a better name - it will be appreciated!  (there are better ways to fix this issue, none is that short and non-intrusive though)

PrzeBrudny (approved):

pholat:
added rai class to handle all cases when we can call non existing class on destruction for the Settings windows

PR #3503, opened 2022-01-20
[BH-1333] Battery subsystem refactor

swichu91:
This PR aims to fix some issues related to the battery subsystem. Some of the changes were dictated by the existing codebase so I had to compromise to meet the deadline. The need for change was dictated mainly by the Bell but some of the changes also affected Pure. Imho the whole battery subsystem in the current state should be treated as a completely distinct entity. The EventManager is no longer a proper place for it. This would also open the way for more refactor.

List of changes:

- Improved abstract battery interface
- Fixed many bugs, especially at corner cases
- Cleaned up a bit Pure BatteryCharger implementation
- Completely reworked Bell BatteryCharger implementation
- Improver board-specific IRQ handling
- Refactored BatteryController
- Removed internal dependencies from battery_level_check, now BatteryState
- Added board agnostic drivers: CW2015 & MP2639B
- Removed EventStore dependency from BSP code(both Pure&Bell)
- Removed handling of Battery
- Added common Units module
- Added WorkerQueue implementation

Issues:

- The startup procedure is not the proper place to calibrate the Bell fuel gauge. This results in the incorrect SOC reported at the start. In the worst case, it can lead to triggering shutdown procedure(if reported SOC is < 5%).
- Pure battery/charger driver seems to not report SOC correctly.

What needs to be done:

- Pure battery driver fixes
- Move the battery controller out of the event manager. It probably should be treated as a separate service or be at least part of the system manager/power manager.
- More sophisticated startup routine -> do not perform battery calibration at the start when the current draw is the highest
- Abandon using EventStore altogether
- Replace high-level notification channel with something more sophisticated than just sending plain uint8_ts.
- Add temperature depended charging logic.


module-bsp/board/linux/hal/battery_charger/BatteryCharger.cpp
```diff
@@ -40,116 +27,117 @@ namespace hal::battery
...
+                case chargerTypeDcdCDP:
+                case chargerTypeDcdDCP:
+                    evt = Events::Charger;
+                    break;
+                default:
+                    continue;
+                }
+                xQueueSend(notificationChannel, &evt, queueTimeoutTicks);
+            }
+            vTaskDelay(taskDelay);
```
pholat, line 132: can't we just hang on read? ( is sleep really needed?)

module-bsp/board/rt1051/bellpx/hal/battery_charger/BatteryCharger.cpp
```diff
@@ -1,86 +1,217 @@
...
+                                                                               }}},
+
+          fuel_gauge_gpio{drivers::DriverGPIO::Create(fuel_gauge_irq_gpio_instance, {})},
+          fuel_gauge{CW2015{drivers::DriverI2C::Create(i2c_instance, i2c_params)}}
+    {
+
+        emergency_poll_timer = xTimerCreate("emg_timer", emergency_poll_time, pdFALSE, this, [](TimerHandle_t xTimer) {
+            BellBatteryCharger *inst = static_cast<BellBatteryCharger *>(pvTimerGetTimerID(xTimer));
+            inst->pollFuelGauge();
+        });
```
pholat, line 100: just stating: each created timer with freertos API can hang our general system timers in worst case and lessens response time for timers :) this is because system timers use freertos timer queue to notify Services about timer event (and then service executes the timer function connected )

module-bsp/board/rt1051/common/WorkerQueue.hpp
```diff
@@ -0,0 +1,97 @@
...
+
+/*
+ * Worker queue implementation is used to defer work from interrupt routines.
+ * This way, it is possible to keep interrupt routines short but at the same time process more intensive tasks.
+ * The design goal was to use as few resources as possible, hence it is a lightweight mechanism.
+ * The stack assigned to the queue worker handler is relatively small.
+ * It is not suitable for performing stack-consuming tasks, for instance handling I/O.
+ * If necessary, stack size can be manually reconfigured upon creation.
+ */
+template <typename Message> class WorkerQueue
```
pholat, line 21: I've dejavu as long time ago we were talking about the fact that all Workers in Services should work with some better WorkerQueue mechanism :) (better than generally used one )

module-bsp/board/rt1051/puretx/hal/battery_charger/BatteryCharger.cpp
```diff
@@ -1,120 +1,201 @@
...
+
+        QueueHandle_t notificationChannel;
+        inline static std::unique_ptr<BatteryWorkerQueue> workerQueue;
+    };
+
+    PureBatteryCharger::PureBatteryCharger(QueueHandle_t irqQueueHandle) : notificationChannel{irqQueueHandle}
+    {
+        workerQueue = std::make_unique<BatteryWorkerQueue>(
+            "battery_charger",
+            [this](const auto &msg) {
```
pholat, line 87: tihs might actually be an issue in cases when there would be further handling based on freertos timer and callback? ( callback to not existing element)

module-bsp/board/linux/hal/battery_charger/BatteryCharger.cpp
swichu91: The answer is no if we use a simple FreeRTOS queue mechanism, This is because we won't be able to handle the stop 'signal' invoked from BatteryCharger destructor. It does really matter that much for the simulator tho. Another solution is to use a more sophisticated messages system, for instance, `WorkQueue` used by rt1051.

swichu91: It is also possible to just invoke `vTaskDelete` from the destructor but I would consider it rather a rough way of exiting.

module-bsp/board/rt1051/puretx/hal/battery_charger/BatteryCharger.cpp
```diff
@@ -1,120 +1,201 @@
...
-    {
-        bsp::battery_charger::deinit();
-    }
+    /// A few constants to make code readability better
+    constexpr auto int_b_soc_change =
+        static_cast<std::uint16_t>(bsp::battery_charger::batteryINTBSource::SOCOnePercentChange);
+    constexpr auto int_b_max_temp = static_cast<std::uint16_t>(bsp::battery_charger::batteryINTBSource::maxTemp);
+    constexpr auto int_b_min_temp = static_cast<std::uint16_t>(bsp::battery_charger::batteryINTBSource::minTemp);
+    constexpr auto int_b_all      = static_cast<std::uint16_t>(bsp::battery_charger::batteryINTBSource::all);
+    constexpr auto int_b_miv_v    = static_cast<std::uint16_t>(bsp::battery_charger::batteryINTBSource::minVAlert);
```
wojtekrzepecki, line 37: Minor: rather use camelCase instead of underscores

module-services/service-evtmgr/battery/BatteryState.cpp
```diff
@@ -0,0 +1,149 @@
...
+    explicit Pimpl(NotifyStateChangedCallback notifyCallback) : notifyCallback{std::move(notifyCallback)}
+    {}
+
+  private:
+    NotifyStateChangedCallback notifyCallback;
+    sml::sm<StateMachine, NotifyStateChangedCallback> sm{notifyCallback};
+
+    static constexpr auto criticalThreshold = 10; // %
+    static constexpr auto shutdownThreshold = 5;  // %
+};
```
wojtekrzepecki, line 140: :+1:

module-bsp/CMakeLists.txt
```diff
@@ -13,6 +13,8 @@ target_sources(
         bsp/common.cpp
         bsp/lpm/bsp_lpm.cpp
         devices/Device.cpp
+        devices/power/CW2015.cpp
+        devices/power/MP2639B.cpp
```
wojtekrzepecki, line 17: Please move it to bell-specific directory

module-bsp/board/rt1051/bellpx/hal/battery_charger/BatteryChargerIRQ.hpp
```diff
@@ -0,0 +1,13 @@
...
+
+#pragma once
+
+#include <cstdint>
+#include "FreeRTOS.h"
+
+namespace hal::battery
+{
+    BaseType_t charger_irq();
+    BaseType_t fuel_gauge_irq();
```
wojtekrzepecki, line 12: camelCase?

module-bsp/board/rt1051/common/WorkerQueue.hpp
wojtekrzepecki: Was this architectural approach discussed with the team/system chapter?

module-bsp/board/rt1051/puretx/bsp/battery_charger/battery_charger.cpp
```diff
@@ -577,18 +571,13 @@ namespace bsp::battery_charger
...
 
         clearAllChargerIRQs();
         clearFuelGuageIRQ(static_cast<std::uint16_t>(batteryINTBSource::all));
         enableTopIRQs();
         enableChargerIRQs();
         IRQPinsInit();
 
-        // Check IRQ status register
-        INTB_IRQHandler();
-
```
wojtekrzepecki, line 591: That was crucial in the init function. Was it moved somewhere or just deleted?

module-bsp/board/rt1051/bellpx/hal/battery_charger/BatteryCharger.cpp
```diff
@@ -1,86 +1,217 @@
...
+#include <board/BoardDefinitions.hpp>
+
+namespace
 {
-    BatteryCharger::BatteryCharger(AbstractBatteryCharger::BatteryChargerEvents &eventsHandler)
-        : eventsHandler(eventsHandler)
-    {}
+    using namespace bsp::devices::power;
+
+    constexpr auto emergency_poll_time = pdMS_TO_TICKS(5 * 60 * 1000); /// 5min
```
mjan-dev, line 20: What exactly is the emergency here? I have a clue, however I believe it will become not obvious quickly. Maybe a really brief comment here would be appropriate.

module-bsp/board/rt1051/puretx/hal/battery_charger/BatteryCharger.cpp
```diff
@@ -1,120 +1,201 @@
 // Copyright (c) 2017-2021, Mudita Sp. z.o.o. All rights reserved.
 // For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
-#include "BatteryCharger.hpp"
+#include "BatteryChargerIRQ.hpp"
```
wojtekrzepecki, line 4: All battery charging has to be retested thoroughly after such a changes. You have to be aware that some things can be broken since this code is outcome of many trials and errors.

module-bsp/CMakeLists.txt
swichu91: These are no longer product-specific so I can't see any reason to put them to bell.

module-bsp/board/rt1051/puretx/bsp/battery_charger/battery_charger.cpp
swichu91: The initial state of SOC, battery voltage, and charger is fetched in the higher layers(BatteryController) so there is no longer a need to check it here.

wojtekrzepecki: Main reason was to clear status register -> can you check if it's happening anywhere else?

module-bsp/board/rt1051/bellpx/hal/battery_charger/BatteryCharger.cpp
swichu91: That's a really good point. It would be a much better solution(as you pointed out) to not use freertos timers directly but to wrap them with a container and send only a notification from the actual timer handler to some kind of queue(we can of course reuse or slightly modify the existing `WorkerQueue` or just create completely new mechanism).

module-bsp/board/rt1051/common/WorkerQueue.hpp
swichu91: This solution is nothing new. We are already using it in the existing event manager, just in the uglier way(breaking SRP and encapsulation). The goal here was to create and use bsp-related os primitives(for now I only added the queue implementation) and do not mix services/app-related things with low-level code(try to think about it as the kernel). We discussed it briefly with @pholat and have a plan of what to do next.

module-bsp/board/rt1051/puretx/hal/battery_charger/BatteryCharger.cpp
swichu91: Yeah, I'm aware of this. That's why I also tested it on Pure ;)

module-bsp/board/rt1051/puretx/hal/battery_charger/BatteryCharger.cpp
swichu91: What should it look like? `IntBMaxTemp` or `INTBMaxTemp` or maybe `intBMaxTemp` ? IMHO, in this scenario, the snake case is more readable hence that's why I picked it over camelCase. I always try to choose the right tool for the job instead of following blindly some rules ;)

wojtekrzepecki: Just pointing out that it is not the style our code is written in ;) For me it helps to see at first glance if it's third party or our code. No strong opinion for that.

module-bsp/CMakeLists.txt
wojtekrzepecki: In this cmakelists there are mainly cpp wrappers for vendor peripheral drivers. For me it does not fit so well here since I cannot see here other product-specific drivers. It may be decoupled from the product logic but it still could be found only in Bell HW.

swichu91: These drivers only appear in Bell because currently are only used by the Bell. It is still better to have a pool of specific device drivers designed that way they are not pinned to the specific board. As a matter of fact, for instance, `MAX77818` doesn't care on which platform it is used. It only requires an I2C instance and if a user wants to use IRQ, he must provide an interrupts environment prepared ahead. That's where `hal` comes into the game. Every board must implement board-specific HAL. Ideally, even boards wouldn't be pinned to the specific application(in our codebase it would be hard to achieve it without some refactor).

module-bsp/board/rt1051/puretx/bsp/battery_charger/battery_charger.cpp
swichu91: Taken from MAX77818 init code:
```
        clearAllChargerIRQs();
        clearFuelGuageIRQ(static_cast<std::uint16_t>(batteryINTBSource::all));
        enableTopIRQs();
        enableChargerIRQs();
        IRQPinsInit();
```
it looks alright to me. All IRQ status registers are cleared before enabling specific irqs. Then it comes board-specific IRQ pins init. At this point, it is safe to send IRQ notifications to the WorkerQueue as it was initialized before the MAX77818 driver. Ideally, it would be better to make MAX77818 driver board-agnostic just like the CW2015 and MP2639B.

paweljonskim (approved):

wojtekrzepecki (approved):

module-bsp/board/rt1051/puretx/hal/battery_charger/BatteryCharger.cpp
swichu91: Do you have any specific test cases you would like me to test? So far I tested:

- brown-out notifications, 
- I'm in the middle of testing the full charge, then full discharge with the phone switched on
- I also discharged the battery completely while in bootloader mode and then proceeded to launch the phone(with and without a charger plugged in)
- I also tested battery critical-level notifications(with and without charger plugged in)

wojtekrzepecki: I'll try to write down everything what comes to my mind:
- Charging until fully charged (it should communicate pluggedNotCharging) 
- Charging current changes due to different types of USB source
- Charging control due to temperature -> heating with hairdryer or cooling battery with ice is helpful.
- Proper readout of the state at system boot (there were some non-deterministic issues after many trials)
- Restart of the system with USB plugged -> charger is constantly on

mjan-dev (approved):

PR #3508, opened 2022-01-24
[BH-1372] Adapt Shutdown for Bell Harmony

tomaszrybarski-digica:
Separated shutdown handling from Pure


products/BellHybrid/sys/SystemManager.cpp
```diff
@@ -64,4 +65,35 @@ namespace sys
...
+
+        if (systemDeinit) {
+            systemDeinit();
+        }
+
+        EndScheduler();
+
+        // Power off system
+        PowerOff();
+    }
```
wojtekrzepecki, line 98: What is the purpose?

mjan-dev: @tomaszrybarski-digica ?

swichu91: IMHO, this is completely wrong (for Pure too) :/

tomaszrybarski-digica: This behavior must be separated since Bell needs a ShutdownReady state issued immediately not depending on whether device is in Charging mode or not (USB cable connected / disconnected). The device is being shutdown properly then both when USB is connected or disconnected.
The current implementation of SystemManager::Run() which Bell uses is common for Pure and Bell and in this implementation ShutdownReady is issued only in Discharge mode (USB disconnected). This causes issues in proper shutdown of Bell when device is being charged

tomaszrybarski-digica: This also has been tested and device is in proper shutdown mode (both for USB connected and disconnected - charging and discharging states). Logs can be provided

wojtekrzepecki: The proper way of doing it will be not to override whole `Run()` method but to create some function with pre-shutdown routine with empty implementation for Bell.

tomaszrybarski-digica: You mean empty implementation for Pure? The current implementation is correct for Pure. A pre-shutdown function with an implementation of just setting ShutdownReady state could be added for Bell then

wojtekrzepecki: Actually I meant something like this. in SystemManagerCommon::Run()
`while (state == State::Shutdown) {
  shutdownState(); // pure-virtual in sysMgrCommon
}
`
for Bell:
`shutdownState(){
 set(State::ShutdownReady);
}`

for Pure:
`shutdownState(){
 // Pure things
}`

PR #3512, opened 2022-01-24
[EGD-6991] Message from the operator is not fully shown

niepiekm:
Added handling of errors in retrieving a message from the modem

### DIFF
--- a/module-services/service-cellular/ServiceCellular.cpp
+++ b/module-services/service-cellular/ServiceCellular.cpp
@@ -912,8 +912,6 @@ std::optional<std::shared_ptr<sys::Message>> ServiceCellular::identifyNotificati
 
 auto ServiceCellular::receiveSMS(std::string messageNumber) -> bool
 {
-    constexpr auto ucscSetMaxRetries = 3;
-
     auto retVal = true;
 
     auto channel = cmux->get(CellularMux::Channel::Commands);
@@ -923,7 +921,7 @@ auto ServiceCellular::receiveSMS(std::string messageNumber) -> bool
     }
 
     auto ucscSetRetries = 0;
-    while (ucscSetRetries < ucscSetMaxRetries) {
+    while (ucscSetRetries < at::AtCmdMaxRetries) {
         if (!channel->cmd(at::AT::SMS_UCSC2)) {
             ++ucscSetRetries;
             LOG_ERROR("Could not set UCS2 charset mode for TE. Retry %d", ucscSetRetries);
@@ -948,8 +946,26 @@ auto ServiceCellular::receiveSMS(std::string messageNumber) -> bool
     std::string messageRawBody;
     UTF8 receivedNumber;
     const auto &cmd = at::factory(at::AT::QCMGR);
-    auto ret        = channel->cmd(cmd + messageNumber, cmd.getTimeout());
-    if (!ret) {
+    at::Result ret;
+    auto qcmgrRetries = 0;
+
+    while (qcmgrRetries < at::AtCmdMaxRetries) {
+        ret = channel->cmd(cmd + messageNumber, cmd.getTimeout());
+        if (!ret) {
+            ++qcmgrRetries;
+            LOG_ERROR("Could not read text message. Retry %d", qcmgrRetries);
+            // There are cases where +QCMGR command is issued to soon after +CMTI URC,
+            // and modem is still busy processing messages. This results with an error
+            // and some garbage frame. To work around this, we delay a bit and retry +QCMGR
+            // to get the message.
+            vTaskDelay(at::AtCmdRetryDelayMS);
+        }
+        else {
+            break;
+        }
+    }
+
+    if (!ret || qcmgrRetries == at::AtCmdMaxRetries) {
         LOG_ERROR("!!!! Could not read text message !!!!");
         retVal = false;
     }
--- a/module-cellular/at/Commands.hpp
+++ b/module-cellular/at/Commands.hpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2021, Mudita Sp. z.o.o. All rights reserved.
+// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
 // For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 #pragma once
@@ -145,4 +145,7 @@ namespace at
     };
 
     std::vector<AT> getCommadsSet(commadsSet set);
+
+    constexpr auto AtCmdMaxRetries   = 3;
+    constexpr auto AtCmdRetryDelayMS = 50;
 }; // namespace at


module-services/service-cellular/ServiceCellular.cpp
```diff
@@ -923,10 +921,11 @@ auto ServiceCellular::receiveSMS(std::string messageNumber) -> bool
     }
 
     auto ucscSetRetries = 0;
-    while (ucscSetRetries < ucscSetMaxRetries) {
+    while (ucscSetRetries < at::AtCmdMaxRetries) {
         if (!channel->cmd(at::AT::SMS_UCSC2)) {
             ++ucscSetRetries;
             LOG_ERROR("Could not set UCS2 charset mode for TE. Retry %d", ucscSetRetries);
+            vTaskDelay(at::AtCmdRetryDelayMS);
```
alekrudnik, line 928: can you at least desribe why this is needed?

module-services/service-cellular/ServiceCellular.cpp
```diff
@@ -948,12 +947,22 @@ auto ServiceCellular::receiveSMS(std::string messageNumber) -> bool
...
+            LOG_ERROR("Could not read text message. Retry %d", qcmgrRetries);
+            vTaskDelay(at::AtCmdRetryDelayMS);
+        }
+        else {
+            break;
+        }
     }
-    else {
+
+    {
```
pholat, line 965: didn't you change logic here? was it on purpose?

module-services/service-cellular/ServiceCellular.cpp
```diff
@@ -948,12 +947,22 @@ auto ServiceCellular::receiveSMS(std::string messageNumber) -> bool
...
-    auto ret        = channel->cmd(cmd + messageNumber, cmd.getTimeout());
-    if (!ret) {
-        LOG_ERROR("!!!! Could not read text message !!!!");
-        retVal = false;
+    at::Result ret;
+    auto qcmgrRetries = 0;
+
+    while (qcmgrRetries < at::AtCmdMaxRetries) {
+        ret = channel->cmd(cmd + messageNumber, cmd.getTimeout());
+        if (!ret) {
```
alekrudnik, line 955: should we retry for each error?

alekrudnik (approved):

module-services/service-cellular/ServiceCellular.cpp
niepiekm: For this command, no error was experienced so far. Delay is added just in case.

module-services/service-cellular/ServiceCellular.cpp
niepiekm: I've been wondering about that, too. We could warp the execution of every command with a retry just in case.

module-services/service-cellular/ServiceCellular.cpp
pholat: to be completly honest - adding sleeps to the code "to be sure" doesn't sound like a good idea.

do you know how `channel->cmd` works?
it waits till there is set response from the modem, or error ...

module-services/service-cellular/ServiceCellular.cpp
pholat: @alekrudnik tbh we should rather have a separate "algorithm piece of code (not embedded strictly into the `ServiceCellular.cpp` which would handle receiving messages as a whole - both short, and concatenated ones.

@niepiekm what is the exact error reason when we trigger the delay? do you know that `channel->cmd()` awaits for "data or error" for selected time?

pholat (changes requested):
doesn't seem right

module-services/service-cellular/ServiceCellular.cpp
pholat: please do not close my comments before I could verify and let me do it.

kkleczkowski (approved):

module-services/service-cellular/ServiceCellular.cpp
niepiekm: `channel->cmd()` executes a command and waits for the specific amount  timeout defined for each command in `module-cellular/at/src/ATFactory.cpp`. The delay this PR introduces is just a delay between retries of the command in case of command error.

module-services/service-cellular/ServiceCellular.cpp
niepiekm: As I mentioned [above](https://github.com/mudita/MuditaOS/pull/3512/files#r791508562) the delay is just for retry of a command. Respective command execution time is handleded by the `channel-cmd()` itself per command.

module-services/service-cellular/ServiceCellular.cpp
pholat: Actually, it waits for data - **or** - specific amount of time defined. The command timeout is meant to set a limit to wait time. We do not wait all that time each time we execute the command.
I am wondering - is the response immediate with error? is there a bus error returned or some other error? what's the exact cause which set `operator bool` invalid? How much time it takes to execute `channel->cmd()` execution?

niepiekm: From implementation, the response is returned immediately in case of an error received from the `cellular`. If a timeout occurs, it's reported as well by the timeout result code.

Execution of `channle->cmd()` is timed in module-cellular/modem/ATCommon.cpp#L117, but it's not used for anything.

pholat: I'm pretty sure that time is passed here:
`Result Channel::cmd(const std::string &cmd, std::chrono::milliseconds timeout, size_t rxCount)` isn't it? 

To be more specific:
1. what causes the at::Result::operator boo() const` returning false in this case?
2. how long doest it take to execute `(!channel->cmd(at::AT::SMS_UCSC2))` without sleep(50ms)? ( you can check it with 117 line, but you have to enable this first - see: `module-utils/log/api/log/debug.hpp` `DEBUG_SCOPED_TIMINGS`z

niepiekm: RE 1: The error case covered by this PR is a deserialization error of malformed response frame:
```
104546 ms DEBUG [ServiceCellular] ATCommon.cpp:cmd:79: start of AT+QCMGR=0
104571 ms DEBUG [CellularMux] CellularMux.cpp:workerTaskFunction:554: DMA buffer full
104573 ms DEBUG [CellularMux] CellularMux.cpp:workerTaskFunction:554: DMA buffer full
{104573 ms DEBUG [CellularMux] CellularMuxFrame.h:deserialize:169: F9
[128 lines of byte dump ...]
104595 ms DEBUG [CellularMux] CellularMuxFrame.h:deserialize:169: F9 
}
104595 ms ERROR [CellularMux] CellularMuxFrame.h:deserialize:177: Received frame FCS [0xBD] != 0xCF error. Dropping.
104595 ms DEBUG [ServiceCellular] ATCommon.cpp:cmdLog:47: [AT]: >AT+QCMGR=0<, ><
104595 ms ERROR [ServiceCellular] ServiceCellular.cpp:receiveSMS:957: Could not read text message. Retry 1
```
RE 2: It takes 3 ms to execute `AT::SMS_UCSC2` command without sleeping for retry
```
71081 ms DEBUG [ServiceCellular] ATCommon.cpp:cmd:79: start of AT+CSCS="UCS2"
71084 ms DEBUG [ServiceCellular] ATCommon.cpp:cmdLog:47: [AT]: >AT+CSCS="UCS2"<, >OK<
71084 ms INFO  [ServiceCellular] ATCommon.cpp:cmdLog:52: [AT] > OK
71084 ms DEBUG [ServiceCellular] ScopedTime.hpp:~Scoped:35: Time to run at commandAT+CSCS="UCS2" time: 3
71084 ms DEBUG [ServiceCellular] ScopedTime.hpp:~Scoped:35: Time to run at commandAT+CSCS="UCS2" time: 3
```

pholat: Yeah, but there is no information on how much it takes when there is an issue on DMA (from logs it seems like 50ms?) , and there is no information about the result in at::Resutl.

but from logs: **there is issue on DMA** and this is why I asked for result code returned in at::Result :(

mjan-dev: Applying some more thought through strategy to respond to errors here is out of scope for this PR and as @pholat mentioned, in fact the underlying problem seems to lie with DMA so we are simply applying a workaround here. Nevertheless, if we have no alternatives, lets apply those changes. Closing this conversation as discussed with @pholat.

niepiekm: @pholat, and just 👍🏻, please.

mjan-dev, line 928: @niepiekm Please at least add a comment on why this delay is present here and what problem it solves.

module-services/service-cellular/ServiceCellular.cpp
mjan-dev: Adding a default delay in cmd() seems to not be a good decision. Lets add some encapsulation/policy in place so it would be easily changeable for specific commands.

module-services/service-cellular/ServiceCellular.cpp
niepiekm: @mjan-dev This particular instance of delay is not actually needed, as the command hasn't caused problems so far. I just added it for consistency but will remove it.

PR #3520, opened 2022-01-26
EGD-8208

pholat:
Removed funky logic for Ring on/off (removing race condition on start->stop->start) when we end call "too fast" )


module-services/service-cellular/CallAudio.cpp
```diff
@@ -0,0 +1,22 @@
...
+    const auto filePath = AudioServiceAPI::GetSound(&owner, audio::PlaybackType::CallRingtone);
+    AudioServiceAPI::PlaybackStart(&owner, audio::PlaybackType::CallRingtone, filePath);
+}
+
+void CallRingAudio::stop()
+{
+    /// NOTE: we should fix it to stop actual played ringtone...
+    /// but we would require async response from PlayBack start or API in audio to be able to
+    /// just stop Ringtone by path
+    AudioServiceAPI::StopAll(&owner);
```
pholat, line 21: there is no way to just cancel if we are playing ring or selected audio path.

module-services/service-cellular/ServiceCellular.cpp
```diff
@@ -2194,9 +2196,9 @@ auto ServiceCellular::handleCellularRingNotification(sys::Message *msg) -> std::
         return std::make_shared<CellularResponseMessage>(hangUpCall());
     }
 
+    /// NOTE: the code below should be investigated as there is something weird in this flow
```
pholat, line 2199: for some weird reason, we are treating RING as CLIP in here. Though currently without this messaging we have automatic call denial.

pholat:
converted to draft as I want to check a few more things to be sure.

module-services/service-cellular/CallAudio.cpp
pholat: now there is

pholat:
need to change call flow :|

kkleczkowski (approved):

module-services/service-cellular/CallAudio.cpp
```diff
@@ -0,0 +1,41 @@
...
+CallRingAudio::~CallRingAudio()
+{
+    delete meta;
+}
+
+void CallRingAudio::play()
+{
+    started = true;
+    const auto file = AudioServiceAPI::GetSound(&owner, audio::PlaybackType::CallRingtone);
+    meta->p         = owner.async_call<AudioStartPlaybackRequest, AudioStartPlaybackResponse>(
```
mjan-dev, line 30: is meta always initialized here?

module-services/service-cellular/ServiceCellular.cpp
```diff
@@ -1923,25 +1924,23 @@ auto ServiceCellular::handleDBNotificationMessage(db::NotificationMessage *msg)
...
 
 auto ServiceCellular::handleCellularIncomingCallMessage(sys::Message *msg) -> std::shared_ptr<sys::ResponseMessage>
 {
     LOG_INFO("%s", __PRETTY_FUNCTION__);
     auto ret     = true;
-    auto message = static_cast<CellularIncominCallMessage *>(msg);
-
-    if (!ongoingCall.isValid()) {
-        ret = ongoingCall.startCall(message->number, CallType::CT_INCOMING);
-    }
```
mjan-dev, line 1937: so after changes this handler doesn't really do anything?

module-services/service-cellular/ServiceCellular.cpp
```diff
@@ -2221,6 +2221,7 @@ auto ServiceCellular::handleCellularCallerIdNotification(sys::Message *msg) -> s
         return std::make_shared<CellularResponseMessage>(hangUpCall());
     }
 
+
```
mjan-dev, line 2224: minor: extra newline

module-sys/PhoneModes/Observer.cpp
```diff
@@ -7,25 +7,6 @@
 
 namespace sys::phone_modes
 {
-    namespace
-    {
-        template <typename F> void onChanged(const F &onChanged, const Observer::OnFinishedCallbacks &onFinished)
```
mjan-dev, line 12: Just to make sure: this observer is only used in the changed functionality and does not impact others?

module-sys/Service/Service.cpp
```diff
@@ -21,35 +21,29 @@
...
+    const char *realname = typeid(*ptr).name();
 
     assert(srvc);
     assert(ptr);
 
-    LOG_DEBUG("Handle message ([%s] -> [%s] (%s) data: %s %s",
+    if (xPortIsInsideInterrupt()) {
+        throw std::runtime_error("message sent from irq");
+    }
+    if (xTaskGetCurrentTaskHandle() == nullptr) {
```
mjan-dev, line 37: the debug messages being sent from interrupts/crit sections have been observed in case of this issue or it is just a general "side" improvement?

module-sys/Service/Service.cpp
```diff
@@ -280,9 +263,6 @@ namespace sys
     {
         auto ret = ReturnCodes::Success;
         switch (message->systemMessageType) {
-        case SystemMessageType::Ping:
```
mjan-dev, line 283: what was this message type used for ? or it was not used at all?

module-sys/Service/Service.cpp
pholat: it's just a minor improvement :) we cant log in irq, nor we can allocate - therefore I've added it.
We should never ever need these two if's.

module-sys/Service/Service.cpp
pholat: it's a legacy Ping message :D it was used for some code proofing when the bus was added, then it was abandoned.
Now it's not used at all - this is why I could remove it :)

module-sys/PhoneModes/Observer.cpp
pholat: this was useless code - it did literally nothing at the moment of removal. 

saying that - with this code removed we can clearly see that we have one observer which has to be modified each time we want to observe something different. We should have separate, can be templated, observers per thing we want to observe.
This 3 min change should explain what I've in mind: 
image

mjan-dev (approved):

PR #3521, opened 2022-01-26
[EGD-8162] Allow importing audio profile settings from json file

SP2FET:
This allows for imporing the audio profile setitngs, including
EQ settings from the json file, thus allowing modification of
those setitngs without the need of recompiling the source code.

### DIFF
--- a/module-audio/Audio/Profiles/ProfileConfigUtils.cpp
+++ b/module-audio/Audio/Profiles/ProfileConfigUtils.cpp
@@ -1,0 +1,108 @@
+// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
+// For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
+
+#include "ProfileConfigUtils.hpp"
+#include <fstream>
+#include <log/log.hpp>
+#include <json11.hpp>
+#include <magic_enum.hpp>
+namespace audio
+{
+    namespace strings
+    {
+        constexpr inline auto samplerate   = "samplerate";
+        constexpr inline auto bitWidth     = "bitWidth";
+        constexpr inline auto flags        = "flags";
+        constexpr inline auto outputVolume = "outputVolume";
+        constexpr inline auto outputPath   = "outputPath";
+        constexpr inline auto inputGain    = "inputGain";
+        constexpr inline auto inputPath    = "inputPath";
+        constexpr inline auto filterParams = "filterParams";
+        constexpr inline auto filterType   = "filterType";
+        constexpr inline auto frequency    = "frequency";
+        constexpr inline auto Q            = "Q";
+        constexpr inline auto gain         = "gain";
+    } // namespace strings
+
+    namespace utils
+    {
+
+        template <typename E, typename T>
+        constexpr inline typename std::enable_if<std::is_enum<E>::value && std::is_integral<T>::value, E>::type toEnum(
+            T value) noexcept
+        {
+            return static_cast<E>(value);
+        }
+
+        equalizer::FilterType toFilterType(const std::string &filterName)
+        {
+            auto filterType = magic_enum::enum_cast<equalizer::FilterType>(filterName);
+            if (filterType.has_value()) {
+                return filterType.value();
+            }
+            else {
+                LOG_ERROR("Unknown filter type, using none");
+                return equalizer::FilterType::None;
+            }
+        }
+
+        const std::string readFileToString(std::filesystem::path filePath)
+        {
+            std::ifstream file;
+            std::string configString;
+            LOG_DEBUG("Reading %s ...", filePath.c_str());
+            file.open(filePath);
+            if (not file.is_open()) {
+                LOG_ERROR("Can't open profile configuration file, using defaults!");
+                throw std::invalid_argument("Can't open file!");
+            }
+            while (file) {
+                std::string line;
+                std::getline(file, line);
+                configString += line;
+            }
+            file.close();
+            return configString;
+        }
+    } // namespace utils
+
+    const audio::codec::Configuration loadConfigurationFromFile(std::filesystem::path filePath)
+    {
+
+        auto configString = utils::readFileToString(filePath);
+        audio::codec::Configuration config;
+
+        json11::Json configJson;
+        std::string err;
+        configJson = json11::Json::parse(configString.c_str(), err);
+        if (!err.empty()) {
+            LOG_ERROR("Failed parsing device string!");
+            throw std::invalid_argument("Can't parse the file!");
+        }
+        config.sampleRate_Hz = configJson[strings::samplerate].int_value();
+        config.bitWidth      = configJson[strings::bitWidth].int_value();
+        config.flags         = configJson[strings::flags].int_value();
+        config.outputVolume  = configJson[strings::outputVolume].number_value();
+        config.inputGain     = configJson[strings::inputGain].number_value();
+        config.inputPath     = utils::toEnum<codec::InputPath>(configJson[strings::inputPath].int_value());
+        config.outputPath    = utils::toEnum<codec::OutputPath>(configJson[strings::outputPath].int_value());
+
+        json11::Json::array paramsArray;
+        audio::equalizer::Equalizer filterParams;
+        paramsArray = configJson[strings::filterParams].array_items();
+
+        for (size_t i = 0; i < equalizer::bands; i++) {
+            auto filterType = utils::toFilterType(paramsArray[i][strings::filterType].string_value());
+            auto frequency  = paramsArray[i][strings::frequency].number_value();
+            auto samplerate = paramsArray[i][strings::samplerate].int_value();
+            auto Q          = paramsArray[i][strings::Q].number_value();
+            auto gain       = paramsArray[i][strings::gain].number_value();
+
+            filterParams.at(i) = qfilter_CalculateCoeffs(filterType, frequency, samplerate, Q, gain);
+        }
+
+        config.filterCoefficients = filterParams;
+        return config;
+    }
+
+} // namespace audio
--- a/module-audio/Audio/Profiles/ProfileRoutingHeadphones.hpp
+++ b/module-audio/Audio/Profiles/ProfileRoutingHeadphones.hpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2021, Mudita Sp. z.o.o. All rights reserved.
+// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
 // For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 #pragma once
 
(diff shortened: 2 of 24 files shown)


module-audio/Audio/Profiles/ProfileConfigUtils.cpp
```diff
@@ -0,0 +1,108 @@
...
+            LOG_ERROR("Failed parsing device string!");
+            return config;
+        }
+        config.sampleRate_Hz = configJson[strings::samplerate].int_value();
+        config.bitWidth      = configJson[strings::bitWidth].int_value();
+        config.flags         = configJson[strings::flags].int_value();
+        config.outputVolume  = configJson[strings::outputVolume].number_value();
+        config.inputGain     = configJson[strings::inputGain].number_value();
+        config.inputPath     = utils::toEnum<codec::InputPath>(configJson[strings::inputPath].int_value());
+        config.outputPath    = utils::toEnum<codec::OutputPath>(configJson[strings::outputPath].int_value());
```
pholat, line 88: +1

module-audio/Audio/Profiles/ProfileConfigUtils.cpp
```diff
@@ -0,0 +1,108 @@
...
+        audio::codec::Configuration config;
+
+        json11::Json configJson;
+        std::string err;
+        configJson = json11::Json::parse(configString.c_str(), err);
+        if (!err.empty()) {
+            LOG_ERROR("Failed parsing device string!");
+            return config;
+        }
+        config.sampleRate_Hz = configJson[strings::samplerate].int_value();
```
pholat, line 82: from the documentation of library used: 
```
There are also implicit constructors that allow standard and user-defined types to be automatically converted to JSON. 
```
Why wont we use it? :)

module-audio/Audio/equalizer/Equalizer.hpp
```diff
@@ -17,19 +18,19 @@ namespace audio::equalizer
         float a2;
     };
 
-    using Equalizer = std::array<QFilterCoefficients, 5>;
+    using Equalizer = std::array<QFilterCoefficients, bands>;
```
pholat, line 21: just curious - do you know how and why we ended up with `bands=5` ? we can set up to 9 filters up to 5 times (if I'm reading the code right :) ) it has not much to do with this PR - it has everything to do with how this code is built and you just happened to touch it therefore you might know the answer :)

module-audio/Audio/test/unittest_config_utils.cpp
```diff
@@ -0,0 +1,61 @@
...
+        REQUIRE(config.filterCoefficients.at(3).a2 == filterCoefficients.at(3).a2);
+        REQUIRE(config.filterCoefficients.at(3).b0 == filterCoefficients.at(3).b0);
+        REQUIRE(config.filterCoefficients.at(3).b1 == filterCoefficients.at(3).b1);
+        REQUIRE(config.filterCoefficients.at(3).b2 == filterCoefficients.at(3).b2);
+
+        REQUIRE(config.filterCoefficients.at(4).a1 == filterCoefficients.at(4).a1);
+        REQUIRE(config.filterCoefficients.at(4).a2 == filterCoefficients.at(4).a2);
+        REQUIRE(config.filterCoefficients.at(4).b0 == filterCoefficients.at(4).b0);
+        REQUIRE(config.filterCoefficients.at(4).b1 == filterCoefficients.at(4).b1);
+        REQUIRE(config.filterCoefficients.at(4).b2 == filterCoefficients.at(4).b2);
```
pholat, line 59: use the for loop Luke? :) what exactly do we test here? :)

pholat (approved):

module-audio/Audio/test/unittest_config_utils.cpp
SP2FET: my mind was blackened during writing those checks :D sure - will change to for loop
It is testing if the parameters read from the file are the same as calculated with the same data provided, thus it checks if the data was parsed correctly. Any mismatch would result in a different coefficients

module-audio/Audio/equalizer/Equalizer.hpp
SP2FET: as I found in the codec datasheet, we can setup (based on the regs config) up to 7 bands. Right now it is set up for 5 bands but it was newer written explicit anywhere. I've moved it to a constexpr to be able to bond test to the proper code without using a magic number

module-audio/Audio/Profiles/ProfileConfigUtils.cpp
SP2FET: TBH it wouldn't clarify it too much - I'd have to write a `toJson` function, which would look just the same :P
But sure, if it would be better I can rewrite it. The conversion is not used anywhere besides the profile constructor

pholat: actually, there are some gains:

layout wise:
1. you will have a specialized class you test.
2. with a specialized class - you can always make a base class, and then create separate specialization if you wish for some other config loaded (i.e. from settings :) ) with the same values, just different loading
3. then you are able to test just derived classes parsing

by use:
1. you can then assign in the constructor and fail early. We cant parse? we throw in ctor and handle the exception in the layer below :) 

TL;DR;
I would say that it's a construction of proper config. I might try to argue it by this: https://isocpp.org/wiki/faq/exceptions#ctors-can-throw or fail fast principle :)

module-audio/Audio/equalizer/Equalizer.hpp
pholat: Thanks for the elaborate description. I don't do this often but I would support leaving a comment explaining it here :) ( I know it's not your code, and you might feel weird passing explanation in possibly codec agnostic code about the codec - but it's a rare place where I think it might be nice to inform somebody about a trap :) )

module-audio/Audio/test/unittest_config_utils.cpp
pholat: tbh you can add `operator==` and just compare the `QFilterCoefficients` as you are actually doing that, don't you? :)

SP2FET: yep - I did the same in Bluetooth settings :thinking:

module-audio/Audio/equalizer/Equalizer.hpp
SP2FET: np problem, I can describe it in a comment in the code if it's so confusing ;)

module-audio/Audio/test/unittest_config_utils.cpp
SP2FET: fixed :D

module-audio/Audio/Profiles/ProfileConfigUtils.cpp
SP2FET: I did read the documentation once again - it is possible to serialize it with `toJson` method, but we need a deserialize here, which is not possible this way :(

module-audio/Audio/equalizer/Equalizer.hpp
SP2FET: described in the comment

alekrudnik (approved):

module-audio/Audio/Profiles/ProfileConfigUtils.cpp
```diff
@@ -0,0 +1,108 @@
...
+        config.bitWidth      = configJson[strings::bitWidth].int_value();
+        config.flags         = configJson[strings::flags].int_value();
+        config.outputVolume  = configJson[strings::outputVolume].number_value();
+        config.inputGain     = configJson[strings::inputGain].number_value();
+        config.inputPath     = utils::toEnum<codec::InputPath>(configJson[strings::inputPath].int_value());
```
mjan-dev, line 87: Can't it be done with magic_enum::enum_cast ?

module-audio/Audio/Profiles/ProfileRoutingHeadphones.hpp
```diff
@@ -11,30 +11,26 @@ namespace audio
...
+                      .inputGain    = static_cast<float>(gain),
+                      .inputPath    = audio::codec::InputPath::Headphones,
+                      .outputPath   = audio::codec::OutputPath::Headphones,
+                      .filterCoefficients =
+                          {qfilter_CalculateCoeffs(audio::equalizer::FilterType::HighPass, 997.f, 44100, 0.701f, 0),
+                           qfilter_CalculateCoeffs(audio::equalizer::FilterType::LowPass, 4993.7f, 44100, 0.701f, 0),
+                           qfilter_CalculateCoeffs(audio::equalizer::FilterType::None, 15975.7f, 44100, 0.701f, -10),
+                           qfilter_CalculateCoeffs(audio::equalizer::FilterType::None, 200.4f, 44100, 0.701f, -10),
+                           qfilter_CalculateCoeffs(audio::equalizer::FilterType::None, 1496.7f, 44100, 0.701f, -4)}},
+                  AudioDevice::Type::Audiocodec)
```
mjan-dev, line 33: Why headset equalizer settings are hardcoded and not in config files as well?

module-audio/Audio/Profiles/README.md
```diff
@@ -0,0 +1,50 @@
+# Profile JSON file
+This document aims to describe the profile file and how to tweak those values.
+
+## First things first
+The JSON files describing the profile parameters are used to configure the audio profile with proper input/output and filter settings. 
+It is **NOT** recommended to change anything than the filter settings as it might make your device speechless.
```
mjan-dev, line 6: Good idea to point it out! :+1:

mjan-dev: Would it be possible to add a fallback configuration or it is tedious to validate provided configs?

module-audio/Audio/equalizer/Equalizer.cpp
```diff
@@ -48,55 +48,55 @@ namespace audio::equalizer
...
+        case FilterType::Parametric:
             filter_coeff.b0 = 1.0 + alpha * gain_abs;
             filter_coeff.b1 = -2.0 * cs;
             filter_coeff.b2 = 1.0 - alpha * gain_abs;
             a0              = 1.0 + alpha / gain_abs;
             filter_coeff.a1 = -2.0 * cs;
             filter_coeff.a2 = 1.0 - alpha / gain_abs;
             break;
-        case FilterType::FilterNone:
+        case FilterType::None:
```
mjan-dev, line 99: :heart: Thanks for cleaning up this naming!

mjan-dev (approved):
Approved with minor notes.

module-audio/Audio/Profiles/ProfileConfigUtils.cpp
SP2FET: to be honest I left it on purpose. Why? You have to think twice how to setup the path correctly, and there's smaller probability that someone will mess up with the paths. But that's my opinion - I can move it to strings and enum_cast if the verbosity is more important here.

module-audio/Audio/Profiles/ProfileRoutingHeadphones.hpp
SP2FET: good point! I thought that we don't mess up with EQ for headphones as the audio path should be as clean as possible - it's up to headphones speakers how they reproduce the sound, but I was wrong - there are some settings for headphones too. I'll move it to the file too.

module-audio/Audio/Profiles/README.md
SP2FET: Yep, I'm actually rewriting it a little bit to provide a fallback when file loading/parsing give an error

SP2FET: Although it won't work if someone e.g. disables the path in properly formatted JSON file - IMO we shouldn't validate the file in this way as this is not a normal file for user to tweak - if someone does something with this - one should know it's consequences :P

module-audio/Audio/Profiles/ProfileConfigUtils.cpp
SP2FET: Moved the load to the ctor with fallback handling

PR #3522, opened 2022-01-26
[EGD-8193] Proper system shutdown durign update

alekrudnik:
Made sure all services are deinitialized properly

### DIFF
--- a/module-services/service-appmgr/model/ApplicationManagerCommon.cpp
+++ b/module-services/service-appmgr/model/ApplicationManagerCommon.cpp
@@ -157,14 +157,12 @@ namespace app::manager
         ActionRequest act = ActionRequest{this->GetName(), app::manager::actions::DisplayLogoAtExit, nullptr};
         switch (closeReason) {
         case sys::CloseReason::SystemBrownout:
-            [[fallthrough]];
         case sys::CloseReason::LowBattery:
             act = ActionRequest{this->GetName(), app::manager::actions::SystemBrownout, nullptr};
             break;
         case sys::CloseReason::RegularPowerDown:
-            [[fallthrough]];
         case sys::CloseReason::Reboot:
-            [[fallthrough]];
+        case sys::CloseReason::RebootToUpdate:
         case sys::CloseReason::FactoryReset:
             break;
         }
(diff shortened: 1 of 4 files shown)


module-services/service-appmgr/model/ApplicationManagerCommon.cpp
```diff
@@ -157,14 +157,12 @@ namespace app::manager
         ActionRequest act = ActionRequest{this->GetName(), app::manager::actions::DisplayLogoAtExit, nullptr};
         switch (closeReason) {
         case sys::CloseReason::SystemBrownout:
-            [[fallthrough]];
```
niepiekm, line 160: Doesn't the compiler complain about the lack of `fallthrough`?

lucckb (approved):

module-services/service-appmgr/model/ApplicationManagerCommon.cpp
alekrudnik: No, it is designed for slightly different case https://en.cppreference.com/w/cpp/language/attributes/fallthrough

module-sys/SystemManager/SystemManagerCommon.cpp
```diff
@@ -513,17 +505,14 @@ namespace sys
...
                 case Code::Restore:
                     RestoreSystemHandler();
                     break;
                 case Code::Reboot:
-                    RebootHandler(State::Reboot);
+                    RebootHandler();
                     break;
                 case Code::RebootToUpdate:
-                    RebootHandler(State::RebootToUpdate, data->updateReason);
+                    RebootToUpdateHandler(data->updateReason);
```
pholat, line 515: please double check if we don't double reverse the list of services somewhere now :) tbh I can't see any sane way when it could harm us, though it would be nice to double-check just in case our system might surprise us.

pholat (approved):

module-sys/SystemManager/SystemManagerCommon.cpp
alekrudnik: Confirmed. Done only once

Maciej-Mudita (approved):

PR #3523, opened 2022-01-27
[EGD-8128] Baudrate change for cellular transmission

Maciej-Mudita:
UART baudrate changed from 460800 to 115200
* lower power consumption by modem
* CPU does not need to wake up to handle URC


mjan-dev (approved):

kkleczkowski (approved):

alekrudnik:
What does it mean CPU does not need to wake up to handle URC?

Maciej-Mudita:
> What does it mean CPU does not need to wake up to handle URC?

@alekrudnik We are currently setting the CPU frequency lock to a minimum of 132 MHz because there were problems with UART communication. After changing the baudrate, this blockade is unnecessary because the transmission works properly even at the lowest CPU frequency of 12 MHz.

niepiekm (approved):

niepiekm (approved):

PR #3526, opened 2022-01-28
BH-1383 Battery scaling table

swichu91:
Removed SOC scaling from GUI. From now, scaling is performed in product-specific `BatteryCharger`.
Fixed minor bugs.
Minor improvements.


module-services/service-evtmgr/battery/BatteryState.cpp
```diff
@@ -136,7 +136,7 @@ class BatteryState::Pimpl
     sml::sm<StateMachine, NotifyStateChangedCallback> sm{notifyCallback};
 
     static constexpr auto criticalThreshold = 10; // %
-    static constexpr auto shutdownThreshold = 5;  // %
+    static constexpr auto shutdownThreshold = 1;  // %
```
mjan-dev, line 139: This shutdown threshold is after scaling?

module-utils/utility/ValueScaler.hpp
```diff
@@ -0,0 +1,60 @@
...
+    template <typename T> struct Entry
+    {
+        const Range<T> input;
+        const Range<T> output;
+    };
+
+    /// Try to find the given input value in the the entries array. If the value is found, perform scaling and then
+    /// return the scaled value. If the value is not found, return std::nullopt.
+    template <typename T, const size_t N>
+    [[nodiscard]] std::optional<T> find_and_scale_value(const std::array<Entry<T>, N> &entries, const T val)
```
mjan-dev, line 49: Do we anticipate a situation when we get nullopt here?

module-utils/utility/tests/value_scaler.cpp
```diff
@@ -0,0 +1,62 @@
...
+            REQUIRE_FALSE(find_and_scale_value(entries, -11));
+            REQUIRE(find_and_scale_value(entries, -10).value() == 0);
+            REQUIRE(find_and_scale_value(entries, -1).value() == 0);
+            REQUIRE(find_and_scale_value(entries, 0).value() == 1);
+            REQUIRE(find_and_scale_value(entries, 6).value() == 2);
+            REQUIRE(find_and_scale_value(entries, 10).value() == 5);
+            REQUIRE(find_and_scale_value(entries, 99).value() == 99);
+        }
+    }
+}
```
mjan-dev, line 62: :+1:

module-services/service-evtmgr/battery/BatteryState.cpp
swichu91: This layer knows nothing about scaling. It only performs a shutdown procedure when battery SOC is lower than 1%. In terms of real SOC values, 1% may translate to 5% or any other value. IMHO it's a much more flexible solution as different hardware may implement this value completely differently. For some systems, it might be reasonable to operate in 10 to 90% and so on.  Currently, both Pure and Bell will be 1% when it is 5%. 

Another thing is battery aging calculations, state of health, etc. It might be also necessary to recalculate these thresholds based on various battery state indicators. It seems like the battery subsystem might be much more complicated than we thought :/ Without a doubt, whole battery management should be removed from the event manager and treated as a separate entity.

module-utils/utility/ValueScaler.hpp
swichu91: There is no such scenario for both Pure and Bell, but we perform return value checks anyway :)

products/PurePhone/sys/SystemManager.cpp
```diff
@@ -134,8 +134,6 @@ namespace sys
     {
         SystemManagerCommon::batteryNormalLevelAction();
         CellularServiceAPI::ChangeModulePowerState(this, cellular::service::State::PowerState::On);
-        auto battNormalMsg = std::make_shared<CriticalBatteryLevelNotification>(false);
-        bus.sendUnicast(std::move(battNormalMsg), service::name::appmgr);
```
wojtekrzepecki, line 138: This will break low level battery notifications for Pure -> it will never recover from the low battery screen

module-services/service-evtmgr/battery/BatteryState.cpp
wojtekrzepecki: What do you mean by: `Currently, both Pure and Bell will be 1% when it is 5%.` . For Pure 5% is 5% and we are shutting down it to leave reasonable amount of  residual charge in the battery.

products/PurePhone/sys/SystemManager.cpp
swichu91: It also triggers a crash during system start  under specific conditions. In a such scenario there is virtually no possibility to turn on the phone.

module-services/service-evtmgr/battery/BatteryState.cpp
swichu91: Have you had a chance to check this PR entirely? As I said, scaling is performed at hal/BatteryCharger level. Shutdown will be still performed when battery SOC is 5%.

products/PurePhone/sys/SystemManager.cpp
wojtekrzepecki: Instead of deleting it, please report the bug with clear reproduction. TBH I have never observed crash caused by it.

pholat (approved):

kkleczkowski (approved):

module-services/service-evtmgr/battery/BatteryState.cpp
mjan-dev: As for my question, it has been answered. I believe @wojtekrzepecki concern has also been cleared.

mjan-dev (approved):

PR #3530, opened 2022-01-28
[CP-1007] Add API for rebooting to USB MSC mode

pawel-mudita:
It will allow to reboot the device to USB MSC mode
without need to interact with a physical keyboard.

### DIFF
--- a/module-services/service-appmgr/model/ApplicationManagerCommon.cpp
+++ b/module-services/service-appmgr/model/ApplicationManagerCommon.cpp
@@ -157,14 +157,13 @@ namespace app::manager
         ActionRequest act = ActionRequest{this->GetName(), app::manager::actions::DisplayLogoAtExit, nullptr};
         switch (closeReason) {
         case sys::CloseReason::SystemBrownout:
-            [[fallthrough]];
         case sys::CloseReason::LowBattery:
             act = ActionRequest{this->GetName(), app::manager::actions::SystemBrownout, nullptr};
             break;
         case sys::CloseReason::RegularPowerDown:
-            [[fallthrough]];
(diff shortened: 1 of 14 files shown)


wojtekrzepecki (approved):
LGTM

module-services/service-appmgr/model/ApplicationManagerCommon.cpp
```diff
@@ -157,14 +157,13 @@ namespace app::manager
...
         switch (closeReason) {
         case sys::CloseReason::SystemBrownout:
-            [[fallthrough]];
         case sys::CloseReason::LowBattery:
             act = ActionRequest{this->GetName(), app::manager::actions::SystemBrownout, nullptr};
             break;
         case sys::CloseReason::RegularPowerDown:
-            [[fallthrough]];
         case sys::CloseReason::Reboot:
-            [[fallthrough]];
```
SP2FET, line 167: why did you remove the [[fallthrough]] modifiers?

pawel-mudita: Because they seem to be used incorrectly.
The compiler may warn about unintentional fallthrough, if a `case` is ended without `break` or `[[fallthrough]]` and have at least one statement. With an empty `case` the compiler will not warn.

SP2FET: sure, isocpp says the same ;)
https://isocpp.org/files/papers/P0188R0.pdf

SP2FET (approved):

niepiekm (approved):

lucckb (approved):

PR #3536, opened 2022-02-01
[EGD-8227] Refactor quotes db

wojtekrzepecki:
Refactoring of quotes db layout to conform with given requirements:
- split to custom/predefined quoted databases
- cleaning & modifications in quotes db API
- Unit tests
- language support in quotes db


image/user/db/predefined_quotes_002-devel.sql
```diff
@@ -0,0 +1,60 @@
+-- Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
```
wojtekrzepecki, line 1: There will be proper set of quotes added for the production phone

image/user/db/custom_quotes_002-devel.sql
```diff
@@ -0,0 +1,10 @@
+-- Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
```
wojtekrzepecki, line 1: Set only for unit tests purposes

PrzeBrudny (approved):

image/user/db/predefined_quotes_002-devel.sql
```diff
@@ -0,0 +1,60 @@
...
+-- For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
+
+INSERT OR REPLACE INTO quote_languages (lang_id, lang_name) VALUES
+    (1, 'Polski'),
+    (2, 'English'),
+    (3, 'Espanol'),
+    (4, 'Francais'),
+    (5, 'Deutsch');
```
SP2FET, line 9: is adding a new language based on adding a new row with ID and string value of the language?

module-services/service-db/agents/quotes/QuotesSettingsSerializer.cpp
```diff
@@ -11,17 +11,28 @@ namespace Quotes
...
     {
         std::stringstream ss(listString);
         IdList list;
+        QuoteID quoteID;
+        bool type = true;
         for (int i; ss >> i;) {
-            list.push_back(i);
+            if (type) {
+                quoteID.first = static_cast<QuoteType>(i);
+                type          = false;
```
SP2FET, line 29: please use the `enum_cast` from the `magic_enum` - it handles the incompatibility.

module-services/service-db/agents/quotes/QuotesUtils.hpp
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
+// For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
+
+#pragma once
+
+namespace Quotes::utils
+{
+    unsigned getCurrentLanguageId();
+} // namespace Quotes::utils
```
SP2FET, line 9: no newline?

wojtekrzepecki: file deleted

image/user/db/predefined_quotes_002-devel.sql
wojtekrzepecki: Right now yes. The only thing to remember that it has to exactly the same as string of display language.

module-services/service-db/agents/quotes/QuotesSettingsSerializer.cpp
wojtekrzepecki: The `enum_cast` converts string to enum. Here I specifically wanted to keep it as int to provide minimal footprint as the whole serialized randomized order could be really long.

image/user/db/predefined_quotes_002-devel.sql
SP2FET: :+1: 
so maybe a comment would be useful there?

module-services/service-db/agents/quotes/QuotesSettingsSerializer.cpp
wojtekrzepecki: Also it is easier to extract it from stringstream as all values are ints.

SP2FET: right - my fault. But what if (by mistake) the value till be different than the enum scope? :D

SP2FET (approved):

module-services/service-db/agents/quotes/QuotesSettingsSerializer.cpp
wojtekrzepecki: Checked and it will cast it to int out of scope of enum which will be simply not processed afterwards.

PR #3548, opened 2022-02-03
[EGD-8251] Add DrawCommand unit test

SP2FET:
Added test for drawing the vector image


pholat (approved):
doesn't seem like it needs new/delete but seems fine :+1:

mjan-dev (approved):
LGTM

PR #3563, opened 2022-02-08
[MOS-144] Fix Increased CPU usage

Maciej-Mudita:
On the locked screen, while refreshing the screen,
the CPU woke up to the maximum frequency,
which resulted in an increased power consumption.

### DIFF
--- a/module-services/service-evtmgr/EventManager.cpp
+++ b/module-services/service-evtmgr/EventManager.cpp
@@ -56,18 +56,6 @@ namespace
     constexpr auto loggerTimerName = "Logger";
 } // namespace
 
-EventManagerSentinel::EventManagerSentinel(std::shared_ptr<sys::CpuSentinel> cpuSentinel,
-                                           bsp::CpuFrequencyMHz frequencyToHold)
-    : cpuSentinel(cpuSentinel)
-{
-    cpuSentinel->HoldMinimumFrequency(frequencyToHold);
-}
-
-EventManagerSentinel::~EventManagerSentinel()
-{
-    cpuSentinel->ReleaseMinimumFrequency();
-}
-
 EventManagerCommon::EventManagerCommon(LogDumpFunction logDumpFunction, const std::string &name)
     : sys::Service(name, "", stackDepth), loggerTimer{sys::TimerFactory::createPeriodicTimer(
                                               this,
@@ -232,7 +220,7 @@ sys::ReturnCodes EventManagerCommon::InitHandler()
     EventWorker->init(settings);
     EventWorker->run();
 
-    cpuSentinel                  = std::make_shared<sys::CpuSentinel>(service::name::evt_manager, this);
+    cpuSentinel                  = std::make_shared<sys::TimedCpuSentinel>(service::name::evt_manager, this);
     auto sentinelRegistrationMsg = std::make_shared<sys::SentinelRegistrationMessage>(cpuSentinel);
     bus.sendUnicast(sentinelRegistrationMsg, service::name::system_manager);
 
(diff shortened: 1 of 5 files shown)


module-services/service-evtmgr/EventManager.cpp
```diff
@@ -321,6 +321,10 @@ int EventManagerCommon::dumpLogsToFile()
 
 void EventManagerCommon::handleMinuteUpdate(time_t timestamp)
 {
+    // temporary frequency lock to handle time update event
+    if (cpuSentinel) {
+        cpuSentinel->HoldMinimumFrequencyForTime(bsp::CpuFrequencyMHz::Level_4, std::chrono::seconds(1));
```
mjan-dev, line 326: Isn't second too long? I believe it usually takes around ~300ms of processing to handle update. Won't keeping the sentinel for longer have negative effect on battery usage?

Maciej-Mudita: Immediately after processing the data (~ 300ms), a drawing on the screen takes place, which lasts ~ 5 seconds, so this 1 second reserve has no effect because at this time the eInk sentinel will be active.

module-sys/SystemManager/CpuSentinel.cpp
```diff
@@ -1,18 +1,27 @@
...
+    {
+        constexpr std::chrono::seconds defaultHoldFrequencyTime{1};
+    } // namespace
 
     CpuSentinel::CpuSentinel(std::string name,
                              sys::Service *service,
                              std::function<void(bsp::CpuFrequencyMHz)> callback)
-        : name(name), owner(service), callback(callback)
+        : name(name), owner(service),
+          callback(callback), timerHandle{sys::TimerFactory::createSingleShotTimer(
```
mjan-dev, line 21: Just asking. Wouldn't it be better to introduce a separate inheriting class like TimedCpuSentinel? Or there are some drawbacks with such approach here?

Maciej-Mudita: You're right. A separate sentinel would be a good idea here

module-services/service-evtmgr/EventManager.cpp
SP2FET, line 326: I think that we can go lower than Level 4 here - as I checked, the redraw + update for eink took max 1.5s when on 24MHz clock

module-sys/SystemManager/tests/unittest_CpuSentinelsGovernor.cpp
```diff
@@ -7,13 +7,42 @@
 #include <SystemManager/CpuGovernor.hpp>
 #include <SystemManager/CpuSentinel.hpp>
 
+namespace sys
+{
+    class MockedServicee : public sys::Service
```
SP2FET, line 12:
```suggestion
    class MockedService : public sys::Service
```

SP2FET (approved):

module-sys/SystemManager/CpuSentinel.cpp
Maciej-Mudita: done

module-services/service-evtmgr/EventManager.cpp
Maciej-Mudita: done

module-sys/SystemManager/tests/unittest_CpuSentinelsGovernor.cpp
Maciej-Mudita: done

mjan-dev (approved):

PR #3564, opened 2022-02-09
[CP-1073] Stack overflow in utils::filesystem::generateRandomId

niepiekm:
Allocating large objects on heap
Moved out generateRandomId() to utils namespace


module-utils/utility/Utils.cpp
```diff
@@ -99,4 +81,25 @@ namespace utils
         return s.str();
     }
 
+    std::string generateRandomId(std::size_t length) noexcept
```
alekrudnik, line 84: maybe some UT?

pawel-mudita (approved):

module-utils/utility/Utils.cpp
niepiekm: Added.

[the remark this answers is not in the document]
niepiekm: Support for it was not added to the source code. And it's out of scope of this fix.

PR #3565, opened 2022-02-09
[CP-1021] Add token field to Device Info API

niepiekm:
Generating a device unique id during on-boarding
Returning device token generated from device unique id in DeviceInfo
Added test for device token in test_devic_info.py

### DIFF
--- a/module-services/service-desktop/ServiceDesktop.cpp
+++ b/module-services/service-desktop/ServiceDesktop.cpp
@@ -51,6 +51,36 @@ auto ServiceDesktop::getSerialNumber() const -> std::string
     return settings->getValue(std::string("factory_data/serial"), settings::SettingsScope::Global);
 }
 
+auto ServiceDesktop::generateDeviceUniqueId() -> void
+{
+    const auto deviceUniqueId = utils::generateRandomId(sdesktop::DeviceUniqueIdLength);
+    LOG_SENSITIVE(LOGINFO, "Device unique id: %s", deviceUniqueId.c_str());
+    setDeviceUniqueId(deviceUniqueId);
+}
+
+auto ServiceDesktop::setDeviceUniqueId(const std::string &token) -> void
+{
+    return settings->setValue(sdesktop::DeviceUniqueIdName, token);
+}
+
+auto ServiceDesktop::getDeviceToken() -> std::string
+{
+    std::string tokenSeed = getDeviceUniqueId();
+
+    if (tokenSeed.empty()) {
+        LOG_DEBUG("Device unique id is empty, generating one...");
+        generateDeviceUniqueId();
+        tokenSeed = getDeviceUniqueId();
+    }
+
+    return (tokenSeed);
+}
+
+auto ServiceDesktop::getDeviceUniqueId() const -> std::string
+{
+    return settings->getValue(sdesktop::DeviceUniqueIdName);
+}
+
 auto ServiceDesktop::getCaseColour() const -> std::string
 {
     return settings->getValue(std::string("factory_data/case_colour"), settings::SettingsScope::Global);
--- a/module-services/service-desktop/include/service-desktop/ServiceDesktop.hpp
+++ b/module-services/service-desktop/include/service-desktop/ServiceDesktop.hpp
@@ -40,6 +40,8 @@ namespace sdesktop
     inline constexpr auto RECEIVE_QUEUE_BUFFER_NAME = "receiveQueueBuffer";
     inline constexpr auto SEND_QUEUE_BUFFER_NAME    = "sendQueueBuffer";
     inline constexpr auto IRQ_QUEUE_BUFFER_NAME     = "irqQueueBuffer";
+    inline constexpr auto DeviceUniqueIdLength      = 32;
+    constexpr auto DeviceUniqueIdName               = "sd_device_unique_id";
 
 }; // namespace sdesktop
 
@@ -83,8 +85,14 @@ class ServiceDesktop : public sys::Service
     auto getNotificationEntries() const -> std::vector<Outbox::NotificationEntry>;
     void removeNotificationEntries(const std::vector<int> &);
     void restartNotificationsClearTimer();
+    auto getDeviceToken() -> std::string;
 
   private:
+    auto getDeviceUniqueId() const -> std::string;
+
+    void generateDeviceUniqueId();
+    void setDeviceUniqueId(const std::string &token);
+
     std::unique_ptr<sdesktop::USBSecurityModel> usbSecurityModel;
     std::unique_ptr<settings::Settings> settings;
     std::unique_ptr<sdesktop::bluetooth::BluetoothMessagesHandler> btMsgHandler;
--- a/module-services/service-appmgr/include/service-appmgr/model/ApplicationManagerCommon.hpp
+++ b/module-services/service-appmgr/include/service-appmgr/model/ApplicationManagerCommon.hpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2021, Mudita Sp. z.o.o. All rights reserved.
+// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
 // For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 #pragma once
--- a/module-services/service-desktop/endpoints/include/endpoints/JsonKeyNames.hpp
+++ b/module-services/service-desktop/endpoints/include/endpoints/JsonKeyNames.hpp
@@ -52,6 +52,7 @@ namespace sdesktop::endpoints::json
     inline constexpr auto fileList       = "fileList";
     inline constexpr auto files          = "files";
     inline constexpr auto backupLocation = "backupLocation";
+    inline constexpr auto deviceToken    = "deviceToken";
 
     namespace updateprocess
     {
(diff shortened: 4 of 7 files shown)


niepiekm:
Please, ignore changes for CP-1073.

module-services/service-desktop/ServiceDesktop.cpp
```diff
@@ -53,6 +54,19 @@ auto ServiceDesktop::getSerialNumber() const -> std::string
...
+{
+    std::string tokenSeed = getDeviceUniqueId();
+    SHA256 sha256;
+
+    return sha256(tokenSeed);
+}
+
+auto ServiceDesktop::getDeviceUniqueId() const -> std::string
+{
+    return settings->getValue(std::string("gs_device_unique_id"), settings::SettingsScope::Global);
```
pawel-mudita, line 67: Not sure if ServiceDesktop should take the value of this setting directly via settings API.
Probably it should use API provided by ServiceAppMgr (owner of the setting) to obtain setting value.
Please double check with https://appnroll.atlassian.net/wiki/spaces/MFP/pages/1281917632/Settings+API+usage+guideline+followup#Usage-guideline

module-services/service-desktop/ServiceDesktop.cpp
```diff
@@ -53,6 +54,38 @@ auto ServiceDesktop::getSerialNumber() const -> std::string
     return settings->getValue(std::string("factory_data/serial"), settings::SettingsScope::Global);
 }
 
+auto ServiceDesktop::generateDeviceUniqueId() -> void
```
niepiekm, line 57: Moved generation of Device Unique Id outside on-boarding, because it won't work after OS update on the already on-boarded device.

module-services/service-desktop/ServiceDesktop.cpp
pawel-mudita: Yes, but I think this is for the owner of a specified setting.
In _Don'ts_ we have:
_use other applications/services settings - use API to get values_
TBH I'm not sure which way is correct.
Creating and using additional API in `ServiceAppMgr` seems like adding code bloat if we can access it directly.
On the other hand, by using `getValue()` on the value that is not owned by `ServiceDesktop` we are simply taking advantage of our poor system architecture. The truth is that this setting probably shouldn't be accessible in the global scope.

@pholat Could you advise here?
Can we use `settings->getValue` in `ServiceDesktop` for getting setting value owned by `ServiceAppMgr`?

pholat: tbh this discussion was done multiple times. This was partially meant to be solved by a proposition by Wiktor to create a static API for immediate access data. 
This way we could ask owners for their (owned) data. Not use global data everywhere.

But - for UID isn't there already API to get it? AFAIR UID data was loaded from separate memory to the settings then there was some wrapper to get it.

niepiekm: This UID is different from Serial Number, it changes after factory reset and is used as a secret substrate for MC to encrypt user data cache.

image/user/db/settings_v2_002.sql
```diff
@@ -44,7 +44,5 @@ INSERT OR IGNORE INTO settings_tab (path, value) VALUES
     ('gs_current_timezone_name', ''),
     ('gs_current_timezone_rules', ''),
     ('\ServiceTime\\gs_automatic_date_and_time_is_on', '1'),
-    ('\ServiceEink\\display_inverted_mode', '0');
-
-
-
+    ('\ServiceEink\\display_inverted_mode', '0'),
+    ('gs_device_unique_id', '');
```
niepiekm, line 48: @mjan-dev @pholat Looks like this addition is triggering some problems with settings after OS update. Do we have a plan for handling database updates?

wojtekrzepecki: If it is only in sql file it should not have impact on the database. Key will be created if nonexistent.

wojtekrzepecki:
Why devices cannot be distinguished by IMEI?

image/user/db/settings_v2_002.sql
niepiekm: I'll retest it.

niepiekm: Confirmed that writing to a newly added field in settings DB causes corruption of settings for a passcode (enabled flag and passcode value). The issue is observed after update to OS version with new field and after the filed was written to.

wojtekrzepecki: If it's going to be investigated, please report the bug and add reference to it here.

niepiekm:
> Why devices cannot be distinguished by IMEI?

The token is intended as an IV to encrypt user data cache in MC. It's expected to change after factory reset.

image/user/db/settings_v2_002.sql
niepiekm: Created and linked a bug https://appnroll.atlassian.net/browse/MOS-236

module-services/service-desktop/ServiceDesktop.cpp
pholat: ok, so why exactly is this UID global? not applocal? and did you check the devel.sql at the top to be working with what's added here? (i.e. with presenting it with some data, for example "2" or something?)

pholat (changes requested):

pholat:
please double check: (global setting value) x ( what's set in SQL file)

module-services/service-desktop/ServiceDesktop.cpp
```diff
@@ -51,6 +51,36 @@ auto ServiceDesktop::getSerialNumber() const -> std::string
...
+{
+    const auto deviceUniqueId = utils::generateRandomId(sdesktop::DeviceUniqueIdLength);
```
pholat, line 56: I believe this note was not tackled in other PR where I wrote it: why do we have some self-grown rand generation when we have TRNG on the chip? we use then this data for cache ciphering so I would assume the better the random number the better? :)

module-services/service-desktop/ServiceDesktop.cpp
niepiekm: Changed the UID to be AppLocal as suggested.

module-services/service-desktop/ServiceDesktop.cpp
niepiekm: It's a good suggestion and should be tracked under separate a task.

module-utils/utility/Utils.cpp
```diff
@@ -93,10 +93,11 @@ namespace utils
...
         std::uniform_int_distribution<> distribution(0, CHARACTERS.size() - 1);
 
-        std::string random_string(length, '\0');
+        std::string random_string;
+        random_string.reserve(length);
 
-        for (std::size_t i = 0; i < length; ++i) {
-            random_string[i] = CHARACTERS[distribution(*generator)];
+        while (length--) {
+            random_string += CHARACTERS[distribution(*generator)];
```
pholat, line 100: whats exactly changed here? :)

module-services/service-desktop/ServiceDesktop.cpp
pholat: and this is the second time my suggestion is closed without me knowing :) if it will be tracked, can you please at least link issue here?

module-utils/utility/Utils.cpp
pholat: didn't string constructor with size in parameter allocate string size if required?
TBC a brief look at the documentation didn't answer my question here but looking at the GCC basic_string header it seems it does. It would be weird if it didn't. 
```
      basic_string(size_type __n, _CharT __c, const _Alloc& __a = _Alloc())
      : _M_dataplus(_M_local_data(), __a)
      { _M_construct(__n, __c); }               <-- here it checks the size and allocates if required.
```

niepiekm: Of course, both methods have the same result. I chose `reserved` for reasons I already can't remember. :)

pholat: Ok, so if it does nothing: then don't change it if it's not needed.

pholat (commented):
:+1: ok

pholat (approved):

module-services/service-desktop/ServiceDesktop.cpp
```diff
@@ -51,6 +51,36 @@ auto ServiceDesktop::getSerialNumber() const -> std::string
...
+{
+    std::string tokenSeed = getDeviceUniqueId();
+
+    if (tokenSeed.empty()) {
+        LOG_DEBUG("Device unique id is empty, generating one...");
+        generateDeviceUniqueId();
+        tokenSeed = getDeviceUniqueId();
+    }
+
+    return (tokenSeed);
```
pawel-mudita, line 76:
```suggestion
    return tokenSeed;
```
Parentheses seem to be redundant here

module-services/service-desktop/include/service-desktop/ServiceDesktop.hpp
```diff
@@ -40,6 +40,8 @@ namespace sdesktop
     inline constexpr auto RECEIVE_QUEUE_BUFFER_NAME = "receiveQueueBuffer";
     inline constexpr auto SEND_QUEUE_BUFFER_NAME    = "sendQueueBuffer";
     inline constexpr auto IRQ_QUEUE_BUFFER_NAME     = "irqQueueBuffer";
+    inline constexpr auto DeviceUniqueIdLength      = 32;
+    constexpr auto DeviceUniqueIdName               = "sd_device_unique_id";
```
pawel-mudita, line 44:
```suggestion
    inline constexpr auto DeviceUniqueIdName               = "sd_device_unique_id";
```

pawel-mudita (approved):
Please check my comments before merging

pawel-mudita:
@niepiekm Why did you merge it while there are still open conversations?

niepiekm:
> @niepiekm Why did you merge it while there are still open conversations?

I enabled auto-merge.

PR #3568, opened 2022-02-09
Fix directory name spelling

GravisZro:
The spelling of directory `config/common_scripts_lib` is incorrect in two config scripts and preventing proper execution.

Could have sworn this got merged previously but here we are.


GravisZro:
Opps... I found the other ticket.

PR #3572, opened 2022-02-10
[BH-1385] Low battery screen on exit

paweljonskim:
Rework last render before shutdown to avoid races
Display low battery screen on exit

### DIFF
--- a/module-apps/apps-common/ApplicationCommon.cpp
+++ b/module-apps/apps-common/ApplicationCommon.cpp
@@ -125,6 +125,8 @@ namespace app
 
         connect(typeid(AppRefreshMessage),
                 [this](sys::Message *msg) -> sys::MessagePointer { return handleAppRefresh(msg); });
+        connect(typeid(AppShutdownRefreshMessage),
+                [this](sys::Message *msg) -> sys::MessagePointer { return handleAppShutdownRefresh(msg); });
         connect(sevm::BatteryStatusChangeMessage(), [&](sys::Message *) { return handleBatteryStatusChange(); });
         connect(typeid(app::manager::DOMRequest),
                 [&](sys::Message *msg) -> sys::MessagePointer { return handleGetDOM(msg); });
@@ -135,14 +137,14 @@ namespace app
 
         addActionReceiver(app::manager::actions::PhoneModeChanged, [this](auto &&params) {
             if (params != nullptr) {
-                auto modeParams = static_cast<gui::PhoneModeParams *>(params.get());
+                auto modeParams                  = static_cast<gui::PhoneModeParams *>(params.get());
                 this->statusIndicators.phoneMode = modeParams->getPhoneMode();
             }
             return actionHandled();
         });
         addActionReceiver(app::manager::actions::BluetoothModeChanged, [this](auto &&params) {
             if (params != nullptr) {
-                auto modeParams     = static_cast<gui::BluetoothModeParams *>(params.get());
+                auto modeParams                      = static_cast<gui::BluetoothModeParams *>(params.get());
                 this->statusIndicators.bluetoothMode = modeParams->getBluetoothMode();
                 refreshWindow(gui::RefreshModes::GUI_REFRESH_FAST);
             }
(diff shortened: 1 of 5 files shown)


wojtekrzepecki (approved):
I don't see anything suspicious but please test on Pure

pholat (changes requested):
why don't we fix the issue at the source, not like that? By "like that" I mean adding yet again a bool flag to the most basic data element in the GUI messaging process.

Actually, we give some time for the system to shut down, why don't we use it properly.
Why don't we check why isn't it showing last screen right? 

We had one boolean flag, now we have the same flag propagated in multiple places just to push it through.

module-apps/apps-common/ApplicationCommon.cpp
```diff
@@ -650,6 +652,32 @@ namespace app
...
+        if (windowsStack.isEmpty()) {
+            LOG_ERROR("Current window is not defined");
+            return sys::msgNotHandled();
+        }
+        if (windowsStack.isEmpty() || (getCurrentWindow()->getName() != msg->getWindowName())) {
+            LOG_DEBUG("Ignore shutdown refresh request for window %s we are on window %s",
+                      msg->getWindowName().c_str(),
+                      windowsStack.isEmpty() ? "none" : getCurrentWindow()->getName().c_str());
+            return sys::msgNotHandled();
+        }
```
pholat, line 668: 1. the First check is not that window is not defined. It's about if there is something else on the windows stack. I would say that adding a window builder to the application is closer to the window handling definition, than the created window on the stack.
2. This check is redundant, first, you check if isEmpty, then if is empty *or* (where you already checked isEmpty so it cant be empty)

module-apps/apps-common/ApplicationCommon.cpp
```diff
@@ -650,6 +652,32 @@ namespace app
...
+
+        // send drawing commands only when if application is in active and visible.
+        auto window = getCurrentWindow();
```
pholat, line 671: this comment is misleading, there was no check for active and visible here. Generally - it's not the place where it's checked either. What's mentioned in the comment is done in ApplicationManager.

module-apps/apps-common/ApplicationCommon.cpp
```diff
@@ -650,6 +652,32 @@ namespace app
...
+        updateStatuses(window);
+
+        auto message = std::make_shared<service::gui::DrawMessage>(getCurrentWindow()->buildDrawList(), msg->getMode());
```
pholat, line 674: `getCurrentWindow()->buildDrawList()` to `window->buildDrawList()`

module-apps/apps-common/ApplicationCommon.cpp
```diff
@@ -948,6 +976,40 @@ namespace app
         return true;
     }
 
+    void ApplicationCommon::requestShutdownWindow(std::string windowName)
```
pholat, line 979: and why exactly this and the following methods are needed?
Can't we just request for final refresh from the application scope via sending special "refresh" which is handled in: `handleAppShutdownRefresh` are these windows built somehow different than others? What do we gain from these methods?


Generally - why don't we handle only one, very specific refresh in the very specific window as a last.
i.e. if user declines to shutdown on request we still handle it as last refresh here?

pholat (changes requested):

module-apps/apps-common/ApplicationCommon.cpp
paweljonskim: what do you mean by "if used declines to shutdown"? - user has nothing to say about that - the system is going down when battery is low and he cant do anything to stop that :)

> Why dont we just send special refresh to be handled in `handleAppShutdownRefresh`?
I do agree that we can do that. It was done like that in previous review, I guess I misunderstood the 3rd point of your first review note. Gonna rework that

pholat: Please respond to all the questions :)

pholat (approved):
assuming it works fine in later parts ( service::Renderer -> Service::GUI ) and therefore fixes race issue, I would say it's fine now ;)

PrzeBrudny (approved):

PR #3578, opened 2022-02-11
[MOS-104] Fix calllog responsivness with large number of contacts

alekrudnik:
Determined 3 reasons of lag
1. each calllog app opening is sending db::query::calllog::SetAllRead
which marks all calllog database entries as read and send CRUD
notification to application and hence the list view is rebuilt at least
twice for each app opening.
After deeper investigation it seems like dead feature (at least not
covered in GUI) and I am going to disable it.

2. DBServiceAPI::MatchContactByPhoneNumber requested for empty string
(private number) take really long time (above 1s).

3. DBServiceAPI::MatchContactByPhoneNumber took about time 150-200ms

This patch fixes only issues 1. and 2. Number 3. requires further
investigation and is consider as separate issue.

After mentioned fixes calllog is pretty responsive but requires ~2s
for each app start or page change.

### DIFF
--- a/module-apps/application-calllog/ApplicationCallLog.cpp
+++ b/module-apps/application-calllog/ApplicationCallLog.cpp
@@ -1,4 +1,4 @@
-﻿// Copyright (c) 2017-2021, Mudita Sp. z.o.o. All rights reserved.
+﻿// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
 // For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 #include "ApplicationCallLog.hpp"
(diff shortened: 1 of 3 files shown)


SP2FET (approved):

pholat (approved):

mjan-dev (approved):

module-apps/application-calllog/ApplicationCallLog.cpp
```diff
@@ -133,9 +133,7 @@ namespace app
         DBServiceAPI::GetQuery(this,
                                db::Interface::Name::Notifications,
                                std::make_unique<db::query::notifications::Clear>(NotificationsRecord::Key::Calls));
-        const auto [succeed, _] = DBServiceAPI::GetQuery(
-            this, db::Interface::Name::Calllog, std::make_unique<db::query::calllog::SetAllRead>());
-        return succeed;
```
wojtekrzepecki, line 138: This changes the logic a bit -> is it consulted?

PrzeBrudny (approved):

module-apps/application-calllog/ApplicationCallLog.cpp
alekrudnik: as stated in the PR description 
`ach calllog app opening is sending db::query::calllog::SetAllRead
which marks all calllog database entries as read and send CRUD
notification to application and hence the list view is rebuilt at least
twice for each app opening.
After deeper investigation it seems like dead feature (at least not
covered in GUI) and I am going to disable it.`

wojtekrzepecki (approved):

PR #3579, opened 2022-02-11
[MOS-143] Fix notifications for unread message/missed calls

niepiekm:
Refactored handling of notifications on home/lock creen

### DIFF
--- a/module-apps/application-messages/windows/SMSThreadViewWindow.cpp
+++ b/module-apps/application-messages/windows/SMSThreadViewWindow.cpp
@@ -1,4 +1,4 @@
-﻿// Copyright (c) 2017-2021, Mudita Sp. z.o.o. All rights reserved.
+﻿// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
 // For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 #include "ApplicationMessages.hpp"
@@ -7,6 +7,7 @@
 #include "SMSThreadViewWindow.hpp"
 
 #include <log/log.hpp>
+#include <module-db/queries/notifications/QueryNotificationsDecrement.hpp>
 #include <module-db/queries/phonebook/QueryContactGetByNumberID.hpp>
 #include <module-db/queries/phonebook/QueryNumberGetByID.hpp>
 #include <OptionsMessages.hpp>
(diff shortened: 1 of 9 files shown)


module-db/queries/notifications/QueryNotificationsIncrement.hpp
```diff
@@ -7,19 +7,22 @@
...
         const utils::PhoneNumber::View number;
+        const ssize_t count;
 
       public:
-        Increment(NotificationsRecord::Key key, const utils::PhoneNumber::View &number);
+        Increment(NotificationsRecord::Key key, const utils::PhoneNumber::View &number, const ssize_t count = 1);
 
         [[nodiscard]] auto getKey() const noexcept -> NotificationsRecord::Key;
         [[nodiscard]] auto getNumber() const noexcept -> const utils::PhoneNumber::View &;
+        [[nodiscard]] auto getCount() const noexcept -> ssize_t;
```
pholat, line 25: now we can have a negative count?

module-db/Interface/NotificationsRecord.cpp
```diff
@@ -170,7 +170,7 @@ std::unique_ptr<db::query::notifications::GetResult> NotificationsRecordInterfac
 std::unique_ptr<db::query::notifications::IncrementResult> NotificationsRecordInterface::runQueryImpl(
     const db::query::notifications::Increment *query)
 {
-    auto ret = processIncrement(query->getKey(), query->getNumber(), 1);
+    auto ret = processIncrement(query->getKey(), query->getNumber(), query->getCount());
```
pholat, line 173: so what will happen here when we get the negative count? :)

module-db/queries/notifications/QueryNotificationsIncrement.hpp
niepiekm: Well, it's a first attempt at abstracting the decrement operation. Ideally, we'd have a facade for incrementing and decrementing particular notifications.

module-db/Interface/NotificationsRecord.cpp
mjan-dev: Is it possible that query returns negative count? I haven't checked all getCount methods, however it seems that all of those return unsigned value. Nevertheless, we could check for it, just in case.

module-apps/application-messages/models/SMSThreadModel.cpp
```diff
@@ -90,6 +92,15 @@ auto SMSThreadModel::handleQueryResponse(db::QueryResult *queryResult) -> bool
 
     number = std::make_unique<utils::PhoneNumber::View>(msgResponse->getNumber());
 
+    if (number) {
+        auto &threadNumber = static_cast<utils::PhoneNumber::View &>(*number);
+
+        DBServiceAPI::GetQuery(
+            application,
+            db::Interface::Name::Notifications,
+            std::make_unique<db::query::notifications::Increment>(NotificationsRecord::Key::Sms, threadNumber, -1));
```
niepiekm, line 101: Sadly, this solution doesn't work as expected: notifications are added for each new (unread) message, but the clearing of notifications has to occur for each thread - and we need to know how many unread messages are there in a thread. We don't keep that information.

alekrudnik: AFAIR this information is stored in different data base. Please take a look at ThreadRecord and unreadMsgCount.

niepiekm: We figured this eventually after discussing with @PrzeBrudny

module-db/queries/notifications/QueryNotificationsIncrement.hpp
niepiekm: We can pass a negative count.

module-apps/application-messages/models/SMSThreadModel.cpp
niepiekm: Resolved

module-db/Interface/NotificationsRecord.cpp
niepiekm: In case of a negative count, the value of notifications in the record will be decremented.

module-apps/application-messages/windows/SMSThreadViewWindow.cpp
```diff
@@ -62,29 +63,34 @@ namespace gui
...
-                smsList->rebuildList();
-            }
-            else if (smsModel->numberID != DB_ID_NONE) {
-                requestContact(smsModel->numberID);
             }
+            smsModel->numberID    = pdata->thread->numberID;
+            smsModel->smsThreadID = pdata->thread->ID;
+            smsList->rebuildList();
+
+            ssize_t countToClear = -pdata->thread->unreadMsgCount;
```
alekrudnik, line 82: if it is ssize_t please check if not negative before clearing.

niepiekm: `countToClear` is expected to be a negative value of `pdata->thread->unreadMsgCount`

niepiekm: This is reworked now - added special query to handle decrementing of notification count.

module-db/queries/notifications/QueryNotificationsIncrement.hpp
niepiekm: Negative values are no longer needed. Reworked - added special query to handle decrementing of notification count.

alekrudnik (approved):

module-db/Interface/NotificationsRecord.cpp
```diff
@@ -243,3 +255,22 @@ bool NotificationsRecordInterface::processIncrement(NotificationsRecord::Key key
 
     return ret;
 }
+
+bool NotificationsRecordInterface::processDecrement(NotificationsRecord::Key key, size_t delta)
+{
+    auto ret = false;
+
+    if (auto record = GetByKey(key); record.isValid() && record.value > 0) {
+        if ((ssize_t(record.value) - delta <= 0)) {
```
pholat, line 264: why cast to unsigned? :) doesn't seem like it's needed?

pholat (approved):
please fix the cast (if it's even needed for anything) and if so - add static_cast there explicitly. 
Add UT for that case too.

module-db/Interface/NotificationsRecord.cpp
niepiekm: Fixed as discussed.

niepiekm:
> please fix the cast (if it's even needed for anything) and if so - add static_cast there explicitly. Add UT for that case too.

Done.

PR #3585, opened 2022-02-15
[CP-1033] Add Outbox endpoint to service-desktop

pawel-mudita:
Outbox endpoint will be used to provide notifications
about changes in messages/threads/calls during connection
with Mudita Center. This mechanism will ensure an efficient
data synchronization between Mudita Pure and Mudita Center.

### DIFF
--- a/module-services/service-desktop/endpoints/include/endpoints/EndpointType.hpp
+++ b/module-services/service-desktop/endpoints/include/endpoints/EndpointType.hpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2021, Mudita Sp. z.o.o. All rights reserved.
+// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
 // For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 #pragma once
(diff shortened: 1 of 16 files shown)


module-services/service-desktop/parser/MessageHandler.cpp
```diff
@@ -37,12 +38,17 @@ namespace sdesktop::endpoints
...
             LOG_ERROR("No way to handle!");
+            return;
+        }
+        handler->handle(*context);
+        auto ownerService = dynamic_cast<ServiceDesktop *>(OwnerServicePtr);
+        if (ownerService == nullptr) {
+            LOG_ERROR("Downcasting error!");
+            return;
         }
+        ownerService->restartNotificationsClearTimer();
```
pholat, line 51: why not just bind ownerService to the ServiceDesktop? or better: pass dependencies in ctor, in this case: callback to `clear`. or: better: if timer is required by this scope: why not create it here?

pholat (approved):
seems ok, would be nice if you applied the comment :bow:

module-services/service-desktop/parser/MessageHandler.cpp
pawel-mudita: Done by passing callback

module-services/service-desktop/endpoints/include/endpoints/EndpointType.hpp
```diff
@@ -21,9 +21,10 @@ namespace sdesktop::endpoints
         calendarEventsPlaceholder,
         developerMode,
         bluetooth,
-        usbSecurity
+        usbSecurity,
+        outbox
     };
 
-    inline constexpr auto lastEndpoint = static_cast<int>(EndpointType::usbSecurity);
+    inline constexpr auto lastEndpoint = static_cast<int>(EndpointType::outbox);
```
pholat, line 28: just saying - magic_enum in the repo has the capability to get enum length. I do not know if I would advise using it in the header (as it's heavy on compilation) but it's nice to know :)

pholat (approved):

PR #3586, opened 2022-02-15
[MOS-97] Fix of wrong label for missed call notification

alekrudnik:
There was no label on missed call notification from private number.
After fix it is diplaying "Missed calls" info.

Original fix from 60bdca524338025dc0117edcc5c5579376441b0b [MOS-104]
was also modifed as it previously introduced regression on menu window.

[MOS-104]: https://appnroll.atlassian.net/browse/MOS-104?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/module-apps/application-calllog/ApplicationCallLog.cpp
+++ b/module-apps/application-calllog/ApplicationCallLog.cpp
@@ -133,7 +133,9 @@ namespace app
         DBServiceAPI::GetQuery(this,
                                db::Interface::Name::Notifications,
                                std::make_unique<db::query::notifications::Clear>(NotificationsRecord::Key::Calls));
-        return true;
+        const auto [succeed, _] = DBServiceAPI::GetQuery(
+            this, db::Interface::Name::Calllog, std::make_unique<db::query::calllog::SetAllRead>());
+        return succeed;
     }
 
 } /* namespace app */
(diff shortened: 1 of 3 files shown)


module-apps/application-calllog/ApplicationCallLog.cpp
```diff
@@ -133,7 +133,9 @@ namespace app
         DBServiceAPI::GetQuery(this,
                                db::Interface::Name::Notifications,
                                std::make_unique<db::query::notifications::Clear>(NotificationsRecord::Key::Calls));
-        return true;
+        const auto [succeed, _] = DBServiceAPI::GetQuery(
+            this, db::Interface::Name::Calllog, std::make_unique<db::query::calllog::SetAllRead>());
+        return succeed;
```
SP2FET, line 138: so indeed it was needed?

SP2FET (approved):

wojtekrzepecki (approved):

module-apps/application-calllog/ApplicationCallLog.cpp
alekrudnik: yep, I missed it. It was needed for dot in menu window, which is marker of unread calls in calllog

PR #3591, opened 2022-02-16
[MOS-227] Fix DnD first call crash

wojtekrzepecki:
fixed call when making first call from unknown number when Pure is in DnD mode


mjan-dev (approved):

SP2FET (approved):
if there's a possibility to get a nullptr in contact then it might be screwed up somewhere in the upper layer, wouldn't it?
But I know, it's only a fix :D

kkleczkowski (approved):

pholat (approved):

wojtekrzepecki:
@SP2FET  I have checked and other usages are checked on nullptr's on the upper layers. Of course the interface could be better but it requires some bigger refactor.

PR #3593, opened 2022-02-16
[MOS-111] Added missing refresh conditions

PrzeBrudny:
Added missing refresh conditions.

### DIFF
--- a/module-services/service-desktop/WorkerDesktop.cpp
+++ b/module-services/service-desktop/WorkerDesktop.cpp
@@ -147,8 +147,8 @@ bool WorkerDesktop::handleMessage(uint32_t queueID)
 
         if (notification == bsp::USBDeviceStatus::Connected) {
             usbSuspendTimer.stop();
-            ownerService->bus.sendUnicast(std::make_shared<sdesktop::usb::USBConnected>(),
-                                          service::name::service_desktop);
+            ownerService->bus.sendMulticast(std::make_shared<sdesktop::usb::USBConnected>(),
+                                            sys::BusChannel::USBNotifications);
             usbStatus = bsp::USBDeviceStatus::Connected;
         }
         else if (notification == bsp::USBDeviceStatus::Configured) {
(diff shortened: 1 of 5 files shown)


wojtekrzepecki (approved):

mjan-dev (approved):

Maciej-Mudita (approved):

niepiekm (approved):

module-services/service-desktop/WorkerDesktop.cpp
```diff
@@ -168,8 +167,8 @@ bool WorkerDesktop::handleMessage(uint32_t queueID)
         else if (notification == bsp::USBDeviceStatus::Disconnected) {
             usbSuspendTimer.start();
             cpuSentinel->ReleaseMinimumFrequency();
-            ownerService->bus.sendUnicast(std::make_shared<sdesktop::usb::USBDisconnected>(),
-                                          service::name::service_desktop);
+            ownerService->bus.sendMulticast(std::make_shared<sdesktop::usb::USBDisconnected>(),
```
pholat, line 170: This is essentially broadcast when you send it to sys::BusChannel::System.
Please add the Multicast channel as it was done for another multicast.
You are free to remove now unused channels if you wish :)

At least we would have a conformant code. Here: it's **the only** multicast that instead of having its own channel, just writes to the system channel.

pholat (changes requested):

module-services/service-desktop/WorkerDesktop.cpp
swichu91: Apart from the good points made by @pholat, I would like to say that, IMHO, it is not the responsibility of the service desktop to send USB notifications to the system. Just think about it in the context of SOLID principles, especially the single responsibility principle. 
Are we sure we want to inject such a responsibility into the unrelated module? 
I know that it was the quickest and the easiest thing to do but do we have to follow this route each time? 
How about at least moving it to the event manager? We could even have a small separate USB-related service.

PrzeBrudny: Sure can add channel.

pholat (approved):
ok :+1:

PR #3610, opened 2022-02-22
[MOS-16] Change favourite contacts font to regular

pawel-mudita:
The font was bold, which was inconsistent with the designs.

https://projects.invisionapp.com/d/main?origin=v7#/console/18241335/381446305/preview?scrollOffset=8749

Before:
favorites_before

After:
fav_after

### DIFF
--- a/module-apps/application-phonebook/widgets/PhonebookItem.cpp
+++ b/module-apps/application-phonebook/widgets/PhonebookItem.cpp
@@ -36,13 +36,12 @@ namespace gui
         };
     }
 
-    // sets copy of alarm's
-    void PhonebookItem::setContact(std::shared_ptr<ContactRecord> contact)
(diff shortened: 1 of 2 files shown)


module-apps/application-phonebook/widgets/PhonebookItem.cpp
```diff
@@ -65,10 +64,6 @@ namespace gui
     void PhonebookItem::markFavourite(bool val)
```
mjan-dev, line 64: Does markFavourite has any role after these changes?

pawel-mudita: Yes, but actually it is used only once and there is no need to keep it as a function anymore.
Changed

mjan-dev (approved):

PR #3611, opened 2022-02-22
Added Chinese localization

gabryatfendor:
Added Chinese localization by FantasyJi


pholat:
@mjan-dev we can actually add Chinese font and show it (it's separated from the input so it's not a big issue)
@gabryatfendor could you please provide us with a utf8 list of characters we would require to support? If not we can just export all that are in this translation, but it might not be enough. On the other hand - we do not know Chinese, but we think that loading to many characters might be a problem too.

We might create a font and test it then. Then we can plan to i.e. add package to optionally load the Chinese font :)

gabryatfendor:
@pholat unfortunately I don't know Chinese, I just copied the translation provided by FantasyJi privately

PR #3618, opened 2022-02-24
[MOS-221] DND - missing notifications

niepiekm:
Sending notification for missing call in DnD mode


kkleczkowski (approved):

module-services/service-cellular/ServiceCellular.cpp
```diff
@@ -1927,9 +1927,17 @@ auto ServiceCellular::handleCellularRingingMessage(CellularRingingMessage *msg)
...
-        CellularServiceAPI::DismissCall(
-            this, phoneModeObserver->getCurrentPhoneMode() == sys::phone_modes::PhoneMode::DoNotDisturb);
+        CellularServiceAPI::DismissCall(this, dndModeOn);
+    }
+
+    if (dndModeOn) {
+        DBServiceAPI::GetQuery(
+            this,
+            db::Interface::Name::Notifications,
+            std::make_unique<db::query::notifications::Increment>(NotificationsRecord::Key::Calls, message->number));
```
pholat, line 1940: why not instead adding lines to 2k+ ServiceCellular.cpp not move it to ongoingCall? 
If not for not extending this large enough file, to maybe keep it in one place?
especially that you would be able to write it expressively:
Please see: CallRing and CallGUI and follow them. Maybe add CallNotifications and guard for it?

pholat (changes requested):

module-services/service-cellular/ServiceCellular.cpp
niepiekm: I have too little knowledge about the purpose of the `Call` class and its ties to `ServiceCellular`. From what I established, this is an optimal place for putting requited fix.

pholat: ok, what needs to be done so that you would have a better perspective?
Now we have CallManager (which I would call: wrapped globals to sell nicely), Call (singular call management class), and random handlings in ServiceCellular.cpp scattered.
Adding more code here doesn't improve the overall state, does it?
There is no agreement to refactor it - therefore the only thing we can do is make things better, one thing at a time :(

pholat (changes requested):

module-services/service-cellular/ServiceCellular.cpp
niepiekm: From what I see in the `Call` class, the `false` branch of `handleCLIP()` method seems the best place to put the notification broadcast. Let me check that.

pholat: Call class is class detached from ServiceCellular to actually declutter `class handling everything.

niepiekm: Done.

PR #3626, opened 2022-03-02
[CP-1103] Harmony device log is not sent to support

niepiekm:
Added returning of diagnostic data info to Harmony DeviceInfo EP


products/BellHybrid/services/desktop/endpoints/deviceInfo/DeviceInfoEndpoint.cpp
```diff
@@ -23,13 +24,98 @@ namespace sdesktop::endpoints
...
+    auto DeviceInfoEndpoint::handleGet(Context &context) -> void
+    {
+        const auto &requestBody = context.getBody();
+
+        if (requestBody[json::fileList].is_number()) {
+
+            const auto diagFileType = static_cast<DiagnosticFileType>(requestBody[json::fileList].int_value());
+
+            if (!magic_enum::enum_contains<DiagnosticFileType>(diagFileType)) {
+                LOG_ERROR("Bad diagnostic type %d requested", static_cast<unsigned>(diagFileType));
```
pholat, line 45: why not (returning enum):
``` 
const auto diagFileType = parseDiagnosticFileType(requestBody[jeson::fileList]);
```
then just use the diagFileType in log.
Especially that you can just use a magic enum to print string evaluation of that enum here. But there would be cast on cast here then.

This way we would have:
- separate parsing logic from switch case handling
- one place where casting is done instead  is in analysis if it's in, then cast )
- more self explaining, readable code :)

products/BellHybrid/services/desktop/endpoints/deviceInfo/DeviceInfoEndpoint.cpp
```diff
@@ -23,13 +24,98 @@ namespace sdesktop::endpoints
...
+
+                context.setResponseStatus(http::Code::BadRequest);
```
pholat, line 47: please return from functions instead of setting "globals" (we can think of context as a "global void"), this way we would have hope to someday change this code to something better.

test/get_os_log.py
```diff
@@ -17,7 +18,10 @@ def set_passcode(harness: Harness, flag: bool):
     '''
     on exit -> restore PhoneModeLock
     '''
-    PhoneModeLock(flag).run(harness)
+    try:
+        PhoneModeLock(flag).run(harness)
+    except:
+        pass
```
pholat, line 24: this is plain wrong.

products/BellHybrid/services/desktop/endpoints/deviceInfo/DeviceInfoEndpoint.cpp
```diff
@@ -23,13 +24,98 @@ namespace sdesktop::endpoints
...
+        json11::Json::array fileArray;
+
+        for (const auto &file : fileList) {
+            fileArray.push_back(file);
+        }
+
+        return json11::Json::object{{json::files, fileArray}};
+    }
+
+    auto DeviceInfoEndpoint::listDirectory(std::string path) -> std::vector<std::string>
```
pholat, line 110: is it even needed to wrap directory_iterator to std::vector(string) here? isn't it just more code?

test/get_os_log.py
niepiekm: This is a poor-man handler of the case when a passcode is not enabled on the phone. Sure, there should be some information about why we ignored the exception.

products/BellHybrid/services/desktop/endpoints/deviceInfo/DeviceInfoEndpoint.cpp
niepiekm: Done

test/get_os_log.py
niepiekm: Thank you. Fixed.

products/BellHybrid/services/desktop/endpoints/deviceInfo/DeviceInfoEndpoint.cpp
niepiekm: It's been a while since the original implementation, so I don't remember the particular reason I did the wrapper but am sure that without it the code didn't work as expected..

module-services/service-desktop/endpoints/include/endpoints/deviceInfo/DeviceInfoEndpointCommon.hpp
```diff
@@ -0,0 +1,49 @@
...
+        auto listDirectory(std::string path) -> std::vector<std::string>;
+        auto fileListToJsonObject(const std::vector<std::string> &fileList) const -> json11::Json::object const;
+        auto requestLogsFlush() const -> void;
+        auto getStorageStats(const std::string &path) -> std::tuple<long, long>;
+        auto getStorageInfo() -> std::tuple<long, long, long>;
+        auto getDeviceToken() -> std::string;
+
+        static constexpr auto OS_RESERVED_SPACE_IN_MB = (1024LU);
+
+      public:
```
pawel-mudita, line 37: Why `public` access specifier is added two times here?

module-services/service-desktop/endpoints/include/endpoints/deviceInfo/DeviceInfoEndpointCommon.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+        auto handleGet(Context &context) -> http::Code;
+        auto gatherListOfDiagnostics(Context &context, DiagnosticFileType diagDataType) -> http::Code;
+        auto parseDiagnosticFileType(const json11::Json &fileList) -> DiagnosticFileType;
+        auto listDirectory(std::string path) -> std::vector<std::string>;
+        auto fileListToJsonObject(const std::vector<std::string> &fileList) const -> json11::Json::object const;
+        auto requestLogsFlush() const -> void;
+        auto getStorageStats(const std::string &path) -> std::tuple<long, long>;
+        auto getStorageInfo() -> std::tuple<long, long, long>;
+        auto getDeviceToken() -> std::string;
```
pawel-mudita, line 33: This method is not implemented in `DeviceInfoEndpointCommon` class.
Maybe it should be virtual?

pawel-mudita (approved):
Please check my comments before merging

module-services/service-desktop/endpoints/include/endpoints/deviceInfo/DeviceInfoEndpointCommon.hpp
niepiekm: Right, this is a Pure-specific method and shouldn't be defined here.

module-services/service-desktop/endpoints/include/endpoints/deviceInfo/DeviceInfoEndpointCommon.hpp
niepiekm: Fixed

test/get_os_log.py
```diff
@@ -7,11 +7,21 @@
...
 from harness.api.developermode import PhoneModeLock
+from harness.api.security import GetPhoneLockStatus
 from harness.api.filesystem import get_log_file, get_log_file_with_path
 from harness.api.device_info import *
 
+def is_phone_locked(harness: Harness):
+    try:
+        GetPhoneLockStatus().run(harness)
+    except TransactionError as error:
+        return error.status == status["Forbidden"]
```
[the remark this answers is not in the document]
niepiekm: This is not a test case, it's just a dev-tool to get OS logs from the device so we don't need to re-raise any exceptions.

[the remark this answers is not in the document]
pholat: First of all - is the new `is_phone_locked` needed in the first place? there is already such function in harness isn't there?
Secondly - I don't think that writing code that might be misleading is ok anywhere. (I remind that it at first was `catch all` )

pholat (approved):

test/get_os_log.py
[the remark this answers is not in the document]
niepiekm: `is_phone_locked` method defined in `CDCSerial` class uses DeveloperMode EP. This impl. uses Security EP. As this tool is for development only, I don't want to redefine original or add another to Harness.

PR #3628, opened 2022-03-02
[MOS-273] Update Harness to detect Mudita Bell devices

niepiekm:
Added detection of Mudita Harmony

### DIFF
--- a/test/pytest/conftest.py
+++ b/test/pytest/conftest.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2017-2021, Mudita Sp. z.o.o. All rights reserved.
+# Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
 # For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 import time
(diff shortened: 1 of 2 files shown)


mjan-dev (approved):

test/pytest/conftest.py
```diff
@@ -121,7 +121,7 @@ def harnesses():
     '''
     Automatically init at least two Pure phones
     '''
-    found_pures = serial.find_Pures()
+    found_pures = serial.find_Devices()
```
SP2FET, line 124: so now it should be renamed :D
Also - wouldn't adding a Harmony on the devices lists break pure-specific tests?

niepiekm: The method is renamed to indicate a new purpose - it now looks not only for Pure :)

If we run Pure-specific tests on Harmony they will fail, obviously. 
Making harness work with Harmony is not necessary for development, I just noticed the lack of it while testing CP-1103. This change can be postponed.

SP2FET: i meant the variable's name :D

niepiekm: Fixed :)

SP2FET (approved):

PR #3637, opened 2022-03-09
[MOS-45] Invalid number decoding

niepiekm:
Added UT case for failing phone number
Fixed decimal semi-octet decoding (AKA ascii combination)

### DIFF
--- a/module-utils/unicode/utf8/test/unittest_utf8.cpp
+++ b/module-utils/unicode/utf8/test/unittest_utf8.cpp
@@ -341,9 +341,26 @@ TEST_CASE("UTF8: insert whole string which doesn't work")
 
 TEST_CASE("UTF8: Convert to ascii if is ascii combination")
 {
-    UTF8 combination = "778568738465";
-    REQUIRE(combination.isASCIICombination());
-    REQUIRE(combination.toASCII() == std::string("MUDITA"));
+    SECTION("6 characters")
+    {
+        UTF8 combination = "778568738465";
+        REQUIRE(combination.isASCIICombination());
+        REQUIRE(combination.toASCII() == std::string("MUDITA"));
+    }
+
+    SECTION("4 characters")
+    {
+        UTF8 combination = "100101101112";
+        REQUIRE(combination.isASCIICombination());
+        REQUIRE(combination.toASCII() == std::string("deep"));
+    }
+
+    SECTION("2 characters")
+    {
+        UTF8 combination = "111107";
+        REQUIRE(combination.isASCIICombination());
+        REQUIRE(combination.toASCII() == std::string("ok"));
+    }
 }
 
 TEST_CASE("UTF8: Not ASCII combination")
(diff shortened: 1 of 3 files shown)


module-utils/unicode/utf8/test/unittest_utf8.cpp
```diff
@@ -356,6 +356,13 @@ TEST_CASE("UTF8: Not ASCII combination")
         REQUIRE_FALSE(combination.toASCII().has_value());
     }
 
+    SECTION("Carier info number - 3 digits")
```
mjan-dev, line 359: Could you please add more UTs? For example with other 3/2/1 digits combinations.

niepiekm: Working on it.

module-utils/unicode/utf8/test/unittest_utf8.cpp
```diff
@@ -356,6 +356,13 @@ TEST_CASE("UTF8: Not ASCII combination")
...
+    {
+        UTF8 combination = "100";
```
pholat, line 361: please add tests for at least: 
for negative cases:
- 1111  
- 11111 
- 111111
positive cases
- 11147
- 8411
- 11182115

module-utils/unicode/utf8/utf8/UTF8.cpp
```diff
@@ -798,7 +798,7 @@ bool UTF8::isASCIICombination() const noexcept
     constexpr uint8_t secondCharacterFactor = 10;
     for (; i < len; i += 2) {
         int firstCharacter = 0;
-        if (data[i] == '1') {
+        if (len > 3 && data[i] == '1') {
```
niepiekm, line 801: Considering what I found out from the spec and observation of Quectel eval board, this seems to be just a workaround this particular problem with `100` number. I'm yet to figure out how to methodically distinguish a number consisting from digits from alphanumeric encoded using BSD.

module-utils/unicode/utf8/test/unittest_utf8.cpp
mjan-dev: Ok, I see added test cases, thanks.

module-utils/unicode/utf8/utf8/UTF8.cpp
mjan-dev: Please add comment clearly stating that it is a workaround for this particular case.

niepiekm: Done

module-utils/unicode/utf8/test/unittest_utf8.cpp
niepiekm: Added representative TCs

mjan-dev (approved):
Approving as a workaround for this particular case, not a systematic solution to the problem.

kkleczkowski (approved):

PR #3649, opened 2022-03-15
[MOS-288] Add community build

paweljonskim:
Główne zmiany:
- Style.hpp
- FontManager.hpp
- FontManager.cpp
- products/BellHybrid/CMakeLists.txt
- image/BellHybrid/CMakeLists.txt


module-gui/gui/core/FontManager.cpp
```diff
@@ -33,29 +53,31 @@ namespace gui
...
     {
-        fontFolder                         = baseDirectory + "/fonts";
-        std::vector<std::string> fontFiles = getFontsList();
+        fontFolder     = baseDirectory + "/fonts";
+        fontMapFile    = fontFolder + "/fontmap.json";
+        auto fontFiles = getFontsList();
 
-        for (std::string fontName : fontFiles) {
-            loadFont(fontName);
+        for (auto font : fontFiles) {
```
swichu91, line 60:
```suggestion
        for (const auto& font : fontFiles) {
```

module-gui/gui/core/FontManager.cpp
```diff
@@ -33,29 +53,31 @@ namespace gui
...
+            loadFont(font.first, font.second);
         }
     }
 
-    RawFont *FontManager::loadFont(std::string filename)
+    RawFont *FontManager::loadFont(Fonts font, std::string path)
```
swichu91, line 65:
```suggestion
    RawFont *FontManager::loadFont(Fonts font, const std::string& path)
```

module-gui/gui/core/FontManager.cpp
```diff
@@ -96,28 +119,61 @@ namespace gui
...
+            else {
+                LOG_INFO("Add font to list: %s - %s",
+                         std::string(magic_enum::enum_name(entry.first)).c_str(),
+                         fontName.c_str());
+                fontFiles.insert({entry.first, fontFolder + "/" + fontName});
             }
         }
-
         LOG_INFO("Total number of fonts: %u", static_cast<unsigned int>(fontFiles.size()));
         return fontFiles;
```
swichu91, line 164: `std::fclose(fd)` missing?

module-gui/gui/core/FontManager.cpp
```diff
@@ -96,28 +119,61 @@ namespace gui
...
+        defaultFontTypeName = infoJson["default_font_type_name"].string_value();
 
-        LOG_INFO("Scanning fonts folder: %s", fontFolder.c_str());
+        auto styleJson = fontmapObjects["style"];
+        std::map<Fonts, std::string> fontFiles;
 
-        for (const auto &entry : std::filesystem::directory_iterator(fontFolder)) {
-            if (!std::filesystem::is_directory(entry) && entry.path().extension() == ".mpf") {
-                fontFiles.push_back(entry.path().string());
+        for (auto &entry : fonts_map::mapJson) {
```
swichu91, line 151:
```suggestion
        for (const auto &entry : fonts_map::mapJson) {
```

cmake/modules/DownloadAsset.cmake
```diff
@@ -65,3 +65,41 @@ function(download_asset_json
...
+endif()
+if (NOT TARGET json-proprietary-target)
+download_asset_release_json(json-rt1051-target
+                            ${CMAKE_CURRENT_SOURCE_DIR}/assets/assets_rt1051.json
+                            ${CMAKE_BINARY_DIR}/sysroot/sys/current/
+                            MuditaOSPublicAssets
+                            0.0.5
+                            ${MUDITA_CACHE_DIR}
+                            )
+endif()
```
swichu91, line 105:
```suggestion
if (NOT TARGET json-proprietary-target)
    download_asset_json(json-proprietary-target
                        ${CMAKE_CURRENT_SOURCE_DIR}/assets/assets_proprietary.json
                        ${CMAKE_BINARY_DIR}/sysroot/sys/current/
                        MuditaOSAssets
                        ${MUDITA_CACHE_DIR}
                        )
download_asset_release_json(json-common-target
                            ${CMAKE_CURRENT_SOURCE_DIR}/assets/assets_common.json
                            ${CMAKE_BINARY_DIR}/sysroot/sys/current/
                            MuditaOSPublicAssets
                            0.0.5
                            ${MUDITA_CACHE_DIR}
                            )
download_asset_release_json(json-community-target
                                ${CMAKE_CURRENT_SOURCE_DIR}/assets/assets_community.json
                                ${CMAKE_BINARY_DIR}/sysroot/sys/current/
                                MuditaOSPublicAssets
                                0.0.5
                                ${MUDITA_CACHE_DIR}
                                )
download_asset_release_json(json-rt1051-target
                            ${CMAKE_CURRENT_SOURCE_DIR}/assets/assets_rt1051.json
                            ${CMAKE_BINARY_DIR}/sysroot/sys/current/
                            MuditaOSPublicAssets
                            0.0.5
                            ${MUDITA_CACHE_DIR}
                            )
endif()
```

module-gui/gui/core/FontManager.cpp
```diff
@@ -33,29 +53,31 @@ namespace gui
...
         if (file == nullptr) {
+            LOG_ERROR("Failed to open file: %s", path.c_str());
             return nullptr;
         }
 
-        auto fileSize = std::filesystem::file_size(filename);
+        auto fileSize = std::filesystem::file_size(path);
         std::rewind(file);
 
         char *fontData = new char[fileSize];
```
swichu91, line 77: Replace it with `std::unique_ptr<char[]> p(new char[fileSize])` or use plain std::vector. Then you won't need to manually manage memory.

swichu91 (commented):
Please fix some minor issues, apart from that it looks good to me.

module-bsp/board/linux/hal/battery_charger/BatteryCharger.cpp
```diff
@@ -44,7 +44,7 @@ namespace hal::battery
 
         xQueueHandle notificationChannel = nullptr;
         TaskHandle_t batteryWorkerHandle = nullptr;
-        unsigned batteryLevel            = 100;
+        unsigned batteryLevel            = 10;
```
swichu91, line 47: Should it be changed?

paweljonskim: No! thanks for catching it

module-gui/gui/core/FontManager.cpp
```diff
@@ -33,57 +54,54 @@ namespace gui
...
-            delete[] fontData;
             return nullptr;
         }
 
         // allocate memory for new font
-        RawFont *font = new RawFont();
-        if (font->load(reinterpret_cast<uint8_t *>(fontData)) != gui::Status::GUI_SUCCESS) {
-            delete font;
-            delete[] fontData;
+        RawFont *rawfont = new RawFont();
```
swichu91, line 93: What if `RawFont` returned nullptr?

module-gui/gui/core/FontManager.cpp
```diff
@@ -96,28 +114,61 @@ namespace gui
         }
     }
 
-    std::vector<std::string> FontManager::getFontsList()
+    std::map<Fonts, std::string> FontManager::getFontsList()
     {
+        auto fd = std::fopen(fontMapFile.c_str(), "r");
```
swichu91, line 119: File descriptor ought to be closed if not used anymore

module-gui/gui/core/FontManager.cpp
```diff
@@ -96,28 +114,61 @@ namespace gui
...
+        if (!err.empty()) {
+            LOG_ERROR("Failed parsing device string!");
+            throw std::invalid_argument("Can't parse the file!");
+        }
 
-        for (const auto &entry : std::filesystem::directory_iterator(fontFolder)) {
-            if (!std::filesystem::is_directory(entry) && entry.path().extension() == ".mpf") {
-                fontFiles.push_back(entry.path().string());
+        auto fontmapObjects = fontmapJson.object_items();
+        auto infoJson       = fontmapObjects["info"];
```
swichu91, line 139:
```suggestion
        const auto fontmapObjects = fontmapJson.object_items();
        const auto infoJson       = fontmapObjects["info"];
```

module-gui/gui/core/FontManager.cpp
```diff
@@ -96,28 +114,61 @@ namespace gui
...
+        defaultFontName     = infoJson["default_font_name"].string_value();
+        defaultFontTypeName = infoJson["default_font_type_name"].string_value();
+
+        auto styleJson = fontmapObjects["style"];
```
swichu91, line 143:
```suggestion
        const auto styleJson = fontmapObjects["style"];
```

module-gui/gui/core/FontManager.cpp
```diff
@@ -161,6 +212,28 @@ namespace gui
         return fonts[num];
     }
 
+    [[nodiscard]] auto FontManager::getFont(Fonts fontType) const -> RawFont *
+    {
+        auto font = find(fonts_map::map[fontType]);
+        // default return first font
+        if (font == nullptr && fonts.size() > 0) {
```
swichu91, line 219:
```suggestion
        if (font == nullptr && not fonts.empty()) {
```

module-gui/gui/core/FontManager.cpp
```diff
@@ -33,57 +54,54 @@ namespace gui
...
-        auto file = std::fopen(filename.c_str(), "rb");
+        auto file = std::fopen(path.c_str(), "rb");
         if (file == nullptr) {
+            LOG_ERROR("Failed to open file: %s", path.c_str());
             return nullptr;
         }
 
-        auto fileSize = std::filesystem::file_size(filename);
+        auto fileSize = std::filesystem::file_size(path);
         std::rewind(file);
```
swichu91, line 76: There is no need for rewinding.

module-gui/gui/core/FontManager.cpp
```diff
@@ -33,57 +54,54 @@ namespace gui
...
-        if (fontData == nullptr) {
-            std::fclose(file);
-            LOG_ERROR(" Failed to allocate temporary font buffer");
-            return nullptr;
-        }
+        std::vector<uint8_t> fontData(fileSize, 0);
 
         // read data to buffer
-        auto bytesRead = std::fread(fontData, 1, fileSize, file);
+        std::ifstream input(path, std::ios::in | std::ifstream::binary);
```
swichu91, line 81: Return value check missing.

module-gui/gui/core/FontManager.cpp
```diff
@@ -33,57 +54,54 @@ namespace gui
...
 
-        auto file = std::fopen(filename.c_str(), "rb");
+        auto file = std::fopen(path.c_str(), "rb");
         if (file == nullptr) {
+            LOG_ERROR("Failed to open file: %s", path.c_str());
             return nullptr;
         }
 
-        auto fileSize = std::filesystem::file_size(filename);
+        auto fileSize = std::filesystem::file_size(path);
```
swichu91, line 75:
```suggestion
        const auto fileSize = std::filesystem::file_size(path);
```

module-gui/gui/core/FontManager.cpp
```diff
@@ -33,57 +54,54 @@ namespace gui
...
-            LOG_ERROR(" Failed to allocate temporary font buffer");
-            return nullptr;
-        }
+        std::vector<uint8_t> fontData(fileSize, 0);
 
         // read data to buffer
-        auto bytesRead = std::fread(fontData, 1, fileSize, file);
+        std::ifstream input(path, std::ios::in | std::ifstream::binary);
+        input.read(reinterpret_cast<char *>(&fontData[0]), fontData.size());
+        auto bytesRead = input.gcount();
```
swichu91, line 83:
```suggestion
        const auto bytesRead = input.gcount();
```

module-gui/gui/widgets/NavBar.cpp
```diff
@@ -43,7 +43,7 @@ namespace gui::nav_bar
         left->setAlignment(Alignment(Alignment::Horizontal::Left, gui::Alignment::Vertical::Bottom));
         left->setPadding({0, 0, 0, style::nav_bar::bottom_padding});
         left->setMinimumHeight(widgetArea.h);
-        left->setFont(style::nav_bar::font::medium);
+        left->setFont(Fonts::Huge);
```
mjan-dev, line 46: navbar medium should be huge?

products/BellHybrid/apps/application-bell-onboarding/windows/OnBoardingOnOffWindow.cpp
```diff
@@ -28,7 +28,7 @@ namespace style::onboarding
     namespace text
     {
         constexpr auto marginLeft   = 15;
-        constexpr auto minimumSizeX = 81;
+        constexpr auto minimumSizeX = 100;
```
mjan-dev, line 31: Why the change in size?

test/assets/assets_common.json
```diff
@@ -0,0 +1,4 @@
+{
+    "assets": [
```
mjan-dev, line 2: what is the reason for this empty file?

mjan-dev (approved):
Approved with minor comments. Please respond before merging.

configure.sh
```diff
@@ -74,6 +74,22 @@ function check_build_type() {
     esac
 }
 
+function check_assets() {
```
pholat, line 77: not needed, please remove - it's not like we add all options to configure & catalog directory.
Also - there is no need to - reconfiguration should be lightweight for this example.

image/CMakeLists.txt
```diff
@@ -9,7 +9,11 @@ add_assets_target(
     SOURCE_DIR ${ASSETS_SOURCE_DIR}
     DEST_DIR ${ASSETS_DEST_DIR}
     DEVEL ${WITH_DEVELOPMENT_FEATURES}
-    DEPENDS json-target json-images-target
+    DEPENDS
+        json-common-target
+        "$<$<STREQUAL:${ASSETS},Proprietary>:json-proprietary-target>"
```
pholat, line 14: please rename to `ASSETS_TYPE` for example, just `ASSETS` seems to ambiguous

image/CMakeLists.txt
```diff
@@ -9,7 +9,11 @@ add_assets_target(
...
+        "$<$<STREQUAL:${ASSETS},Community>:json-community-target>"
+        "$<$<STREQUAL:${PROJECT_TARGET},TARGET_RT1051>:json-rt1051-target>"
```
pholat, line 16: why do we need a flag for RT1051 nowhere? 
@swichu91 won't it be better to just include rt1051 in `only rt1051` cmakes?

module-apps/apps-common/popups/BluetoothAuthenticatePopup.cpp
```diff
@@ -81,7 +81,7 @@ namespace gui
         inputText->drawUnderline(false);
         inputText->setTextType(TextType::SingleLine);
         inputText->setTextEllipsisType(TextEllipsis::Both);
-        inputText->setFont(style::window::font::largelight);
+        inputText->setFont(Fonts::LargeLight);
```
pholat, line 84: all these renames actually add nothing. 
why can't you just change past constexprs:
`style::window::font ....` to the enum you've added?

It will result in much less code change, with literally no difference in these lines. And mass renames result only in less readable PR :(

pholat (changes requested):
First review:
1. please remove changes in configure.sh - these are not needed
2. please add the required documentation for this new feature
3. please revert not needed renames - these pollute the PR and make it unreadable. If you wish to make a mass rename at least add it in separate PR with just that.

module-gui/gui/core/FontManager.cpp
```diff
@@ -7,16 +7,37 @@
...
 #include <cstdio>
 
-namespace style::window::font
-{
-    inline constexpr auto default_fallback_font = "dejavu_sans_bold_27";
-}
-
 namespace gui
 {
+    namespace fonts_map
```
pholat, line 17: so enum was not needed in the first place? :)

products/BellHybrid/apps/application-bell-onboarding/windows/OnBoardingOnOffWindow.cpp
paweljonskim: I changed that so community fonts would fit. Reverting since we dont wanna make sure they do

module-utils/log/api/log/debug.hpp
```diff
@@ -19,4 +19,4 @@
 #define DEBUG_TIMER                  0 /// debug timers system utility
 #define DEBUG_SETTINGS_DB            0 /// show extensive settings logs for all applications
 #define DEBUG_SERVICE_CELLULAR       0 /// show various logs in cellular service
-#define DEBUG_MISSING_ASSETS         1 /// show debug concerning missing assets
+#define DEBUG_MISSING_ASSETS         0 /// show debug concerning missing assets
```
pholat, line 22: :bow:

image/CMakeLists.txt
paweljonskim: To exclude LUTS.bin from linux builds

module-gui/gui/core/FontManager.cpp
paweljonskim: handled

module-gui/gui/core/FontManager.cpp
paweljonskim: added fclose  few lines below

module-utils/log/api/log/debug.hpp
paweljonskim: does this mean an approval?

module-gui/gui/widgets/NavBar.cpp
paweljonskim: reverted, fixed

test/assets/assets_common.json
paweljonskim: I autocleaned that via some sed command so didnt notice its empty already.
removed

configure.sh
paweljonskim: removed all changes on configure.sh layer

image/CMakeLists.txt
paweljonskim: renamed

module-apps/apps-common/popups/BluetoothAuthenticatePopup.cpp
paweljonskim: reverted all changes to setFont-like methods
also taking in mind:
https://github.com/mudita/MuditaOS/pull/3649#discussion_r829848895
removing Fonts enum and replacing it with string to string mapping (typename to filename ie "medium" -> "gt_pressura_99")

module-gui/gui/core/FontManager.cpp
paweljonskim: removed

CMakeLists.txt
```diff
@@ -28,10 +28,15 @@ include(DiskImage)
 include(AddPackage)
 include(AutoModuleOption)
 
+if(NOT ASSETS_TYPE)
```
pholat, line 31: assets should be an `option` shouldn't these? we actually have quite a few options, please see here:
doc/ProjectConfig.md

image/CMakeLists.txt
pholat: please see that other platform depended things have things like that done by including platform catalog. 

Now - these `JSON` targets are different per option. 

So instead making wonky ifs on strings in here. Why not define `json` target which will be defined based on options combinations to either:
json = json-common + proprietary + rt1051
json = json-common + oss + rt1051
json = json-common + oss
?

module-utils/log/api/log/debug.hpp
pholat: yest it is :)

module-gui/test/test-catch-text/CMakeLists.txt
```diff
@@ -6,17 +6,17 @@ add_catch2_executable(
...
                 test-gui-TextBlock.cpp
                 test-gui-TextBlockCursor.cpp
                 test-gui-TextDocument.cpp
                 test-gui-MultiTextLine.cpp
                 test-gui-TextParse.cpp
-                test-gui-Font.cpp
-                test-gui-TextLinesCursor.cpp
-		        test-gui-TextInLineCursor.cpp
-		INCLUDE
+                "$<$<STREQUAL:${ASSETS},Proprietary>:test-gui-Text.cpp>"
```
pholat, line 14: I believe you missed `ASSETS_TYPE`.

This is a place where you should rather just move files to ASSET_TYPE dependent catalog and just source them from there. Then you will have one include of each - each ASSET_TYPE dependent one.

pholat (changes requested):
Still seems like needs a bit of work

image/CMakeLists.txt
paweljonskim: well isnt like we are doing this already by defining target `assets`? All it does is to combine different jsons (like you mentioned into one target. How it would be different from the `assets` target?

pholat: actually `add_assets_target(` calls our cmake function which does some magic, which creates targets and does some more magic. The easiest explanation I can find is:

In my opinion, it would be better to have the approach:
`if -> call to function ` 
I see it as the classic approach and it would mean to include target-dependent things, not add string comparisons in the function call.

Right now you are doing:
```
foo ( if ( something == propritetary) add_something (else if something == community) add_something_else && if (project_target == rt1051 ) add_rt1051)
```
where I think would make a bit more sense to:
```
list(actuall_dependencies)
# somewhere: ...

# 1: have a safe if's
if ( something == proprietary ) list(acctuall_dependencies append propritetary-json)
if ( something ==community) list(actuall_dependencies append community-json)

# 2: actually split PROJECT_TARGET dependent elements
if ( something != proprietary && something != community ) message(ERROR "This is fail safe: wrong $ASSETS_TARGET only proprietary and community available)
include ( ${PROJECT_TARGET}/assets)

# ... 

# then pass it to:
add_assets_target( .... DEPENDS actuall_dependencies)
```

In the end: yes. It's only about organizing the code here.  At best I would see: the removal hack of development assets too as a target, not wonky function call in `add_assets_target` and I know we have to live with what we have. I would just love if we could write it a bit more readable at least :) 

If you disagree. Then it's fine with me. I'm no expert in cmake and I do not feel this is the most important part of my review anyway :)

pholat:
What still needs some work (update from: https://github.com/mudita/MuditaOS/pull/3649#pullrequestreview-914131602 ) 

- [ ] Write documentation for this feature
- [ ] double check ASSETS_TARGET ( I believe in some cases it was not renamed properly and it shouldn't work) 
- [ ] add at least in one place check if ASSETS_TARGET has a proper string value ( I still think that set of OPTIONS would be better, but it doesn't matter)

What's **not required** for me to unblock but **would be highly appreciated**:
- [ ] Try to write cmakes in a fail-fast approach.

Unfortunately, cmake is one of these wonky languages where all variables are just strings and no variable is just an empty string on its own. This promotes situations where no variable - which normally we would treat as an error, would be treated as valid textual input.

doc/fonts.md
```diff
@@ -0,0 +1,51 @@
...
+        "big":                  "dejavu_sans/dejavu_sans_regular_30.mpf",
+        "biglight":             "dejavu_sans/dejavu_sans_light_30.mpf",
+        "mediumbold":           "dejavu_sans/dejavu_sans_bold_27.mpf",
+        "medium":               "dejavu_sans/dejavu_sans_regular_27.mpf",
+        "mediumlight":          "dejavu_sans/dejavu_sans_light_27.mpf",
+        "smallbold":            "dejavu_sans/dejavu_sans_bold_24.mpf",
+        "small":                "dejavu_sans/dejavu_sans_regular_24.mpf",
+        "verysmallbold":        "dejavu_sans/dejavu_sans_bold_20.mpf",
+        "verysmall":            "dejavu_sans/dejavu_sans_regular_20.mpf"
+    }
```
pholat, line 36: consider just linking to existing example in code :bow: and just pasting here one or two exemplary lines

doc/fonts.md
```diff
@@ -0,0 +1,51 @@
...
+There are two values in `info` part:
+- `default_font_name` - fallback font in case of inability to load specified font
+- `default_font_type_name` - default font family name used in methods which specify only size and weight of a font
+
+
+Every key in `style` part defines font type to be used in fonts list in MuditaOS, which later can be used in `FontManager::getFont` method (ie `getFont("supersizemelight")`)
+
+## Fonts usage in MuditaOS
+
+`FontManager` loads all fonts basing on `fontmap.json`. They are later accessible in methods (listing only suggested ones, for full API please see `FontManager.hpp`):
```
pholat, line 48:
```
FontManager loads all fonts based on `fontmap.json`. These can be later accessed in GUI methods, specifically via: 
```

doc/fonts.md
```diff
@@ -0,0 +1,51 @@
...
+}
+```
+There are two values in `info` part:
+- `default_font_name` - fallback font in case of inability to load specified font
+- `default_font_type_name` - default font family name used in methods which specify only size and weight of a font
+
+
+Every key in `style` part defines font type to be used in fonts list in MuditaOS, which later can be used in `FontManager::getFont` method (ie `getFont("supersizemelight")`)
+
+## Fonts usage in MuditaOS
```
pholat, line 46: Consider adding some more info that:

- MuditaOS uses its own binary format
- That means that all fonts to be used in format have to be first converted to it
- Would be nice to link to font converter (which is an opensource project fork adapted to export our fonts @mjan-dev can we open it ?) 

This is done to:
1. lessen the workload on GPUless processor
2. lessen the size required for fonts to be stored on disc - and loaded on RAM

doc/i18n.md
```diff
@@ -100,6 +100,8 @@ On Mudita Pure phone we use:
 - default font (`gt_pressura` font family)
 - default fallback font (DeJavu Sans Bold size 27 **Emojis only**)
 
+On Mudita Harmony we use only DeJavu Sans font family
```
pholat, line 103: 1. There is no GT pressura on harmony? :o 
2.  i18n is just a product-agnostic library - why add it here? add it to the product specification if required ( Pure Phone features are linked from main readme )

doc/quickstart.md
```diff
@@ -100,15 +100,19 @@ We are using J-Link driver in version J-Link v634f ([Ubuntu download](https://ww
 
 **WARNING:** newer J-Link software revisions seem to work fine, but loading via them does not result in usable software on the hardware target.
 
-#### Add GitHub token
+#### Proprietary assets
 
 We are storing assets on github pages and these are downloaded automatically before software images are built.
-Please follow github token configuration here: [download assets documentation](../doc/download_asset.py)
+There are two assets sources - proprietary and open for community.
+
```
pholat, line 107: users still have to configure the token - not configuration of the token is not mentioned explicitly. 

Can you ignore adding the token when you download nonproprietary assets? No, you can not. Therefore it has to be mentioned here.

module-gui/gui/core/FontManager.cpp
```diff
@@ -96,28 +99,60 @@ namespace gui
         }
     }
 
-    std::vector<std::string> FontManager::getFontsList()
+    std::map<std::string, std::string> FontManager::getFontsList()
```
pholat, line 102: as we change Font Manager - can we add some UT here? i.e. if all loaded are loaded, what happens when json states that there is font but there is no such font on disc etc
i.e. here: module-gui/test/test-catch-text/test-gui-Font.cpp

module-gui/gui/core/FontManager.cpp
```diff
@@ -161,6 +196,39 @@ namespace gui
...
+    {
+        try {
+            auto font = find(fontMap.at(fontType));
+            return font;
+        }
+        catch (const std::out_of_range &e) {
+            if (not fonts.empty()) {
+#if DEBUG_MISSING_ASSETS == 1
+                LOG_ERROR("=> font not found: %s using default", name.data());
+#endif
```
pholat, line 209: why not: https://en.cppreference.com/w/cpp/container/map/find ?

pholat (approved):
:+1: Great work :D we accidentally killed a few birds with one stone here.
- we finally have open assets for the OSS community :bow: 
but as a plus:
- we can simply change fonts for different styles & fonts for tests with simple config ( there is actually task for it on roadmap )

doc/fonts.md
mjan-dev: @pholat I believe, we can open it. Lets talk about opening this one and updater at once.

PR #3650, opened 2022-03-16
[BH-1350] Meditation timer update

swichu91:
Meditation timer application update.
Refactored gui spinners, list items.
Fixed minor issues in snooze alarm settings.


module-apps/apps-common/widgets/spinners/Container.hpp
```diff
@@ -0,0 +1,226 @@
+// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
+// For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
+
+#pragma once
+
+#include <vector>
+#include <string>
+#include <type_traits>
+
+/// Generic elements container.
```
mjan-dev, line 10: :+1:

products/BellHybrid/apps/application-bell-meditation-timer/models/ChimeVolume.cpp
```diff
@@ -0,0 +1,15 @@
...
+#include "ChimeVolume.hpp"
+
+namespace app::meditation::models
+{
+
+    void ChimeVolume::setValue(std::uint32_t value)
+    {}
+    std::uint32_t ChimeVolume::getValue() const
+    {
+        return 5;
```
mjan-dev, line 13: should it return hard coded value?

swichu91: I added only a dummy implementation just to be able to compile. This should be properly implemented as part of BH-1433.

products/BellHybrid/apps/application-bell-meditation-timer/presenter/ReadyGoingPresenter.hpp
```diff
@@ -24,16 +29,19 @@ namespace app::meditation
...
     class ReadyGoingPresenter : public ReadyGoingPresenterContract::Presenter
     {
         app::ApplicationCommon *app{};
+        models::StartDelay &startDelayModel;
         void activate() override;
+        std::chrono::seconds getStartDelay() override;
 
       public:
-        explicit ReadyGoingPresenter(app::ApplicationCommon *app);
+        ReadyGoingPresenter(app::ApplicationCommon *app, models::StartDelay &startDelayModel);
```
mjan-dev, line 45:
```suggestion
        ReadyGoingPresenter(app::ApplicationCommon *app, const models::StartDelay &startDelayModel);
``` ?

File: products/BellHybrid/apps/application-bell-meditation-timer/presenter/SettingsPresenter.cpp [hunk 4]
```diff
@@ -0,0 +1,107 @@
...
+        }
+    };
+
+} // namespace app::list_items
+
+namespace app::meditation
+{
+    using namespace gui;
+    SettingsPresenter::SettingsPresenter(app::ApplicationCommon *app,
+                                         models::ChimeInterval &chimeIntervalModel,
```
mjan-dev, line 59: minor: all references probably could be const

products/BellHybrid/apps/common/src/LanguageUtils.cpp
```diff
@@ -3,21 +3,43 @@
 
 #include <i18n/i18n.hpp>
 
-namespace utils::language
+namespace
 {
-    auto getCorrectMinutesNumeralForm(int val) -> std::string
+    std::string transform(const std::uint32_t val,
+                          const std::string &minute_lower,
```
mjan-dev, line 9: Why use underscore style if camelCase is commonly used?

mjan-dev (approved):
Approved with minor comments.

products/BellHybrid/apps/application-bell-meditation-timer/models/ChimeVolume.cpp
mjan-dev: Please add comment then.

products/BellHybrid/apps/application-bell-meditation-timer/presenter/ReadyGoingPresenter.hpp
swichu91: Fixed.

products/BellHybrid/apps/application-bell-meditation-timer/presenter/SettingsPresenter.cpp
swichu91: It won't work as each provider's item needs to have access to both `setValue` and `getValue` :(

products/BellHybrid/apps/common/src/LanguageUtils.cpp
swichu91: Fixed

image/assets/lang/Deutsch.json
```diff
@@ -80,6 +80,10 @@
   "common_minutes_lower": "minuten",
   "common_minutes_lower_genitive": "minuten",
   "common_minute_short": "min",
+  "common_second_lower": "TODO",
+  "common_seconds_lower": "TODO",
+  "common_seconds_lower_genitive": "TODO",
+  "common_second_short": "TODO",
```
pholat, line 86: just do not add translations for languages that are not used. Default English will be taken in their place. If there is none for English - ten i18n keys will be taken i.e. `common_second_lower` 
This is how it works by design :)

module-apps/apps-common/widgets/spinners/Container.hpp
```diff
@@ -0,0 +1,226 @@
...
+    {}
+
+    ElementType get() const
+    {
+        try {
+            return elements.at(pos);
+        }
+        catch (...) {
+            return {};
+        }
```
pholat, line 41: please don't leave catch-all directives :)

module-apps/apps-common/widgets/spinners/Container.hpp
pholat, line 10: just nitpicking but isn't it more like a model than a container? :) It wraps the generic stl container and provides user-specific stored data to the view, doesn't it?

In the listview, we call such an item provider. Would you consider taking a quick peek there too? just for reference.

module-apps/apps-common/widgets/spinners/Container.hpp
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+  private:
+    [[nodiscard]] std::uint32_t range_top() const
+    {
+        return elements.size() - 1;
+    }
+
+    range elements;
+    std::uint32_t pos{};
```
pholat, line 120: We work on stl containers here - why not iterator? 
Also... is this logic really required - can't we have derived "rotary" or not "rotary" which would just change `next` logic? for rotary you could use `std::rotate` most probably :)

products/BellHybrid/apps/application-bell-meditation-timer/windows/StatisticsWindow.cpp
```diff
@@ -0,0 +1,70 @@
...
+        setFocusItem(sideListView);
+    }
+
+    bool StatisticsWindow::onInput(const gui::InputEvent &inputEvent)
+    {
+        if (sideListView->onInput(inputEvent)) {
+            return true;
+        }
+        if (inputEvent.isShortRelease(KeyCode::KEY_ENTER)) {
+            exit();
```
pholat, line 51: presenter->handleEnter() ? then whole exit may be in presenter? :)

products/BellHybrid/apps/application-bell-settings/models/alarm_settings/AlarmSettingsListItemProvider.cpp
```diff
@@ -4,63 +4,66 @@
...
 
         constexpr auto volumeStep = 1U;
         constexpr auto volumeMin  = 1U;
         constexpr auto volumeMax  = 10U;
-        auto alarmVolume          = new NumListItem(model.getAlarmVolume(),
-                                           UIntegerSpinner::Range{volumeMin, volumeMax, volumeStep},
-                                           utils::translate("app_bell_settings_alarm_settings_volume"));
-        alarmVolume->setOnValueChanged([this](const UIntegerSpinner::Type &val) {
+        auto alarmVolume =
+            new list_items::Number(list_items::Number::spinner_type::range{volumeMin, volumeMax, volumeStep},
```
pholat, line 51: this uppercase `Number` seems weird now (just saying)

products/BellHybrid/apps/application-bell-settings/models/alarm_settings/AlarmSettingsListItemProvider.hpp
```diff
@@ -11,10 +11,10 @@ namespace app::bell_settings
     class AlarmSettingsListItemProvider : public SettingsListItemProvider
     {
       public:
-        AlarmSettingsListItemProvider(AbstractAlarmSettingsModel &model, std::vector<UTF8> alarmToneRange);
+        AlarmSettingsListItemProvider(AbstractAlarmSettingsModel &model, std::vector<std::string> alarmToneRange);
```
pholat, line 14: afair I understand this is forced by check for stl type in template now. But utf8 is first class citizen in the os - can't we just add this case to the template instead of changing everywhere `UTF8 -> std::string` ?

pholat (commented):
- [ ] It would be nice to not rename utf8 to string everywhere, just accept utf8 in the template too
- [ ] I would suggest changing `Container to Model` name :) 
- [ ] Where are tests for `Container` :D 

some other comments ;)

image/assets/lang/Deutsch.json
swichu91: Fixed

module-apps/apps-common/widgets/spinners/Container.hpp
swichu91: Fixed

products/BellHybrid/apps/application-bell-settings/models/alarm_settings/AlarmSettingsListItemProvider.hpp
swichu91: True, fixed.

products/BellHybrid/apps/application-bell-meditation-timer/windows/StatisticsWindow.cpp
swichu91: Good point, fixed!

module-apps/apps-common/widgets/spinners/Container.hpp
swichu91: Of course we can and it would be better than the current solution as we don't need to change this behavior at runtime. Unfortunately, there is a lot of code that is using such an approach already. As you might already know this would trigger more and more changes. IMHO this is a good proposal. Why not create a jira task for it?

module-apps/apps-common/widgets/spinners/Container.hpp
swichu91: Well, actually it's not limited to the GUI. It can contain anything you want. Its main purpose is to provide specific iteration model (fixed/rotating).

pholat: it's in widgets and used only for widgets built on top of actually generic (stl) containers using integers instead iterators. If you wish for it's generic use - then use to utils as it's in the wrong place then?
Although in my opinion it would be a bit faking what it's not :) Saying that - I do not care much. Do whatever you want here.

swichu91: Fixed :)

PR #3651, opened 2022-03-16
[MOS-124] Add Code Comparision pairing logic

SP2FET:
Added missing logic for code comparison pairing, fixed popup text


module-bluetooth/Bluetooth/interface/profiles/GAP/GAP.cpp
```diff
@@ -243,8 +243,20 @@ namespace bluetooth
         case HCI_EVENT_USER_CONFIRMATION_REQUEST: {
             bd_addr_t addr;
             hci_event_user_confirmation_request_get_bd_addr(packet, addr);
-            hci_connection_t *conn = hci_connection_for_bd_addr_and_type(addr, BD_ADDR_TYPE_ACL);
-            hci_send_cmd(&hci_user_confirmation_request_reply, &conn->address);
+
+            auto code = hci_event_user_passkey_notification_get_numeric_value(packet);
+            auto it   = devices().find(addr);
+            if (it == devices().end()) {
+                gap_remote_name_request(addr, PAGE_SCAN_MODE_STANDARD, 0);
```
mjan-dev, line 250: Just a noob bt question. Can this method not get the name of device with given addr? What then?

mjan-dev, line 250: Just a noob bt question. Can this method not get the name of device with given addr? What then?

SP2FET: yes -> by default we get the device by MAC address. During the scan, we can get the name or request for it, but in this case, we don't start scanning, thus if we want to get the human-readable name we have to request for it. There's a race condition though - name can came after the popup is displayed, but that's a minor issue as it does not interfere with functionality

SP2FET: I've made something similar in scan list - if we don't get the name, we display the BT MAC address

mjan-dev (approved):

PR #3656, opened 2022-03-18
[BH-1434] Fix Bedtime Tone Return

tomaszrybarski-digica:
Fix audio playback on return

**Description**

Audio playback did not properly stop on return without save from Bedtime Tone Settings and Alarm Settings. Furthermore a crash due to segmentation fault was observed upon re-entering into those settings. This PR fixes those issues

### DIFF
--- a/products/BellHybrid/apps/application-bell-settings/presenter/BedtimeSettingsPresenter.cpp
+++ b/products/BellHybrid/apps/application-bell-settings/presenter/BedtimeSettingsPresenter.cpp
@@ -22,7 +22,6 @@ namespace app::bell_settings
         this->provider->onExit = [this]() { getView()->exit(); };
 
         this->provider->onToneEnter  = playSound;
-        this->provider->onToneExit   = [this](const auto &) { stopSound(); };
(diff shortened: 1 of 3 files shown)


products/BellHybrid/apps/application-bell-settings/presenter/BedtimeSettingsPresenter.cpp
```diff
@@ -65,6 +64,7 @@ namespace app::bell_settings
     }
     void BedtimeSettingsPresenter::exitWithoutSave()
     {
+        this->stopSound();
```
mjan-dev, line 67: is the sound stopped properly if the changes are saved?

tomaszrybarski-digica: Yes, it is being stopped properly and this has been tested - stopSound() invoked with the onVolumeExit callback guarantees that

pholat (approved):

mjan-dev (approved):

PR #3666, opened 2022-03-22
[BH-1432] Split Language Display Handling

tomaszrybarski-digica:
Separate implementations of handleDisplayLanguageChanged()

**Description**
This PR fixes BellHybrid application assertion during language change. HandleDisplayLanguageChanged() needs to have different behaviour now between Bell and Pure

### DIFF
--- a/products/BellHybrid/services/appmgr/ApplicationManager.cpp
+++ b/products/BellHybrid/services/appmgr/ApplicationManager.cpp
@@ -9,6 +9,7 @@
 #include <application-bell-main/ApplicationBellMain.hpp>
 #include <application-bell-onboarding/BellOnBoardingNames.hpp>
 #include <service-appmgr/Constants.hpp>
+#include <service-db/agents/settings/SystemSettings.hpp>
 
 namespace app::manager
 {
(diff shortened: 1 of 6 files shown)


products/BellHybrid/services/appmgr/ApplicationManager.cpp
```diff
@@ -45,6 +46,20 @@ namespace app::manager
         }
     }
 
+    auto ApplicationManager::handleDisplayLanguageChange(app::manager::DisplayLanguageChangeRequest *msg) -> bool
```
mjan-dev, line 49: It should probably be in the common part as this implementation is shared between Pure and Bell. 
The difference should be in the Pure implementation:
DBServiceAPI::InformLanguageChanged(this);

mjan-dev (changes requested):

products/BellHybrid/services/appmgr/ApplicationManager.cpp
tomaszrybarski-digica: So the idea behind this PR is such:
-`handleDisplayLanguageChange() ` is now virtual in ApplicationManagerCommon
ApplicationManagers for Pure and Bell have this method overriden:
- Pure uses the one from ApplicationManagerCommon without changes
- Bell has this method slightly modified with DBServiceAPI::InformLanguageChanged(this) removed. This was causing an assertion during language change and as i understand is not needed in Bell

tomaszrybarski-digica: Applied changes suggested in the comment

paweljonskim (approved):

PR #3681, opened 2022-03-24
[MOS-240] PVS Studio analysis bugs fix

Lefucjusz:
Fixes to major part of issues found
by PVS Studio static analysis
mainly micro-optimisations.

### DIFF
--- a/module-apps/application-calendar/include/application-calendar/data/CalendarData.hpp
+++ b/module-apps/application-calendar/include/application-calendar/data/CalendarData.hpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2021, Mudita Sp. z.o.o. All rights reserved.
+// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
 // For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 #pragma once
@@ -46,7 +46,7 @@ class DayMonthData : public gui::SwitchData
         return dateFilter;
     };
 
-    virtual void setData(std::string dayMonthText, const TimePoint &dateNumb)
+    virtual void setData(const std::string &dayMonthText, const TimePoint &dateNumb)
     {
         dayMonth   = dayMonthText;
         dateFilter = dateNumb;
--- a/module-apps/application-settings/windows/OptionsWidgetMaker.cpp
+++ b/module-apps/application-settings/windows/OptionsWidgetMaker.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2021, Mudita Sp. z.o.o. All rights reserved.
+// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
 // For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 #include "OptionsWidgetMaker.hpp"
@@ -10,11 +10,14 @@ namespace gui
     OptionsWidgetMaker::OptionsWidgetMaker(OptionWindow *window) : mWindow(window)
     {}
 
-    void OptionsWidgetMaker::addSwitchOption(
-        std::list<gui::Option> &list, const UTF8 &text, bool &option, std::function<void()> callback, bool indent)
+    void OptionsWidgetMaker::addSwitchOption(std::list<gui::Option> &list,
+                                             const UTF8 &text,
+                                             bool &option,
+                                             const std::function<void()> &callback,
+                                             bool indent)
     {
         list.emplace_back(std::make_unique<option::OptionSettings>(
-            std::move(text),
+            text,
             [=](Item &item) {
                 callback();
                 return true;
@@ -35,11 +38,11 @@ namespace gui
 
     void OptionsWidgetMaker::addSelectOption(std::list<gui::Option> &list,
                                              const UTF8 &text,
-                                             std::function<void(void)> callback,
+                                             const std::function<void(void)> &callback,
                                              bool indent)
     {
         list.emplace_back(std::make_unique<option::OptionSettings>(
-            std::move(text),
+            text,
             [=](Item &item) {
                 callback();
                 return true;
--- a/module-apps/application-settings/windows/advanced/ColorTestWindow.cpp
+++ b/module-apps/application-settings/windows/advanced/ColorTestWindow.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2021, Mudita Sp. z.o.o. All rights reserved.
+// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
 // For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 #include "ColorTestWindow.hpp"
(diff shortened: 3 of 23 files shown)


module-apps/application-calendar/include/application-calendar/data/CalendarData.hpp
```diff
@@ -46,7 +46,7 @@ class DayMonthData : public gui::SwitchData
         return dateFilter;
     };
 
-    virtual void setData(std::string dayMonthText, const TimePoint &dateNumb)
+    virtual void setData(const std::string &dayMonthText, const TimePoint &dateNumb)
```
SP2FET, line 49: I think it's worth checking if the string being passed here is not copied n-times before it reaches this function :P

module-apps/application-settings/windows/advanced/ColorTestWindow.cpp
```diff
@@ -76,7 +76,7 @@ namespace gui
         if (scheme != currentColorScheme) {
             currentColorScheme = scheme;
             application->bus.sendUnicastSync(
-                std::make_shared<service::gui::ChangeColorScheme>(std::move(scheme)), service::name::gui, 100);
+                std::make_shared<service::gui::ChangeColorScheme>(scheme), service::name::gui, 100);
```
SP2FET, line 79: why the scheme is not moved here? It has already been copied to `currentColorScheme` so it shouldn't be an issue

module-audio/Audio/equalizer/Equalizer.cpp
```diff
@@ -21,6 +21,9 @@ namespace audio::equalizer
         if ((Q < qMinValue || Q > qMaxValue) && filter != FilterType::None) {
             throw std::invalid_argument("Q out of range");
         }
+        if (samplerate == 0 && filter != FilterType::None) {
+            throw std::invalid_argument("Sample rate has to be greater than zero");
+        }
```
SP2FET, line 26: what with the case, when the filter is set to none? :trollface:

module-services/service-audio/ServiceAudio.cpp
```diff
@@ -717,6 +716,7 @@ std::string ServiceAudio::getSetting(const Setting &setting,
             targetPlayback = PlaybackType::CallRingtone;
         }
         else {
+            const std::string defaultValue = {};
             return defaultValue;
```
SP2FET, line 720: this one was for clarity - it's easier to read then

products/BellHybrid/BellHybridMain.cpp
```diff
@@ -50,8 +50,7 @@ int main()
 {
     constexpr auto ApplicationName = "BellHybrid";
 
-    const std::vector<std::string> fileIndexerAudioPaths = {
-        {purefs::dir::getCurrentOSPath() / "assets/audio/bg_sounds"}};
+    std::vector<std::string> fileIndexerAudioPaths = {{purefs::dir::getCurrentOSPath() / "assets/audio/bg_sounds"}};
```
SP2FET, line 53: I don't get why it shouldn't be const

SP2FET (commented):
Please do the smoke tests before merging as this PR touches a lot of different things in the OS

module-apps/application-settings/windows/advanced/ColorTestWindow.cpp
Lefucjusz: scheme is a const variable in this context, so it will be copied despite using std::move, as it cannot be moved

module-audio/Audio/equalizer/Equalizer.cpp
Lefucjusz: Actually this shouldn't be a problem. The division by zero will happen, that's true, so omega value will be incorrect, but when filter is set to none, that value is not used anywhere.

products/BellHybrid/BellHybridMain.cpp
Lefucjusz: It's the same issue as in ColorTestWindow.cpp - this variable is later passed to std::move:

`systemServices.emplace_back(sys::CreatorFor<service::ServiceFileIndexer>(std::move(fileIndexerAudioPaths)));`

If it's const, the std::move will have no effect.

module-services/service-audio/ServiceAudio.cpp
Lefucjusz: Ok, I get it; I'll restore it to the previous version if you prefer, this is such a subtle change that I doubt if it's gonna affect the performance to any measurable degree :P

module-apps/application-settings/windows/advanced/ColorTestWindow.cpp
SP2FET: true :D

module-services/service-audio/ServiceAudio.cpp
SP2FET: I think that this one will look better when reverted - let's stay with one style :P

SP2FET (approved):

module-apps/application-calendar/include/application-calendar/data/CalendarData.hpp
Lefucjusz: It looks that this function is not used anywhere currently ;)

module-services/service-audio/ServiceAudio.cpp
Lefucjusz: Done ;)

module-apps/application-settings/windows/OptionsWidgetMaker.cpp
```diff
@@ -11,7 +11,7 @@ namespace gui
     {}
 
     void OptionsWidgetMaker::addSwitchOption(
-        std::list<gui::Option> &list, const UTF8 &text, bool &option, std::function<void()> callback, bool indent)
+        std::list<gui::Option> &list, UTF8 &&text, bool &option, const std::function<void()> &callback, bool indent)
```
pholat, line 14: as for && can't we just remove std::move on the text below on `text`? and both: && and std::move wont be required :)

module-audio/Audio/equalizer/Equalizer.cpp
pholat: do we catch this division by zero anywhere? ( except in test :joy: )

module-gui/gui/widgets/text/core/cursors/TextBlockCursor.cpp
```diff
@@ -367,7 +367,7 @@ namespace gui
 
     auto BlockCursor::begin() -> std::list<TextBlock>::iterator
     {
-        return document == nullptr ? document->blocks.end() : document->blocks.begin();
+        return document->blocks.begin();
```
pholat, line 370: consider either assert for non existing document or rising an exception

products/BellHybrid/BellHybridMain.cpp
pholat: cant we in this case just remove std::move there?

module-audio/Audio/equalizer/Equalizer.cpp
Lefucjusz: After quick analysis - I don't think so :P None of this exceptions seem to be caught anywhere...

module-gui/gui/widgets/text/core/cursors/TextBlockCursor.cpp
Lefucjusz: Added asserts

module-apps/application-settings/windows/OptionsWidgetMaker.cpp
Lefucjusz: Changed as you suggested ;)

products/BellHybrid/BellHybridMain.cpp
Lefucjusz: Yeah, that seems to be a better solution, as the performance boost from using std::move in this place is rather insignificant. Changed according to your suggestion.

pholat (approved):

PR #3693, opened 2022-03-29
[CP-1080] Collect contact notifications in outbox endpoint

pawel-mudita:
**Description**

When contact record is created/updated/removed during 
connection with Mudita Center, notification of that             
action should be collected in outbox endpoint to
keep contacts data synchronized between MuditaOS
and Mudita Center. In upcoming PRs, SMS and SMSThread
records changes will be handled in a similar way.

Make sure that this PR:

### DIFF
--- a/module-services/service-db/ServiceDBCommon.cpp
+++ b/module-services/service-db/ServiceDBCommon.cpp
@@ -99,8 +99,8 @@ sys::ReturnCodes ServiceDBCommon::SwitchPowerModeHandler(const sys::ServicePower
     return sys::ReturnCodes::Success;
 }
 
-void ServiceDBCommon::sendUpdateNotification(db::Interface::Name interface, db::Query::Type type)
+void ServiceDBCommon::sendUpdateNotification(db::Interface::Name interface, db::Query::Type type, uint32_t recordId)
 {
-    auto notificationMessage = std::make_shared<db::NotificationMessage>(interface, type);
+    auto notificationMessage = std::make_shared<db::NotificationMessage>(interface, type, recordId);
     bus.sendMulticast(notificationMessage, sys::BusChannel::ServiceDBNotifications);
 }
--- a/module-services/service-desktop/ServiceDesktop.cpp
+++ b/module-services/service-desktop/ServiceDesktop.cpp
@@ -4,6 +4,7 @@
 #include "service-evtmgr/BatteryMessages.hpp"
 #include "system/SystemReturnCodes.hpp"
 #include <service-appmgr/messages/DOMRequest.hpp>
+#include <service-db/DBNotificationMessage.hpp>
 #include <service-desktop/DesktopMessages.hpp>
 #include <service-desktop/ServiceDesktop.hpp>
 #include <service-desktop/WorkerDesktop.hpp>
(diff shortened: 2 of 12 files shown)


module-services/service-desktop/ServiceDesktop.cpp
```diff
@@ -34,14 +35,17 @@
...
-          sys::TimerFactory::createSingleShotTimer(this,
-                                                   sdesktop::notificationsClearTimerName,
-                                                   sdesktop::notificationsClearTimerDelayMs,
-                                                   [this](sys::Timer & /*timer*/) { notificationEntries.clear(); })}
+      notificationsClearTimer{sys::TimerFactory::createSingleShotTimer(this,
+                                                                       sdesktop::notificationsClearTimerName,
+                                                                       sdesktop::notificationsClearTimerDelayMs,
+                                                                       [this](sys::Timer & /*timer*/) {
+                                                                           notificationEntries.clear();
+                                                                           notificationCurrentUid = 0;
```
pholat, line 43: this deserves a function. Preferably in NotificationsEntries ( or whatever named - I'm not very good at naming, please follow your intuition) class :)

module-services/service-desktop/ServiceDesktop.cpp
```diff
@@ -241,6 +245,54 @@ sys::ReturnCodes ServiceDesktop::InitHandler()
         return btMsgHandler->handle(msgl);
     });
 
+    connect(typeid(db::NotificationMessage), [&](sys::Message *msg) {
+        if (!notificationsClearTimer.isActive()) {
```
pholat, line 249: please at least create a function, for this code :) 
1. will you add different notifications handling in the service desktop? 
2. will new if's be added here?
3. can we create a class from NotificationsEntries instead of working everywhere on a generic vector?

module-services/service-db/ServiceDBCommon.cpp
```diff
@@ -99,8 +99,8 @@ sys::ReturnCodes ServiceDBCommon::SwitchPowerModeHandler(const sys::ServicePower
     return sys::ReturnCodes::Success;
 }
 
-void ServiceDBCommon::sendUpdateNotification(db::Interface::Name interface, db::Query::Type type)
+void ServiceDBCommon::sendUpdateNotification(db::Interface::Name interface, db::Query::Type type, uint32_t recordId)
```
pholat, line 102: Tbh I do not like the solution here. What if there will be multiple records changed?
what if records will be removed? what if these will be added? 
Is there a way to not change the general notifications logic in ServiceDB (used in multiple apps and services in the bell and pure) and add the required logic in ServiceDesktop?

pholat (changes requested):
There are two aspects here that I think should be a bit worked through:

1. Please create classes and functions expressing how code should work, preferably with some additional logic control and unit tests instead of writing raw code in callbacks. 
2. Please try to solve issue with notifications not extending the general notification logic in the system, but solving problems locally in the service desktop. It's not the first place where additional data on notification was needed so I believe (better, or worse) examples could be found in code for this particular code.

pholat:
Talked through with @pawel-mudita :) https://github.com/mudita/MuditaOS/pull/3693#pullrequestreview-927057280 
docs will be added, the class will be added. The notification approach may be changed.

pholat (approved):
aprove in advance. Please do not merge till not agreed parts are done :+1:

module-services/service-db/ServiceDBCommon.cpp
pawel-mudita: Will be changed in the next PR, as discussed.
For now, it stays as is.

module-services/service-desktop/ServiceDesktop.cpp
pawel-mudita: Created a method in a separated class.

module-services/service-desktop/ServiceDesktop.cpp
pawel-mudita: Moved to function in a separate class.
Will be reworked in the next PR's.

pawel-mudita:
> Talked through with @pawel-mudita :) [#3693 (review)](https://github.com/mudita/MuditaOS/pull/3693#pullrequestreview-927057280) docs will be added, the class will be added. The notification approach may be changed.

Class added as discussed, but this is not its final version.
It will be reworked in upcoming PR's, alongside a new approach to notifications.
Docs: https://appnroll.atlassian.net/wiki/spaces/CP/pages/1599996137/Outbox+Endpoint+API+14

mjan-dev (approved):

PR #3698, opened 2022-03-30
[MOS-361] Navbar fonts size fix

paweljonskim:
Selection_430
RED - new look
BLACK - look before regression took place (community build)
So basically 100% overlap except time

Unfortunatelly the screen does not match MIRO perfectly (main time and date are shifted), but since im not fluent with Pure workflow I will only fix the refression I created and make a bug for the rest in JIRA


pawel-mudita (approved):

PrzeBrudny (approved):
Hmm a bit weird because as I recall all fonts how approx. 1 pixel per sign width issue coming from exporter so all should be a bit wider on Pure.

paweljonskim:
> Hmm a bit weird because as I recall all fonts how approx. 1 pixel per sign width issue coming from exporter so all should be a bit wider on Pure.

And it is :) The pic is not the MIRO to Pure comparison but rather OLD pure to NEW pure comparison :) so its all the same

SP2FET (approved):

PR #3699, opened 2022-03-30
[MOS-264] Connect signal strength and operator name with HFP profile

SP2FET:
**Description**

Connected signal strength and operator name with HFP profile so user can see the signal strength and operator name in the infotainment system.

There was also a stack modification needed:
https://github.com/mudita/btstack/pull/5

This PR is divided into 3 commits to be more eye-pleasant:
- Cellular preparation
- Command handling rework
- Connection of the signal strength and operator name messages

***IMPORTANT INFO***
Due to invalid logic in `profileManager` all of those statuses will be available after performing a phone call.
Or when the default profile in the manager will be changed to HFP. This bug will be investigated in MOS-347.

Make sure that this PR:

### DIFF
--- a/module-bluetooth/Bluetooth/BluetoothWorker.cpp
+++ b/module-bluetooth/Bluetooth/BluetoothWorker.cpp
@@ -10,6 +10,7 @@
 #include "interface/profiles/HSP/HSP.hpp"
 #include "audio/BluetoothAudioDevice.hpp"
 #include "BtKeysStorage.hpp"
+#include "command/DeviceData.hpp"
 
 #if DEBUG_BLUETOOTH_HCI_COMS == 1
 #define logHciComs(...) LOG_DEBUG(__VA_ARGS__)
@@ -95,7 +96,7 @@ BluetoothWorker::BluetoothWorker(sys::Service *service)
 {
     init({
         {queues::io, sizeof(bluetooth::Message), queues::queueLength},
-        {queues::cmd, sizeof(bluetooth::Command), queues::queueLength},
+        {queues::cmd, sizeof(bluetooth::Command::CommandPack), queues::queueLength},
         {queues::btstack, sizeof(bool), queues::triggerQueueLength},
     });
     registerQueues();
@@ -140,11 +141,12 @@ auto BluetoothWorker::run() -> bool
 
 auto BluetoothWorker::handleCommand(QueueHandle_t queue) -> bool
 {
-    bluetooth::Command command(bluetooth::Command::None);
-    if (xQueueReceive(queue, static_cast<void *>(&command), 0) != pdTRUE) {
+    bluetooth::Command::CommandPack pack{};
+    if (xQueueReceive(queue, static_cast<void *>(&pack), 0) != pdTRUE) {
         LOG_ERROR("Queue receive failure!");
         return false;
     }
+    auto command = bluetooth::Command(std::move(pack));
 
     switch (command.getType()) {
     case bluetooth::Command::PowerOn:
@@ -154,11 +156,12 @@ auto BluetoothWorker::handleCommand(QueueHandle_t queue) -> bool
     case bluetooth::Command::PowerOff:
         controller->turnOff();
         break;
-    case bluetooth::Command::Unpair:
+    case bluetooth::Command::Unpair: {
         controller->processCommand(command);
-        removeFromBoundDevices(command.getDevice().address);
-        handleUnpairDisconnect(command.getDevice());
-        break;
+        auto device = std::get<Devicei>(command.getData());
+        removeFromBoundDevices(device.address);
+        handleUnpairDisconnect(device);
+    } break;
     case bluetooth::Command::None:
         break;
     default:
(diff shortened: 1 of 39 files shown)


module-bluetooth/Bluetooth/BluetoothWorker.cpp
```diff
@@ -156,15 +157,16 @@ auto BluetoothWorker::handleCommand(QueueHandle_t queue) -> bool
...
+        removeFromBoundDevices(std::get<Devicei>(command.getData()).address);
+        handleUnpairDisconnect(std::get<Devicei>(command.getData()));
         break;
     case bluetooth::Command::None:
         break;
     default:
         controller->processCommand(command);
         break;
     }
+    command.cleanup();
```
SP2FET, line 169: I can do cleanup while executing the `getData()` method, but then I don't have any control of the data when the user doesn't retrieve the data from the command. What would you suggest @pholat @swichu91

module-bluetooth/Bluetooth/command/Command.hpp
```diff
@@ -0,0 +1,57 @@
...
+         *  FreeRTOS queue.
+         */
+        Command(Command::Type type, const std::shared_ptr<CommandData> &data);
+        explicit Command(Command::Type type) : type(type)
+        {}
+        /*
+         * Be aware of calling cleanup() after command with data processing has finished, otherwise
+         * a memory leak will occur.
+         */
+        void cleanup();
```
SP2FET, line 48: the same goes here @pholat @swichu91

module-bluetooth/Bluetooth/BluetoothWorker.cpp
pholat: cleaning up before function exit is not bad in here. This is very c like - and handleCommand is very c style function :)

I would think of a way to remove the switch/case needed here. If we could move it somewhere else for parsing? Then we can unit test parsing :) (then use data in designated handlers after it's parsed) especially that there is generic `else: I don't know what I do in switch case -> controler->procesCommand(command)`

module-bluetooth/Bluetooth/CommandHandler.cpp
```diff
@@ -51,15 +51,15 @@ namespace bluetooth
         case bluetooth::Command::StopScan:
             return stopScan();
         case bluetooth::Command::Pair:
-            return pair(command.getDevice());
+            return pair(std::get<Devicei>(command.getData()));
```
pholat, line 54: can this logic be moved to generic command? i.e. it would just return variant on `getParsed` it would require knowing all command types in the base class. ( Could be moved to "parser" class though)

module-bluetooth/Bluetooth/command/Command.hpp
pholat: Why not change it so that data sent and received is data only? You will be able to create a Command from it - but it it'self would be dumb which would only represent the shared memory sent:
`size_t size, void*/char* data`
Then it would be passed through a parser that would:
```
take data size pair
decode msgpack with:
[command id, size, whatever else]
pass it for further processing :)
```

pholat (commented):
Looks, ok, a bit of polishing should be required prior merging to master :)

module-bluetooth/Bluetooth/BluetoothWorker.cpp
swichu91: Do we need to manage underlying data manually? We can store `std::shared_ptr` as a `Command` member instead of a raw pointer. By doing so, memory will be managed automatically. The only potential problem I see here is raw FreeRTOS queue doesn't take ownership of the data it contains.

SP2FET: @swichu91 that's how it's made - the queue makes the problem as after copying the previous owner makes the data out of scope, so it's getting deleted. I tried with `shared_ptr` and I've always received `nullptr`.

@pholat I think that's possible, but I did this command refactor to be able to finish the task - I didn't want to rewrite whole BT with this job :P

module-bluetooth/Bluetooth/CommandHandler.cpp
SP2FET: the base class has to know all types because of `DataVariant`. I think it's possible to do a generic command and return there any optional data, but it makes another layer of decision making - either switch/case or overloaded methods for each data type

module-bluetooth/Bluetooth/command/Command.hpp
SP2FET: The case is that sometimes we send command without data - when we serialize both command and data to one buffer then indeed, it would make it much simpler, but then during deserializing we would have to somehow process the command type - we loose the polymorphism here.

module-bluetooth/Bluetooth/CommandHandler.cpp
```diff
@@ -39,7 +39,7 @@ namespace bluetooth
         this->driver->registerPowerOnCallback([profilePtr = this->profileManager]() { profilePtr->init(); });
     }
 
-    Error::Code CommandHandler::handle(Command command)
+    Error::Code CommandHandler::handle(Command &command)
```
pholat, line 42: can this switch case be removed with proper variant use? :)

SP2FET: I don't think so - some commands do not use additional data, so there's no way to distinguish commands via data type

pholat (approved):

PR #3701, opened 2022-03-31
Revert "[MOS-343] Harmony defined flash size is less than physical size"

paweljonskim:
This reverts commit 1a2c4e45f593397e539c99efdda3d81a758dde70.
This commit causes inabilty to flash image to Harmony


mjan-dev (approved):

swichu91 (approved):
Do we know why it was changed?

paweljonskim:
> Do we know why it was changed?

no :( tried to get to it, but @mjan-dev  has no idea and the author is no longer available

PR #3704, opened 2022-03-31
MOS-258 Ui tooling to see the system usage

pholat:
Requires:
https://github.com/mudita/MuditaOSTop/pull/1

It enables us to use RTT channel 1 to gather system usage.

Make sure that this PR:

Thanks for your work ♥

### DIFF
--- a/module-bsp/sink/FileSink.cpp
+++ b/module-bsp/sink/FileSink.cpp
@@ -1,0 +1,19 @@
+// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
+// For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
+
+#include "sink/FileSink.hpp"
+
+FileSink::FileSink(const std::string &filepath)
+{
+    fs.open(filepath, std::fstream::out);
+}
+
+FileSink::~FileSink()
+{
+    fs.close();
+}
+
+void FileSink::put(const char *data, size_t size)
+{
+    fs.write(data, size);
+}
--- a/module-sys/SystemManager/CpuPackPrinter.cpp
+++ b/module-sys/SystemManager/CpuPackPrinter.cpp
@@ -3,15 +3,25 @@
 
 #include "SystemManager/CpuPrinter.hpp"
 #include "SystemManager/SysCpuUpdateResult.hpp"
-#include <SystemManager/SystemManagerCommon.hpp>
-#include "third-party/msgpack11/msgpack11/msgpack11.hpp"
-#include "battery_charger/battery_charger.hpp"
-#include "time/time_conversion.hpp"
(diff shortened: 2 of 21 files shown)


SP2FET (approved):
Is the magic_enum submodule update necessary?

module-sys/SystemManager/stats/stats/Sink.hpp
```diff
@@ -0,0 +1,13 @@
...
+
+#pragma once
+
+#include <cstddef>
+
+class Sink
+{
+  public:
+    virtual ~Sink()                                 = default;
+    virtual void put(const char *data, size_t size) = 0;
```
swichu91, line 12: Consider replacing data/size pair args with std::span.

Maciej-Mudita (approved):

module-sys/SystemManager/stats/stats/Sink.hpp
SP2FET: but do we support C++20? :D

swichu91: Good point. https://en.cppreference.com/w/cpp/compiler_support says  std::span was introduced in GCC10, so it should work. We might not have C++20 enabled in CMake, though.

swichu91: I tried to enable the c++20 support, but it seems that some parts of the nxp fsl library use deprecated constructs(`compound assignment with 'volatile'-qualified left operand is deprecated [-Werror=volatile]`). Of course, it can be easily fixed, but I believe such actions are out of the scope of this PR.

swichu91 (commented):
Please check comments. Apart from that it looks good to me :)

swichu91 (approved):

module-bsp/sink/FileSink.cpp
```diff
@@ -0,0 +1,19 @@
...
+
+FileSink::~FileSink()
+{
+    fs.close();
+}
+
+void FileSink::put(const char *data, size_t size)
+{
+    fs.write(data, size);
+}
```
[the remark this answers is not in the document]
pholat: it's it's not used at the moment in code - as sink is selected in build time ( not in runtime via configuration :) )
I think it should be changed and possibly expanded so that we would have config which would set overall PowerManagement setup: select sinks for selected printer, set power profiles for cpu calling etc.

module-sys/SystemManager/CpuPackPrinter.cpp
```diff
@@ -20,23 +30,35 @@ namespace sys::cpu::stats
         Power   = 4,
     };
 
+    PackPrinter::PackPrinter()
+    {
+        sink = new sink::RTT(rtt_config);
```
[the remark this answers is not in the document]
pholat: sure, will require default deleter for this class. (added)

PR #3709, opened 2022-04-04
[BH-1443] Lower CPU freq before going to SNVS

paweljonskim:
Going to SNVS requires usage of periph clock
That reuires further cpu configuration changes
Easiest way is to just change cpu freq to level 2

### DIFF
--- a/module-bsp/board/rt1051/bellpx/board.cpp
+++ b/module-bsp/board/rt1051/bellpx/board.cpp
@@ -5,12 +5,13 @@
 #include "board.h"
 #include "drivers/gpio/DriverGPIO.hpp"
 #include "board/BoardDefinitions.hpp"
+#include "lpm/CpuFreqLPM.hpp"
 
 namespace
 {
     using namespace drivers;
 
-    void board_power_off()
+    void power_off()
     {
         /// No memory allocation here as this specific GPIO was initialized at the startup. We are just grabbing here a
         /// reference to the already existing object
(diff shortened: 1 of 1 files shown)


module-bsp/board/rt1051/bellpx/board.cpp
```diff
@@ -10,7 +10,7 @@ namespace
 {
     using namespace drivers;
 
-    void board_power_off()
+    void board_exit_power_off()
```
swichu91, line 13: How about simple `power_off`?

module-bsp/board/rt1051/bellpx/board.cpp
```diff
@@ -27,7 +27,7 @@ namespace
         SNVS->LPCR |= SNVS_LPCR_TOP(1); /// Enter SNVS mode
     }
 
-    void board_reset()
+    void board_exit_reset()
```
swichu91, line 30: Is it exit or reset? ;) How about simple `reset`?

mjan-dev (approved):

module-bsp/board/rt1051/bsp/lpm/RT1051LPMCommon.cpp
```diff
@@ -28,6 +28,7 @@ namespace bsp
 
     int32_t RT1051LPMCommon::PowerOff()
     {
+        SetCpuFrequency(bsp::CpuFrequencyMHz::Level_2);
```
swichu91, line 31: Shouldn't it be placed in the Bell specific board exit?

paweljonskim: well if we want this fix only for Harmony then okay :)

swichu91: From what I know, Pure does not use SNVS.

mjan-dev: Yes, Pure indeed doesn't use SNVS.

swichu91 (approved):

PR #3710, opened 2022-04-04
[BH-1442] Eliminate race on shutdown screen

paweljonskim:
There is another race still possible, but far less likely


pholat (commented):
do you know that we actually have a mechanism which enables us to close service one after another, why cant it be used?

swichu91:
> do you know that we actually have a mechanism which enables us to close service one after another, why cant it be used?

It would be nice to avoid flags/conditionals if possible. It only makes code harder to analyze and more prone to errors.

PR #3711, opened 2022-04-05
[BH-1446] RTC Calibration per product

paweljonskim:
RTC Calibration per product


swichu91 (commented):
It looks okay, but IMHO we can do better. Please, notice that `getRtcCalibrationValue` is only used by the RTC implementation thus there is no need to put it into public API. If you don't want to create another header file only for a function declaration, you can use the `extern` declaration inside .cpp.

swichu91 (approved):

tomaszrybarski-digica (approved):

mjan-dev (approved):

PR #3715, opened 2022-04-05
[BH-1433] Chime Volume Model Fix

tomaszrybarski-digica:
Fixed saving of volume in Chime Model

**Description**
This fixes the following issues of [BH-1433]
- Proper saving of volume value
- Activated subscription to Audio Notifications for volume change callback to properly work

[BH-1433]: https://appnroll.atlassian.net/browse/BH-1433?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/products/BellHybrid/apps/application-bell-meditation-timer/models/ChimeVolume.cpp
+++ b/products/BellHybrid/apps/application-bell-meditation-timer/models/ChimeVolume.cpp
@@ -8,11 +8,12 @@ namespace app::meditation::models
 
     void ChimeVolume::setValue(std::uint8_t value)
     {
+        defaultValue = value;
         audioModel.setVolume(value, AbstractAudioModel::PlaybackType::Meditation, {});
     }
     std::uint8_t ChimeVolume::getValue() const
     {
-        return defaultValue;
+        return audioModel.getVolume(AbstractAudioModel::PlaybackType::Meditation).value_or(0);
     }
 
     void ChimeVolume::restoreDefault()
--- a/products/BellHybrid/apps/application-bell-meditation-timer/MeditationTimer.cpp
+++ b/products/BellHybrid/apps/application-bell-meditation-timer/MeditationTimer.cpp
@@ -30,7 +30,9 @@ namespace app
                                      StartInBackground startInBackground,
                                      uint32_t stackDepth)
         : Application(std::move(name), std::move(parent), statusIndicators, startInBackground, stackDepth)
-    {}
+    {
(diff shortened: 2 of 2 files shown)


swichu91 (approved):

products/BellHybrid/apps/application-bell-meditation-timer/models/ChimeVolume.cpp
```diff
@@ -8,11 +8,12 @@ namespace app::meditation::models
 
     void ChimeVolume::setValue(std::uint8_t value)
     {
+        defaultValue = value;
```
paweljonskim, line 11: what is the point of defaultValue if it is changed on each setVolume?
It will be always the same as value stored in model, right?

swichu91: Good point! From what I see in the settings application, it works because models are recreated each time a user enters the application. In this case model is created only once and it is loaded only once with the default value.

tomaszrybarski-digica: It has to be done this way by "remembering" the previously set variable in a class member defaultValue, otherwise we have erroneous behavior when returning back without save (back press). Without storing defaultValue, e.g if we do just this: 

`    void ChimeVolume::restoreDefault()
    {
        setValue(audioModel.getVolume(AbstractAudioModel::PlaybackType::Meditation).value_or(0));
    }
`
If we change the volume from e.g. 6 to 7 and don't approve by light press, just use back press, upon re-entering Volume settings again we will see a value of 7 which is incorrect

tomaszrybarski-digica: > Good point! From what I see in the settings application, it works because models are recreated each time a user enters the application. In this case model is created only once and it is loaded only once with the default value.

The fact that the model here is created only once implies that defaultValue variable has to be updated every time in setValue() since it won't get the proper updated value from the models constructor (as it is in the models in settings app). This implementation seems to work well on the hardware

swichu91: Update: It also works on the hardware :)

mjan-dev (approved):

PR #3720, opened 2022-04-06
Fix permissions issues with scripts

GravisZro:
**Description**

Some scripts use programs that require elevated permissions to execute properly because they use commands like `mke2fs` and `sfdisk` which are privileged by default. An additional script has been added to use the `sudo` command but only if `sudo` is installed, it's necessary, is non-interactive, it would provide the user with the required permissions.

**Problems with `sudo` in `config/bootstrap.sh` that are fixed by `maysudo`:**
* checks if `sudo` is _installed_
* checks if `sudo` is _required_ for a particular command (i.e. current user lacks rights)
* checks if you the appropriate _rights_ to run a program using `sudo` (i.e. if permitted via sudoers)
* checks if `sudo` will run _non-interactively_ (so you don't get prompted 20 times per build)

**Repository directory identification scheme**
```bash
if [ -z ${L_GIT_DIR+unbound} ] && [ -z $(export L_GIT_DIR=$(git rev-parse --show-toplevel 2>/dev/null)) ]; then
  export L_GIT_DIR=$(readlink -f "$(dirname "$0")/../../..");  # varies depending on the script location within the repository
fi
```
This line of code translates to this:
> **IF** `L_GIT_DIR` is unbound **AND** export results in a null string (because the `git` command failed) **THEN** export a relative location

Without this, when you run a script using `sudo` it will spit out
```
fatal: unsafe repository ('path2repo/MuditaOS' is owned by someone else)
To add an exception for this directory, call:

        git config --global --add safe.directory path2repo/MuditaOS
```

Make sure that this PR:

### DIFF
--- a/module-platform/linux/tests/gendisktestimg.sh
+++ b/module-platform/linux/tests/gendisktestimg.sh
@@ -1,25 +1,34 @@
 #!/bin/bash -e
-# Copyright (c) 2017-2021, Mudita Sp. z.o.o. All rights reserved.
+# Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
 # For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
-#Copyright (c) 2017-2021, Mudita Sp. z.o.o. All rights reserved.
+#Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
 #For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 if [ $# -ne 1 ]; then
 	echo "Error! Invalid argument count"
 	exit -1
 fi
+
+if [ -z ${L_GIT_DIR+unbound} ] && [ -z $(export L_GIT_DIR=$(git rev-parse --show-toplevel 2>/dev/null)) ]; then
+  export L_GIT_DIR=$(readlink -f "$(dirname "$0")/../../..");
+fi
+
+. "$L_GIT_DIR"/tools/locate_bins.sh
+require_unprivileged dd head tail
+require_commands sfdisk
+
 IMAGE_FILE_DIR="$1"
 
 dd if=/dev/zero of=$IMAGE_FILE_DIR/test_disk.img bs=512 count=8
-sfdisk $IMAGE_FILE_DIR/test_disk.img << ==sfdisk
+maysudo sfdisk "$IMAGE_FILE_DIR/test_disk.img" << ==sfdisk
 ,1,L
 ,1,L
 ,,L
 ==sfdisk
 
 dd if=/dev/zero of=$IMAGE_FILE_DIR/test_disk_ext.img bs=512 count=16
-sfdisk $IMAGE_FILE_DIR/test_disk_ext.img << ==sfdisk
+maysudo sfdisk "$IMAGE_FILE_DIR/test_disk_ext.img" << ==sfdisk
 ,1,L
 ,,E
 ,1,L
@@ -31,7 +40,7 @@ sfdisk $IMAGE_FILE_DIR/test_disk_ext.img << ==sfdisk
 ==sfdisk
 
 dd if=/dev/zero of=$IMAGE_FILE_DIR/test_disk_bad_tmp.img bs=512 count=16
-sfdisk $IMAGE_FILE_DIR/test_disk_bad_tmp.img << ==sfdisk
+maysudo sfdisk "$IMAGE_FILE_DIR/test_disk_bad_tmp.img" << ==sfdisk
 ,1,L
 ==sfdisk
 
--- a/module-platform/linux/tests/genlfstestimg.sh
+++ b/module-platform/linux/tests/genlfstestimg.sh
@@ -1,5 +1,5 @@
 #!/bin/bash -e
-#Copyright (c) 2017-2021, Mudita Sp. z.o.o. All rights reserved.
+#Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
 #For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 usage() {
@@ -27,19 +27,20 @@ if [ ! -d "${SYSROOT}/sys" ]; then
 	exit -1
 fi
 
-_REQ_CMDS="sfdisk truncate"
-for cmd in $_REQ_CMDS; do
-	if [ ! $(command -v $cmd) ]; then
-		echo "Error! $cmd is not installed, please use 'sudo apt install' for install required tool"
-		exit -1
-	fi
-done
+if [ -z ${L_GIT_DIR+unbound} ] && [ -z $(export L_GIT_DIR=$(git rev-parse --show-toplevel 2>/dev/null)) ]; then
+  export L_GIT_DIR=$(readlink -f "$(dirname "$0")/../../..");
+fi
+
+. "$L_GIT_DIR"/tools/locate_bins.sh
+require_unprivileged truncate stat
+require_commands sfdisk
+
 truncate -s $IMAGE_SIZE $IMAGE_FILE
 
 SECTOR_START=2048
 SECTOR_END=$(( $(stat -c "%s" $IMAGE_FILE)/512 - $SECTOR_START))
 
-sfdisk $IMAGE_FILE << ==sfdisk
+maysudo sfdisk "$IMAGE_FILE" << ==sfdisk
 label: dos
 unit: sectors
 
--- a/tools/generate_image.sh
+++ b/tools/generate_image.sh
@@ -35,13 +35,15 @@ if [ ! -d "$SYSROOT" ]; then
 	usage
 	exit -1
 fi
-_REQ_CMDS="sfdisk mtools awk truncate"
-for cmd in $_REQ_CMDS; do
-	if [ ! $(command -v $cmd) ]; then
-		echo "Error! $cmd is not installed, please use 'sudo apt install' for install required tool"
-		exit -1
-	fi
-done
+
+if [ -z ${L_GIT_DIR+unbound} ] && [ -z $(export L_GIT_DIR=$(git rev-parse --show-toplevel 2>/dev/null)) ]; then
+  export L_GIT_DIR=$(readlink -f "$(dirname "$0")/..");
+fi
+
+. "$L_GIT_DIR"/tools/locate_bins.sh
+require_unprivileged truncate basename realpath sync
+require_commands sfdisk mke2fs mtools mformat mcopy mmd awk
+
 #mtools version
 _AWK_SCRIPT='
 /[0-9]/ {
@@ -51,7 +53,7 @@ _AWK_SCRIPT='
 	}
 	exit 0;
 }'
-MTOOLS_OK=$(mtools --version | awk "${_AWK_SCRIPT}")
+MTOOLS_OK=$(maysudo mtools --version | awk "${_AWK_SCRIPT}")
 
 if [ ! $MTOOLS_OK ]; then
 	echo "Invalid mtools version, please upgrade mtools to >= 4.0.24"
@@ -71,7 +73,7 @@ PART3_SIZE=$(($DEVICE_BLK_COUNT - $PART1_SIZE - $PART2_SIZE - $PART1_START))
 echo "Remove previous image file"
 rm -f $IMAGE_NAME
 truncate -s $(($DEVICE_BLK_COUNT * $DEVICE_BLK_SIZE)) $IMAGE_NAME
-sfdisk $IMAGE_NAME << ==sfdisk
+maysudo sfdisk "$IMAGE_NAME" << ==sfdisk
 label: dos
 label-id: 0x09650eb4
 unit: sectors
@@ -84,10 +86,10 @@ unit: sectors
 
 # Format FAT partitions
 PART1="$IMAGE_NAME@@$(($PART1_START * $DEVICE_BLK_SIZE))"
-mformat -i "$PART1" -F -T $PART1_SIZE -M $DEVICE_BLK_SIZE -v MUDITAOS
+maysudo mformat -i "$PART1" -F -T "$PART1_SIZE" -M "$DEVICE_BLK_SIZE" -v MUDITAOS
 
 PART2="$IMAGE_NAME@@$(($PART2_START * $DEVICE_BLK_SIZE))"
-mformat -i "$PART2" -F -T $PART2_SIZE -M $DEVICE_BLK_SIZE -v RECOVER
+maysudo mformat -i "$PART2" -F -T "$PART2_SIZE" -M "$DEVICE_BLK_SIZE" -v RECOVER
 
 if [ ! -d "${SYSROOT}/sys" ]; then
 	echo "Fatal! Image folder sys/ missing in build. Check build system."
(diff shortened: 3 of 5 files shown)


tools/locate_bins.sh
```diff
@@ -0,0 +1,57 @@
...
+      return 2
+    fi
+  else
+    return 1
+  fi
+  eval "export -p call_${1}=\"$path_val\""
+  return 0
+}
+
+require_commands () {
```
pholat, line 43: what programs did you not have installed? these should be all installed via `bootstrap.sh` therefore checking for existence, in theory, shouldn't be required.
This is actually pretty important as bootstrap was used to create docker images too so that we have docker images with all tools required.

tools/locate_bins.sh
```diff
@@ -0,0 +1,57 @@
...
+  for name in $@; do
+    get_command $name
+    case $? in
+      1)
+        echo "Error: $name is not installed, please use 'sudo apt install' for install required tool"
```
pholat, line 48: We do not limit ourselves to apt-based Linux. Just write "follow your OS installation guide" or something like that.

tools/locate_bins.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/sh -e
+# Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
+# For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
+
+have_command () {
+  eval "$1 --version >> /dev/null 2>&1"
```
pholat, line 6: this is straight wrong! Many commands do not have either `--version` or `--help`. Commands are just programs somebody left on the path. 

One valid way to check if a command exists and should be checking which one should be in use. Example in fish:
```
[I] ⋊> ~ which cut; echo $status                                                                                                                                                                                                                       08:20:14
/usr/bin/cut
0
[I] ⋊> ~ which lololol; echo $status                                                                                                                                                                                                                   08:20:17
which: no lololol in (/home/pholat/.pyenv/bin:/home/pholat/.local/bin:/usr/local/bin:/usr/bin:/bin:/usr/local/sbin:/usr/lib/jvm/default/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl:/var/lib/snapd/snap/bin)
1
```

tools/locate_bins.sh
```diff
@@ -0,0 +1,57 @@
...
+  if [ $? -ne 0 ]
+  then
+    echo "false"
+  fi
+  echo "true"
+}
+
+eval "sudo -n echo >> /dev/null 2>&1"
```
pholat, line 14: use functions

tools/locate_bins.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/sh -e
```
pholat, line 1: its prefered to add:
```
#!/bin/env bash
```
Also always in new scripts add:
```
set -euo pipefail
```
Not only
```
set -e
```

tools/locate_bins.sh
pholat, line 14: can you integrate it so that bootstrap.sh and this script would have it added the same?

module-platform/linux/tests/genlfstestimg.sh
```diff
@@ -27,19 +26,15 @@ if [ ! -d "${SYSROOT}/sys" ]; then
...
 
-_REQ_CMDS="sfdisk truncate"
-for cmd in $_REQ_CMDS; do
-	if [ ! $(command -v $cmd) ]; then
-		echo "Error! $cmd is not installed, please use 'sudo apt install' for install required tool"
-		exit -1
-	fi
-done
-truncate -s $IMAGE_SIZE $IMAGE_FILE
+. $(dirname $0)/../../../tools/locate_bins.sh
```
pholat, line 29: this is error-prone when the directory changes. Use:
```
L_GIT_DIR=$(git rev-parse --show-toplevel)
```
(from:  config/style_check_hook.sh )

pholat (changes requested):
There are issues with this pr

tools/locate_bins.sh
GravisZro: I have all the programs installed, the problem is that they have to be invoked with root level permissions.  This is why I check that they can be invoked or invoked using sudo.  Checking that they are installed is merely a reflection of what the tools and tests scripts already do.

tools/locate_bins.sh
GravisZro: I'm OK with that.  The text was lifted from the other scripts.

tools/locate_bins.sh
GravisZro: `which` is not a POSIX program and thus there is decent chance it's not installed.  You are correct that some programs do not have `--version` but all the ones being checked do have it.  This point of this is not to identify it's location but to ensure that it executes.

module-platform/linux/tests/genlfstestimg.sh
GravisZro: OK.

tools/locate_bins.sh
GravisZro: This needs to be executed when the script is included.  Should I just make a function and then call it at the end of the script?

tools/locate_bins.sh
GravisZro: I've removed this function entirely.

tools/locate_bins.sh
GravisZro: `#!/bin/bash` seems to be the norm in scripts (configure.sh included) so I'm going to go with that.  `set -euo pipefail` is added.

tools/locate_bins.sh
GravisZro: I don't know what you mean by this.

tools/generate_image.sh
```diff
@@ -14,48 +16,40 @@ Usage: $(basename $0) image_path image_partitions build_dir [version.json_file]
...
 #mtools version
-_AWK_SCRIPT='
-/[0-9]/ {
-	split($4,vers,".");
-	if(vers[1]>=4 && vers[2]>=0 && vers[3] >= 24) {
-		print "true";
-	}
-	exit 0;
-}'
-MTOOLS_OK=$(mtools --version | awk "${_AWK_SCRIPT}")
```
GravisZro, line 54: Note: This AWK script had to be replaced because it was fundamentally flawed.  It would cause version "5.0.0" to fail.

tools/generate_image.sh
```diff
@@ -14,51 +16,55 @@ Usage: $(basename $0) image_path image_partitions build_dir [version.json_file]
...
+fi
+if [ $# -gt 4 ]; then
+	VERSION_FILE="$5"
+fi
+if [ $# -gt 5 ]; then
+	BIN_FILE="$6"
+fi
+if [ $# -gt 6 ]; then
+	UPDATER_FILE="$7"
+fi
```
GravisZro, line 53: `set -u` requires `$4`/`$5`/`$6`/`$7` only be used if they are defined and therefore we must define variables as being empty strings beforehand.

tools/generate_image.sh
```diff
@@ -14,51 +16,55 @@ Usage: $(basename $0) image_path image_partitions build_dir [version.json_file]
 ==usage
 }
 
-if [[ ( $# -lt 3 )  || ( $# -gt 8 ) ]]; then
+if [ $# -lt 3 ] || [ $# -gt 7 ]; then
```
GravisZro, line 19: Avoid bash-isms.  Also, seven is the maximum number of arguments, not eight!

module-platform/linux/tests/gendisktestimg.sh
```diff
@@ -1,25 +1,31 @@
...
+set -uo pipefail
 
 if [ $# -ne 1 ]; then
 	echo "Error! Invalid argument count"
-	exit -1
+	exit 1
 fi
+
+L_GIT_DIR=$(git rev-parse --show-toplevel)
+if [ ! -d "$L_GIT_DIR" ]; then L_GIT_DIR=$(readlink -f "$(dirname "$0")"/../../..); fi
```
pholat, line 13: can you please just use the L_GIT_DIR or fail? 
what and why do you want to accomplish here?

tools/locate_bins.sh
pholat: @GravisZro what do you mean "is not POSIX program" can you please elaborate? And frankly can you please tell any distro that didn't have which out of the box, I've never met any :) Saying that - it's just a method that works, and `--version` checking doesn't. 
Saying that - you can check if the command exists, then it is executable, and even if its binary form matches the platform you are running it on. What's important is - what actually do you want to assert here.

tools/locate_bins.sh
pholat: what's wrong with one-line functions? it would at least explain this line's purpose.
Generally speaking, it's cleaner to write bash scripts in a manner where you write functions and then just execute them at the end of the script or in a separate script for longer utilities. It improves readability and debugging.

tools/locate_bins.sh
pholat: As for why use bin env vs no bin env - here you have the explanation. https://stackoverflow.com/a/16365367/4239386
Thats a minor.

`-e` is important - you need commands to fail - then **catch the fail** explicitly.

tools/locate_bins.sh
pholat: Did you read how bootstrap.sh handles sude?

tools/locate_bins.sh
pholat: Did you check how it's done in bootstrap? do we really require to create bash wrappers just for wrappers?
These can be run with root, but you can add them to your sudo config to run them without sudo, add group, etc. to not use root permissions etc.

pholat (changes requested):
1. arent we trying to kill a fly with a bazooka here? please check how bootstrap handles it if it really matters to you and trim down changes to what's only required. Do not try to invent new tooling which is not required nor common.
2. this PR has more and move changes diverging from the main topic which was the root access requirement for commands

tools/version_check.sh
```diff
@@ -0,0 +1,35 @@
...
+L_GIT_DIR=$(git rev-parse --show-toplevel)
+if [ ! -d "$L_GIT_DIR" ]; then L_GIT_DIR=$(readlink -f "$(dirname "$0")"/..); fi
+
+. "$L_GIT_DIR"/tools/locate_bins.sh
+require_commands head sed cut
+
+# first argument is the "binname --version" output
+# second argument is the minimum acceptable version
+minimum_version () {
+  VERSION=$(echo "$1" | $call_sed -e "s/[^0-9]*\([0-9]\+\)\.\([0-9]\+\)\.\([0-9]\+\).*/\\1.\\2.\\3/")
```
pholat, line 17: you are aware that not all commands are following sem ver? So in fact this function is not generic it says it's.
Also - this function is so overcomplicated...
you can just:
1. read the sem ver to array ( see how `IFS='.' read -r -a <<< $string` would work) 
```
[pholat@pholat-mudita ~]$ echo $string
1.2.3
[pholat@pholat-mudita ~]$ IFS='.' read -r -a array <<< "$string"
[pholat@pholat-mudita ~]$ echo ${array[1]}
2
```
2. check against the second array

With variable checks enabled it will be done without any external commands and in something like 3 lines.

tools/locate_bins.sh
GravisZro: I have checked bootstrap script and it simply uses `sudo` without 
* checking if `sudo` is installed
* checking if `sudo` is required for a particular command
* checking if you the right to run a program using `sudo`
* checking if `sudo` will run non-interactively (it will prompt you literally five times for the same password in the boostrap script)

The when making the command wrappers, the script checks if `sudo` is installed but **only** uses it when it is needed (current user lacks rights), allowed (you are permitted via sudoers), and non-interactive (so you don't get prompted 20 times per build).

tools/locate_bins.sh
GravisZro: There is nothing wrong with one-line functions, I simply was unsure of what you were trying to tell me.

tools/locate_bins.sh
GravisZro: Well the problem is that `/bin/env` doesn't exist but `/usr/bin/env` does.  For compatibility I'm sticking with `/bin/bash`

I've looked more into how `-e` works and now made sure to include it's use.

module-platform/linux/tests/gendisktestimg.sh
GravisZro: If you execute a script that relies upon `L_GIT_DIR` then it will currently fail if you run that script using `sudo` because `git` will spit out that you don't own the git repo.  Furthermore, if you download the source instead of cloning the git repo then it will also fail.  What I'm doing is providing a last ditch effort to not fail.

tools/version_check.sh
GravisZro: I believe the utilities that need to have their versions checked do follow semantic versioning.

Your version checking solution is quite nice but it relies on non-standard command behavior for `read`.  Meanwhile `awk` is standardized, it's not installed by default on many Linux systems, and it's behavior is inconsistent across implementations due to it's complexity.

I agree my solution is not very "clean" but it is highly portable.

GravisZro: I've remove the version checking modification from this PR.

GravisZro:
> 1. arent we trying to kill a fly with a bazooka here? please check how bootstrap handles it if it really matters to you and trim down changes to what's only required. Do not try to invent new tooling which is not required nor common.

I've revised how this functions.  I agree that the other solution was a bit too clever.  Instead of outright using `sudo` it uses a function called `maysudo` which may or may not use `sudo`.

 
> 2. this PR has more and move changes diverging from the main topic which was the root access requirement for commands

Yeah, that's my bad and it's been rectified.  I will move other changes to other PRs.

module-platform/linux/tests/gendisktestimg.sh
```diff
@@ -1,25 +1,34 @@
...
 #For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 if [ $# -ne 1 ]; then
 	echo "Error! Invalid argument count"
 	exit -1
 fi
+
+if [ -z ${L_GIT_DIR+unbound} ] && [ -z $(export L_GIT_DIR=$(git rev-parse --show-toplevel 2>/dev/null)) ]; then
+  export L_GIT_DIR=$(readlink -f "$(dirname "$0")/../../..");
+fi
```
GravisZro, line 15: This line of code translates to this:
> **IF** `L_GIT_DIR` is unbound **AND** export results in a null string **THEN** export a relative location

GravisZro: Without this, when you run a script using `sudo` it will spit out
```
fatal: unsafe repository ('path2repo/MuditaOS' is owned by someone else)
To add an exception for this directory, call:

        git config --global --add safe.directory path2repo/MuditaOS
```

tools/locate_bins.sh
GravisZro: I have checked bootstrap script and it simply uses `sudo` without 
* checking if `sudo` is installed
* checking if `sudo` is required for a particular command
* checking if you the right to run a program using `sudo`
* checking if `sudo` will run non-interactively (it will prompt you literally five times for the same password in the boostrap script)

The when making the command wrappers, the script checks if `sudo` is installed but **only** uses it when it is needed (current user lacks rights), allowed (you are permitted via sudoers), and non-interactive (so you don't get prompted 20 times per build).

GravisZro:
@pholat 
Are there any remaining issues that need addressing?  If not, this commit needs a ticket number.

PR #3724, opened 2022-04-08
[MOS-380]  Sentinels for each that removes unused data added

pholat:
As in topic :)

Make sure that this PR:

Thanks for your work ♥

### DIFF
--- a/module-sys/SystemManager/GovernorSentinelOperations.cpp
+++ b/module-sys/SystemManager/GovernorSentinelOperations.cpp
@@ -1,0 +1,49 @@
+// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
+// For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
+
+#include "GovernorSentinelOperations.hpp"
+
+namespace sys
+{
+
+    void for_each_governed_sentinel(GovernorSentinelsVector &sentinels, governed_callback &foo)
+    {
+        for (auto sentinel = sentinels.rbegin(); sentinels.rend() != sentinel; ++sentinel) {
+            // remove dangling unique ptrs
+            if (not *sentinel) {
+                sentinels.erase(sentinel.base());
+                continue;
+            }
+            auto sentinelWeakPointer = (*sentinel)->GetSentinel();
+            // remove expired ptrs
+            if (sentinelWeakPointer.expired()) {
+                sentinels.erase(sentinel.base());
+                continue;
+            }
+            // call foo on sentinel
+            if (foo(**sentinel)) {
+                return;
+            };
+        }
+    };
+
+    void for_each_sentinel(GovernorSentinelsVector &sentinels, sentinel_foo &foo)
+    {
+        governed_callback cb = [&](const GovernorSentinel &sentinel) {
+            std::shared_ptr<CpuSentinel> sharedResource = sentinel.GetSentinel().lock();
+            if (sharedResource) {
+                if (foo(sharedResource)) {
+                    return true;
+                }
+            }
+            return false;
+        };
+        for_each_governed_sentinel(sentinels, cb);
+    };
+
+    void clean(GovernorSentinelsVector &sentinels)
+    {
+        governed_callback empty_governed_callback = [&](const GovernorSentinel &) { return false; };
+        for_each_governed_sentinel(sentinels, empty_governed_callback);
+    }
+} // namespace sys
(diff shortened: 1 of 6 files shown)


Maciej-Mudita (approved):

module-sys/SystemManager/tests/unittest_CpuSentinelsGovernor.cpp
```diff
@@ -90,3 +92,110 @@ TEST_CASE("Power Manager CPU sentinels governor test")
...
 }
+
+TEST_CASE("GovernorSentinelsVector - single sentinel add and remove")
+{
+    using namespace sys;
+    GovernorSentinelsVector container;
+    std::shared_ptr<CpuSentinel> new_sentinel = std::make_shared<CpuSentinel>("test", nullptr);
+    container.push_back(std::make_unique<GovernorSentinel>(new_sentinel));
+
+    SECTION("single sentinel added and removed - count til end")
```
mjan-dev, line 103:
```suggestion
    SECTION("single sentinel added and removed - count till end")
```

module-sys/SystemManager/GovernorSentinelOperations.cpp
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
+// For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
+
+#include "GovernorSentinelOperations.hpp"
+
+namespace sys
+{
+
+    void for_each_governed_sentinel(GovernorSentinelsVector &sentinels, governed_callback &foo)
```
mjan-dev, line 9: Why underscore style? 
The naming as well doesn't imply what it really does. Maybe cleanInvalidSentinels/checkSentinels or smth like this.

module-sys/SystemManager/GovernorSentinelOperations.cpp
```diff
@@ -0,0 +1,49 @@
...
+                continue;
+            }
+            // call foo on sentinel
+            if (foo(**sentinel)) {
+                return;
+            };
+        }
+    };
+
+    void for_each_sentinel(GovernorSentinelsVector &sentinels, sentinel_foo &foo)
```
mjan-dev, line 30: Same

mjan-dev (approved):
Added some minor comments. Other than that seems good.

mjan-dev (commented):
Added some minor comments. Other than that seems good.

module-sys/SystemManager/GovernorSentinelOperations.cpp
pholat: 1. style: I followed the stl style here because it made sense to me that if normal for each is written `for_each` and if I would extend it to for each with if it would be `for_each_if` then here it would be `for_each_governed...` 
2. functionality: `cleanInvalidSentinels` would definitely be wrong. Please see that it's first and foremost for_each functionality with additional overzealous sanity check and invalid data removal. It's well documented in header and you can check how it works in tests :) 

I can change it to camel case and name it whatever you want, just tell me what would be your exact wishes :)

pholat: @SP2FET what part doesn't comply with our guideline? I didn't find a word about camel case in it

module-sys/SystemManager/GovernorSentinelOperations.cpp
```diff
@@ -0,0 +1,49 @@
...
+                continue;
+            }
+            auto sentinelWeakPointer = (*sentinel)->GetSentinel();
+            // remove expired ptrs
+            if (sentinelWeakPointer.expired()) {
+                sentinels.erase(sentinel.base());
+                continue;
+            }
+            // call foo on sentinel
+            if (foo(**sentinel)) {
```
mjan-dev, line 24: should this return if foo return true?

module-sys/SystemManager/GovernorSentinelOperations.cpp
mjan-dev: I get the stl style motivation and I accept it. As for guidelines, we prefer camelCase in most cases, however we don't force using it everywhere (if it is justified). Please consider moving cleaning of dangling and expired sentinels to another method and run it separately with clear intent as those checks are needed, however it is not obvious that those are done in a for_each type of method.

pholat: I know, I was there enforcing it in the first place and i.e. promoting automatic hooks with @rwicik :) to be completely honest we might revisit automatic style checks for functions and enums.
Saying that - these have documentation in header and I thought are trivial enough but sure, I'll move the `erase` line to another function

module-sys/SystemManager/GovernorSentinelOperations.cpp
pholat: yes, this was the goal - from the docs:
```
    /// loop over all sentinels:
    /// - uses for_each_governed_sentinel
    /// - may remove tracking non existent sentinels
    /// - returns on first sentinel_foo callback with 'true' result
```

PR #3743, opened 2022-04-14
[CP-1085][CP-1082] Collect sms and sms thread notifications in outbox ep

pawel-mudita:
**Description**

When sms thread record or sms record is modified during connection
with Mudita Center, notification of that action
should be collected in outbox endpoint to
keep contacts data synchronized between MuditaOS
and Mudita Center.
Documentation: 
[Outbox endpoint API](https://appnroll.atlassian.net/wiki/spaces/CP/pages/1599996137/Outbox+Endpoint+API+14)
[How to notify MC about updates to user data in almost real time](https://appnroll.atlassian.net/wiki/spaces/CP/pages/1598193852/How+to+notify+MC+about+updates+to+User+data+in+almost+real+time)

Make sure that this PR:

### DIFF
--- a/module-db/Interface/ContactRecord.cpp
+++ b/module-db/Interface/ContactRecord.cpp
@@ -35,8 +35,8 @@ ContactRecordInterface::ContactRecordInterface(ContactsDB *db)
 
 auto ContactRecordInterface::Add(ContactRecord &rec) -> bool
 {
-    bool ret = contactDB->contacts.add(ContactsTableRow{Record(DB_ID_NONE), .speedDial = rec.speeddial});
-    if (!ret) {
+    bool result = contactDB->contacts.add(ContactsTableRow{Record(DB_ID_NONE), .speedDial = rec.speeddial});
+    if (!result) {
         return false;
     }
     uint32_t contactID = contactDB->getLastInsertRowId();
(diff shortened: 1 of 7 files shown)


module-db/Interface/ContactRecord.cpp
```diff
@@ -595,6 +598,9 @@ auto ContactRecordInterface::updateQuery(const std::shared_ptr<db::Query> &query
     auto ret         = ContactRecordInterface::Update(updateQuery->rec);
     auto response    = std::make_unique<db::query::ContactUpdateResult>(ret);
     response->setRequestQuery(query);
+    if (ret) {
```
mjan-dev, line 601: minor suggestion: as you are at it, could you unify naming of functions returns to result in this interface? it is either ret or result :smile:

pawel-mudita: Ok, I'll unify this

module-db/Interface/ThreadRecord.cpp
```diff
@@ -203,6 +203,9 @@ std::unique_ptr<db::QueryResult> ThreadRecordInterface::markAsReadQuery(const st
 
     auto response = std::make_unique<db::query::MarkAsReadResult>(ret);
     response->setRequestQuery(query);
+    if (ret) {
+        response->setRecordID(localQuery->id);
```
mjan-dev, line 207: just asking, should the response id be set only if removing was successful? same question for all queries.

pawel-mudita: From Mudita Center perspective only notification after successful operations makes sense.
But this is not a big problem if I'll remove this condition.
In case of notification after an unsuccessful operation, MC just asks about the specific record id and checks its status.

mjan-dev: I agree in this use case, however I think it would be more logical to return as much info in query result as possible. If we already introduce the ID parameter, we could as well allow binding the result with it each time and process it as needed in e.g. application logic. I suppose it would be more universal and easier to use in the future.

pawel-mudita: Ok, I'll remove the condition

pawel-mudita: Done

mjan-dev (approved):

kkleczkowski (approved):

PR #3756, opened 2022-04-20
[MOS-347] Separate A2DP from HFP in ProfileManager

SP2FET:
**Description**

Now, both profiles can operate separately. Additonally, minor
fixes has been applied to allow calling via car audio system.
Some infotainment systems still don't work though.

Make sure that this PR:


kkleczkowski (approved):

module-bluetooth/Bluetooth/interface/profiles/HFP/HFP.cpp
```diff
@@ -422,18 +419,10 @@ namespace bluetooth
...
-            connect();
-        }
-        hfp_ag_set_speaker_gain(aclHandle, 8);
-        hfp_ag_set_microphone_gain(aclHandle, 10);
-    }
-
-    void HFP::HFPImpl::stop()
+    void HFP::HFPImpl::terminateCall() const noexcept
     {
         hfp_ag_terminate_call();
```
Lefucjusz, line 424: Do we have a possibility to distinguish between situations when call was answered, then ended and just rejected? As far as I understand, these two cases should be handled differently - when call was answered and then we want to end it, it should be terminated by hfp_ag_terminate_call(). If call was rejected without prior answering, it should be terminated via hfp_ag_call_dropped().

Explicitly calling hfp_ag_release_audio_connection() (line below) should not be needed when the logic is handled as I suggested...at least I hope so :P

Lefucjusz (changes requested):

module-bluetooth/Bluetooth/interface/profiles/HFP/HFP.cpp
SP2FET: we have more states that we should execute :
image

Those should be connected properly with cellular events based on the state machine diagram that you created in the MOS-366 task. 
I don't want to add next x flags to handle call logic when it should be properly used with events. I just added the terminateCall method as a complementary to initiateCall, as the audio profile has start/stop methods.

Lefucjusz: Ok, if it works, let's leave it as it is for now. I'll finish this state machine task ASAP and then we'll do it the way it is supposed to be done, not by adding more and more flags as a workaround.

.github/workflows/github-actions-pr-checklist-check.yml
```diff
@@ -5,6 +5,7 @@ on:
 
 jobs:
   task-check:
+    permissions: write-all
```
pholat, line 8: should it be here? :)

pholat (approved):
generally, I've nothing to say except what I said previously :) 
having a huge switch case in a state machine is counterproductive here. But I accept it's not a part of this pr.

.github/workflows/github-actions-pr-checklist-check.yml
SP2FET: this is a small fix for task-check for community :)

SP2FET: moved to separate PR

PR #3761, opened 2022-04-22
[BH-1469] Main alarm default volume

tomaszrybarski-digica:
Default main alarm volume set to 5 (BH-1469)
Default Bedtime time set to 22:00 (BH-1468)


swichu91 (approved):
As far as I remember, it was changed on purpose by @mjan-dev. Is it still relevant, or was it changed along the way?

paweljonskim (approved):

PR #3762, opened 2022-04-23
Fix word splitting in shell scripts

GravisZro:
**Description**

This patch adds quotes around variable names in shell scripts to ensure that paths containing spaces will no longer cause the script to fail.  Adding quotes to arrays has been explicitly been avoided to ensure nothing is broken.

See [ShellCheck SC2086](https://www.shellcheck.net/wiki/SC2086) for details about the issues not quoting can cause.

Make sure that this PR:

### DIFF
--- a/config/apply_update.sh
+++ b/config/apply_update.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright (c) 2017-2021, Mudita Sp. z.o.o. All rights reserved.
+# Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
 # For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 #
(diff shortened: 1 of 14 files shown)


config/apply_update.sh
```diff
@@ -73,35 +73,35 @@ if [ "$PHONE_DEV" == "" ]; then
...
 	sync
 fi
 
 echo "Copyind data"
-cp $TMPDIR/boot.bin $PHONE_MOUNT/current/
-cp $TMPDIR/Luts.bin $PHONE_MOUNT/current/
-cp $TMPDIR/country-codes.db $PHONE_MOUNT/current/
-cp $TMPDIR/version.json $PHONE_MOUNT/current/
-cp -r $TMPDIR/assets $PHONE_MOUNT/current/
+cp "$TMPDIR"/boot.bin "$PHONE_MOUNT"/current/
```
mjan-dev, line 100: Just a note. A whole expresion could be quoted. (e.g. "$TMPDIR/boot.bin").
In my opinion it is a more readable form.

mjan-dev (approved):
LGTM, added minor comment.

config/apply_update.sh
GravisZro: I've updated the commit to reflect the desired semantic path quoting.

PR #3764, opened 2022-04-25
[BH-1449] Fix gui/eink dependencies.

swichu91:
**Description**

* Dependencies between service-gui and service-eink
have been correctly defined.
* System close routine was replaced with the sequential one
which should solve many bell issues.
* It also fixes BH-1406 and BH-1442.
* Fixed some weird CMake issues in module-audio and
several other modules.
* Fixed service-gui closing routine
* Removed unneeded messages/states from service-gui/eink.
* Enhanced app-mgr closing routine. Fixed closing action handling.

Make sure that this PR:


kkleczkowski (approved):

module-services/service-gui/service-gui/ServiceGUI.hpp
```diff
@@ -67,24 +64,18 @@ namespace service::gui
...
-        void setState(State state) noexcept;
-        bool isInState(State state) const noexcept;
-
         sys::MessagePointer handleDrawMessage(sys::Message *message);
         sys::MessagePointer handleGUIRenderingFinished(sys::Message *message);
-        sys::MessagePointer handleEinkInitialized(sys::Message *message);
         sys::MessagePointer handleImageDisplayedNotification(sys::Message *message);
         sys::MessagePointer handleChangeColorScheme(sys::Message *message);
 
+        ::gui::Size displaySize;
```
mjan-dev, line 72: why is the displaySize added to ServiceGui in this PR?

module-sys/SystemManager/SystemManagerCommon.cpp
```diff
@@ -444,20 +463,41 @@ namespace sys
...
 
-        servicesPreShutdownRoutineTimeout = sys::TimerFactory::createPeriodicTimer(
-            this, "servicesPreShutdownRoutine", preShutdownRoutineTimeout, [this](sys::Timer &) { CloseServices(); });
-        servicesPreShutdownRoutineTimeout.start();
+        serviceCloseTimer = sys::TimerFactory::createSingleShotTimer(
+            this, "serviceCloseTimer", serviceCloseResponseTimeout, [this](sys::Timer &) { CloseServices(); });
+        serviceCloseTimer.start();
+
+        /// Start the sequence by sending close request to the first service on the list
+        bus.sendUnicast(std::make_shared<ServiceCloseReasonMessage>(closeReason), servicesToClose.front());
```
mjan-dev, line 478: Is it possible that some service doesn't close at all and blocks? If that happens CloseServices() is called, forcefully closing all the services left. When closing services in sequence, shouldn't CloseServices() close particular stuck service instead? Or is it out of the scope here?

module-services/service-gui/service-gui/ServiceGUI.hpp
swichu91: It was added because before display size was fetched from `EinkInitialized` message. This message was removed altogether as it is no longer needed, thanks to the properly defined dependencies. From now, such information is injected through service's constructor.

module-sys/SystemManager/SystemManagerCommon.cpp
swichu91: Yes, such a situation can happen. We can discuss it as I'm not sure if closing only non-responding service and then proceed with the rest normally will make any difference. In either case, it's abnormal situation that cannot be handled properly in system without MMU. Anyway I'm open for discussion.

mjan-dev: Fair point. I guess, it is acceptable to "abort" the proper termination anyway.

PR #3777, opened 2022-04-27
[BH-1449] Fix crash during startup

swichu91:
**Description**

* Moved all Bell-specific paths to AlarmSoundPaths.
* Fixed loading default layout logic. Currently, there is no
db update schema which in specific cases can lead to
layout setting field missing. In this case, load default layout and
update missing db field.
* Added support for Aarch64 Linux.

Make sure that this PR:


host-tools/genlittlefs/CMake/FindBLKID.cmake
```diff
@@ -25,7 +25,7 @@ else (APPLE)
 endif (APPLE)
 FIND_LIBRARY(BLKID_LIBRARIES
         NAMES ${BLKID_NAMES}
-        PATHS /lib64 /lib /usr/lib64 /usr/lib /usr/local/lib64 /usr/local/lib /usr/lib/x86_64-linux-gnu
+        PATHS /lib64 /lib /usr/lib64 /usr/lib /usr/local/lib64 /usr/local/lib /usr/lib/x86_64-linux-gnu /usr/lib/aarch64-linux-gnu
```
mjan-dev, line 28: is this change connected with the issue?

mjan-dev (approved):

host-tools/genlittlefs/CMake/FindBLKID.cmake
swichu91: No, but I decided I would add it here as it's so minor that there is no need to create separate PR for it. It enables support for Aarch64 Linux (as I pointed in the PR description).

SP2FET (approved):

PR #3782, opened 2022-04-28
Add VPI conversion tools

GravisZro:
**Description**
These are publicly available tools for converting VecMap Images to and from PNG files.  They rely on Netpbm to function properly.

Make sure that this PR:

### DIFF
--- a/tools/pngtohpp.sh
+++ b/tools/pngtohpp.sh
@@ -1,0 +1,168 @@
+#!/bin/bash
+# Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
+# For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
+
+set -euo pipefail
+
+build_tool () {
+  cat > "${1}.c" <<\EOF
+#include <stdint.h>
+#include <string.h>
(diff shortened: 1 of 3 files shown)


tools/pngtohpp.sh
```diff
@@ -0,0 +1,168 @@
+#!/bin/bash
+# Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
+# For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
+
+set -euo pipefail
+
+build_tool () {
```
pholat, line 7: 1. if you want to add c/c++ code add it properly to cmake as a target.
2. how is this tool documented and integrated?

pholat (changes requested):
image

Id doesn't compy - these checks have a purpose.

1. do not put c code in shell scripts
2. if you want to integrate it? integrate it properly...
3. add documentation for the code added ( see the checks you put)
4. add tests if possible. It should be fairly easy to add UT which would check if the image generated would be the same as from our tooling.

Lefucjusz:
@GravisZro our image converter repo has been opened recently, so I close this request. Feel free to contribute there.

PR #3791, opened 2022-04-29
[MOS-263] Fix DTMF handling in Pure

SP2FET:
**Description**

By accident, I've broke DTMF handling via Pure's keyboard
This was because of non strict description of API method.
Now, it has been unified to use number, not its ASCII code

Make sure that this PR:

### DIFF
--- a/module-bluetooth/Bluetooth/interface/profiles/HFP/HFP.cpp
+++ b/module-bluetooth/Bluetooth/interface/profiles/HFP/HFP.cpp
@@ -319,7 +319,12 @@ namespace bluetooth
         case HFP_SUBEVENT_TRANSMIT_DTMF_CODES: {
             auto digitStr = hfp_subevent_transmit_dtmf_codes_get_dtmf(event);
             LOG_DEBUG("Send DTMF Codes: '%s'\n", digitStr);
-            cellularInterface->sendDTMFCode(const_cast<sys::Service *>(ownerService), utils::toNumeric(digitStr));
+            try {
+                cellularInterface->sendDTMFCode(const_cast<sys::Service *>(ownerService), DTMFCode(digitStr));
+            }
+            catch (std::out_of_range &e) {
+                LOG_ERROR("Can't send DTMF code for digit: %s", digitStr);
+            }
             hfp_ag_send_dtmf_code_done(aclHandle);
         } break;
         case HFP_SUBEVENT_CALL_ANSWERED:
--- a/module-services/service-cellular/DTMFCode.cpp
+++ b/module-services/service-cellular/DTMFCode.cpp
@@ -1,0 +1,46 @@
+// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
+// For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
+
+#include "DTMFCode.hpp"
+#include <Utils.hpp>
+#include <stdexcept>
+
+DTMFCode::DTMFCode(const char digitChar)
+{
+    if ((digitChar >= '0' && digitChar <= '9') || digitChar == '*' || digitChar == '#') {
+        this->digitChar = digitChar;
+    }
+    else {
+        this->digitChar = 0;
+        throw std::out_of_range("DTMF: Invalid number!");
+    }
+}
+DTMFCode::DTMFCode(const char *digitStr)
+{
+    auto digitTmpString = std::string{digitStr};
+    if (digitTmpString == "*" || digitTmpString == "#") {
+        digitChar = digitTmpString[0];
+    }
+    else {
+        try {
+            auto digit = std::stoi(digitTmpString);
+            if (digit > 9) {
+                throw std::out_of_range("DTMF: Invalid number!");
+            }
+            digitChar = digit + '0';
+        }
+        catch (std::invalid_argument &e) {
+            LOG_ERROR("Can't parse digit string to DTMF ASCII code");
+            digitChar = 0;
+            throw std::out_of_range("DTMF: Invalid number!");
+        }
+    }
+}
+auto DTMFCode::getDigitASCIICode() const -> char
+{
+    return digitChar;
+}
+DTMFCode::operator std::string() const
+{
+    return "\"" + std::string(1, getDigitASCIICode()) + "\"";
+}
--- a/module-services/service-cellular/DTMFCode.hpp
+++ b/module-services/service-cellular/DTMFCode.hpp
@@ -1,0 +1,24 @@
+// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
+// For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
+
+#pragma once
+#include <stdint.h>
+#include <string>
+
+class DTMFCode
+{
+  public:
+    /// \brief Creates a DTMFCode instance via passing a digit character (its ASCII code)
+    /// \param digitChar - digit's ASCII code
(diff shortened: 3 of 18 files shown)


module-services/service-cellular/DTMFCode.cpp
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
+// For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
+
+#include "DTMFCode.hpp"
+#include <Utils.hpp>
+#include <stdexcept>
+
+DTMFCode::DTMFCode(const char digitChar)
```
pholat, line 8: I'm wondering - maybe add `_dtmf` user literat? :) https://en.cppreference.com/w/cpp/language/user_literal

module-services/service-cellular/DTMFCode.cpp
```diff
@@ -0,0 +1,46 @@
...
+{
+    if ((digitChar >= '0' && digitChar <= '9') || digitChar == '*' || digitChar == '#') {
+        this->digitChar = digitChar;
+    }
+    else {
+        this->digitChar = 0;
+        throw std::out_of_range("Invalid number!");
```
pholat, line 15: please add some more information (i.e "DTMF: Invalid number" in string just in case (as we are using basic exception)

module-services/service-cellular/DTMFCode.cpp
```diff
@@ -0,0 +1,46 @@
...
+{
+    auto digitTmpString = std::string{digitStr};
+    try {
+        auto digit = std::stoi(digitTmpString);
+        if (digit > 9) {
+            throw std::out_of_range("Invalid number!");
+        }
+        digitChar = digit + '0';
+    }
+    catch (std::invalid_argument &e) {
```
pholat, line 28: here actually I would check for `*` and `#` prior to stoi. 
Having in mind that:

_exception is meant to handle something unexpected, not to create if-else possibilities_

module-services/service-cellular/DTMFCode.hpp
```diff
@@ -0,0 +1,19 @@
...
+#include <string>
+
+class DTMFCode
+{
+  public:
+    explicit DTMFCode(char digitChar);
+    explicit DTMFCode(const char *digitString);
+    [[nodiscard]] auto getDigit() -> char;
+    [[nodiscard]] auto getDigitASCIICode() -> char;
+    [[nodiscard]] auto getDTMFCodeString() -> const std::string;
```
pholat, line 15: why we have such redundancy: we are using DTMF so we know that we are using DTMF string if we return string? So why not even just add `operator str()` or if you really have to `toString` method?

I strongly believe that adding `DTMFCode`in `DTMFCode` class doesn't carry any information :)

module-services/service-cellular/DTMFCode.hpp
```diff
@@ -0,0 +1,19 @@
...
+
+#pragma once
+#include <stdint.h>
+#include <string>
+
+class DTMFCode
+{
+  public:
+    explicit DTMFCode(char digitChar);
+    explicit DTMFCode(const char *digitString);
```
pholat, line 12: Please add doxy docks how these work :) ( is `char` ascii char, or integer value - it would be nice to tell here :) )

module-services/service-cellular/ServiceCellular.cpp
```diff
@@ -1468,17 +1468,16 @@ std::string ServiceCellular::GetScanMode(void)
...
     at::Result resp;
     if (channel) {
         auto command           = at::factory(at::AT::QLDTMF);
-        std::string dtmfString = "\"" + utils::singleDigitToString(digit) + "\"";
-        resp                   = channel->cmd(command.getCmd() + dtmfString);
+        resp                   = channel->cmd(command.getCmd() + code.getDTMFCodeString());
         if (resp) {
             command = at::factory(at::AT::VTS);
-            resp    = channel->cmd(command.getCmd() + dtmfString);
+            resp    = channel->cmd(command.getCmd() + code.getDTMFCodeString());
```
pholat, line 1480: I believe that with `operator str()` this line could be not changed :)

pholat (commented):
I would kindly ask for minor changes then I'll gladly approve :)

module-services/service-cellular/DTMFCode.hpp
SP2FET: imo that was the problem, that uint32_t has been used to transfer ASCII code instead of built-in type - `char`. Now it's much more explicit that it has to be digit CHAR, not the digit itself, but sure, will add

module-services/service-cellular/DTMFCode.cpp
SP2FET: sure, but wouldn't the trace show the file either way?

module-services/service-cellular/ServiceCellular.cpp
SP2FET: sure, that's a nice idea

module-services/service-cellular/DTMFCode.hpp
SP2FET: added string cast operator

module-services/service-cellular/DTMFCode.cpp
SP2FET: I think that this is more clear this way - I'd use user literal when doing e.g. units conversions etc

module-services/service-cellular/DTMFCode.cpp
SP2FET: added

module-services/service-cellular/DTMFCode.cpp
SP2FET: fixed

module-services/service-cellular/DTMFCode.hpp
SP2FET: added

Lefucjusz (approved):

module-services/service-cellular/DTMFCode.cpp
pholat: I wrote because it's a conversion :) literal just state type - nothing more tbh

pholat (approved):

module-bluetooth/Bluetooth/interface/profiles/HFP/HFP.cpp
```diff
@@ -319,7 +319,12 @@ namespace bluetooth
...
             auto digitStr = hfp_subevent_transmit_dtmf_codes_get_dtmf(event);
             LOG_DEBUG("Send DTMF Codes: '%s'\n", digitStr);
-            cellularInterface->sendDTMFCode(const_cast<sys::Service *>(ownerService), utils::toNumeric(digitStr));
+            try {
+                cellularInterface->sendDTMFCode(const_cast<sys::Service *>(ownerService), DTMFCode(digitStr));
+            }
+            catch (std::out_of_range &e) {
+                LOG_ERROR("Can't send DTMF code for digit: %s", digitStr);
+            }
             hfp_ag_send_dtmf_code_done(aclHandle);
```
paweljonskim, line 328: should this line be outside of `try`?

paweljonskim (approved):

module-bluetooth/Bluetooth/interface/profiles/HFP/HFP.cpp
SP2FET: nope - BT state has to be updated, otherwise BT state machine could be hanged up

PR #3802, opened 2022-05-02
[MOS-425] Crash fix on unchecked nullptr

pholat:
Crash fix 

Make sure that this PR:

### DIFF
--- a/module-services/service-db/ServiceDBCommon.cpp
+++ b/module-services/service-db/ServiceDBCommon.cpp
@@ -35,10 +35,18 @@ sys::MessagePointer ServiceDBCommon::DataReceivedHandler(sys::DataMessage *msgl,
         auto query     = msg->getQuery();
         auto queryType = query->type;
         auto result    = interface->runQuery(std::move(query));
-        auto recordID  = result->getRecordID().value_or(0);
-        responseMsg    = std::make_shared<db::QueryResponse>(std::move(result));
+        std::optional<uint32_t> id;
+        if (result != nullptr) {
+            id = result->getRecordID();
+        }
+        else {
+            LOG_WARN("There is no response associated with query: %s!", query ? query->debugInfo().c_str() : "");
+        }
+        responseMsg = std::make_shared<db::QueryResponse>(std::move(result));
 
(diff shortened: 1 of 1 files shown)


Maciej-Mudita (approved):

SP2FET (approved):

module-services/service-db/ServiceDBCommon.cpp
```diff
@@ -35,10 +35,18 @@ sys::MessagePointer ServiceDBCommon::DataReceivedHandler(sys::DataMessage *msgl,
...
-        auto recordID  = result->getRecordID().value_or(0);
-        responseMsg    = std::make_shared<db::QueryResponse>(std::move(result));
+        std::optional<uint32_t> val;
+        if (result != nullptr) {
+            val = result->getRecordID();
+        }
+        else {
+            LOG_WARN("There is no response associated with query: %s!", query ? query->debugInfo().c_str() : "");
+        }
+        responseMsg = std::make_shared<db::QueryResponse>(std::move(result));
```
swichu91, line 45: Do we want to pass `nullptr` to `db::QueryResponse`?

module-services/service-db/ServiceDBCommon.cpp
```diff
@@ -35,10 +35,18 @@ sys::MessagePointer ServiceDBCommon::DataReceivedHandler(sys::DataMessage *msgl,
...
 
-        sendUpdateNotification(msg->getInterface(), queryType, recordID);
+        if (val) {
+            sendUpdateNotification(msg->getInterface(), queryType, val.value());
```
swichu91, line 48: Calling `val.value()` looks a bit strange to me. How about replacing it with `*val` or changing the name of the `val` variable to something more adequate?

module-services/service-db/ServiceDBCommon.cpp
pholat: Yes, I didn't want to change logic here while introducing the fix..

module-services/service-db/ServiceDBCommon.cpp
pholat: That's true. I'll rename it?

swichu91 (approved):

PR #3806, opened 2022-05-05
[MOS-402] Time legibility on lock screen fix

Lefucjusz:
**Description**

Fix of the issue that time on the lock screen was
getting illegible after a few hours. This change
fixes the problem by performing deep refresh
instead of fast refresh every 30 minutes. The
refresh rate can be easily changed, as it is stored
in settings database.

Make sure that this PR:

### DIFF
--- a/module-apps/apps-common/ApplicationCommon.cpp
+++ b/module-apps/apps-common/ApplicationCommon.cpp
@@ -467,13 +467,9 @@ namespace app
 
     sys::MessagePointer ApplicationCommon::handleMinuteUpdated(sys::Message *msgl)
     {
-        if (state == State::ACTIVE_FORGROUND && getCurrentWindow()->updateTime()) {
-
-            if (isOnPhoneLockWindow()) {
-                updateStatusBarOnPhoneLockWindow();
-            }
-
-            refreshWindow(gui::RefreshModes::GUI_REFRESH_FAST);
+        if (state == State::ACTIVE_FORGROUND) {
+            auto requestedRefreshMode = getCurrentWindow()->updateTime();
+            refreshWindow(requestedRefreshMode);
         }
         return sys::msgHandled();
     }
@@ -1049,13 +1045,5 @@ namespace app
     bool ApplicationCommon::isOnPhoneLockWindow()
     {
         return getCurrentWindow()->getName() == gui::popup::window::phone_lock_window;
-    }
-
-    void ApplicationCommon::updateStatusBarOnPhoneLockWindow()
-    {
-        getCurrentWindow()->updateSignalStrength();
-        getCurrentWindow()->updateNetworkAccessTechnology();
-        getCurrentWindow()->updateBatteryStatus();
-        getCurrentWindow()->updateSim();
     }
 } /* namespace app */
--- a/module-apps/apps-common/popups/lock-popups/PhoneLockedWindow.hpp
+++ b/module-apps/apps-common/popups/lock-popups/PhoneLockedWindow.hpp
@@ -18,7 +18,8 @@ namespace gui
       public:
         PhoneLockedWindow(app::ApplicationCommon *app,
                           const std::string &name,
-                          std::unique_ptr<WallpaperPresenter> &&presenter);
+                          std::unique_ptr<WallpaperPresenter> &&presenter,
+                          unsigned lockScreenDeepRefreshRate);
 
         bool onInput(const InputEvent &inputEvent) override;
         void onBeforeShow(ShowMode mode, SwitchData *data) override;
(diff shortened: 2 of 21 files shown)


module-apps/apps-common/ApplicationCommon.cpp
```diff
@@ -1058,4 +1060,37 @@ namespace app
...
+        }
+        else {
+            LOG_DEBUG("Requesting fast refresh on phone lock screen");
+            refreshWindow(gui::RefreshModes::GUI_REFRESH_FAST);
+        }
+    }
+
+    void ApplicationCommon::initializeDeepRefreshCounter()
+    {
+        constexpr auto fallbackRefreshRate = 15;
```
SP2FET, line 1080: why fallback is different than the DB value?
consider using `std::chrono`'s literals

SP2FET (approved):

Jenkinsfile
```diff
@@ -13,7 +13,7 @@ pipeline {
     }
     environment {
         JOBS=30
-        PATH="/usr/local/cmake-3.21.3-linux-x86_64/bin:/usr/local/gcc-arm-none-eabi-10-2020-q4-major/bin:$PATH"
+        PATH="/usr/local/cmake-3.21.3-linux-x86_64/bin:/usr/local/gcc-arm-11.2-2022.02-x86_64-arm-none-eabi/bin:$PATH"
```
pholat, line 16: this shouldn't be here?

module-apps/apps-common/ApplicationCommon.cpp
```diff
@@ -468,12 +468,13 @@ namespace app
     sys::MessagePointer ApplicationCommon::handleMinuteUpdated(sys::Message *msgl)
     {
         if (state == State::ACTIVE_FORGROUND && getCurrentWindow()->updateTime()) {
-
             if (isOnPhoneLockWindow()) {
                 updateStatusBarOnPhoneLockWindow();
+                refreshTimeOnPhoneLockWindow();
```
pholat, line 473: why not:
- override handleMinuteUpdated in the PhoneLockWindow
- change return type there
It would remove Window specific code fully from the ApplicationComomn

module-apps/apps-common/ApplicationCommon.cpp
```diff
@@ -1058,4 +1060,37 @@ namespace app
...
+        if (lockScreenRefreshCounter.currentValue >= lockScreenRefreshCounter.referenceValue) {
+            lockScreenRefreshCounter.currentValue = 0;
+            LOG_DEBUG("Requesting deep refresh on phone lock screen");
+            refreshWindow(gui::RefreshModes::GUI_REFRESH_DEEP);
+        }
+        else {
+            LOG_DEBUG("Requesting fast refresh on phone lock screen");
+            refreshWindow(gui::RefreshModes::GUI_REFRESH_FAST);
+        }
+    }
```
pholat, line 1076: move check to the struct  and handle refresh separately? similar below

module-apps/apps-common/popups/lock-popups/PhoneLockedWindow.hpp
```diff
@@ -18,24 +18,52 @@ namespace gui
...
 
         std::unique_ptr<WallpaperPresenter> wallpaperPresenter;
         bool refreshedOnPhoneLockTimeLock = false;
+
+        void initializeDeepRefreshCounter(std::uint8_t deepRefreshRate);
+
+        struct
+        {
+            std::uint8_t currentValue;
+            std::uint8_t referenceValue;
```
pholat, line 43: why not plain `unsigned int` ? It's not like we care about integer precision here ;) 

small nitpick - do whatever you want with this comment :bowing_man:

module-apps/apps-common/ApplicationCommon.cpp
```diff
@@ -467,13 +467,13 @@ namespace app
 
     sys::MessagePointer ApplicationCommon::handleMinuteUpdated(sys::Message *msgl)
     {
-        if (state == State::ACTIVE_FORGROUND && getCurrentWindow()->updateTime()) {
-
+        if (state == State::ACTIVE_FORGROUND) {
+            auto requestedRefreshMode = getCurrentWindow()->updateTime();
             if (isOnPhoneLockWindow()) {
                 updateStatusBarOnPhoneLockWindow();
             }
```
pholat, line 474: didn't we agree that it can be moved to the `PhoneLockWindow` and both these methods can be removed from the ApplicationCommon? :) It's not much, but when we can remove unrelated code from "common" parts - it's worth it ;)

pholat (approved):
Please remove the code that is not required and can be solved with existing inheritance, then close the comments and merge. Good job :beer:

module-apps/apps-common/ApplicationCommon.cpp
Lefucjusz: My mistake, fixed now.
As for literals, I don't think using them here is required. I'm only counting calls of this function to count the time, not using some "real" timer ;)

Jenkinsfile
Lefucjusz: True, it was added to test some CI-related issue, removed it now

module-apps/apps-common/popups/lock-popups/PhoneLockedWindow.hpp
Lefucjusz: Just a habit from coding for smaller, 8-bit CPUs :P Changed to unsigned.

module-apps/apps-common/ApplicationCommon.cpp
Lefucjusz: isOnPhoneLockWindow() is used in multiple different places, so I left this method, but updateStatus... content was moved to PhoneLockedWindow as you suggested ;)

pholat (approved):

module-apps/apps-common/ApplicationCommon.cpp
pholat: tx, I'll see if it could be possibly removed in other places too :)

PR #3819, opened 2022-05-09
[MOS-459] Wait with GUI change till RING delay 

pholat:
Normally the incomming call works like that:
- RING indicates that somebody calls us
- CLIP carries caller ID **and is optional** 
Therefore I've added an additional 1sec delay till we handle RING awaiting the CLIP - which usually comes just after RING.

The issue was: if we managed to show RING before CLIP for some reason: be it late CLIP, or faster/slower code - then we had `private number` - because for private numbers you do not get `CLIP`

Make sure that this PR:


mjan-dev (approved):
LGTM

mjan-dev (commented):
Please fix PR name before merging.

PR #3822, opened 2022-05-10
[MOS-453] Fix HF on Low Battery Notification

Maciej-Mudita:
**Description**

Fix in action handling in ApplicationManager

Make sure that this PR:

### DIFF
--- a/module-services/service-appmgr/model/ApplicationManagerCommon.cpp
+++ b/module-services/service-appmgr/model/ApplicationManagerCommon.cpp
@@ -560,8 +560,9 @@ namespace app::manager
                                        (actionParams && actionParams->disableAppClose));
         SwitchRequest switchRequest(
             service::name::appmgr, targetApp->name(), targetApp->switchWindow, std::move(targetApp->switchData));
-        return handleSwitchApplication(&switchRequest, focusedAppClose) ? ActionProcessStatus::Accepted
-                                                                        : ActionProcessStatus::Skipped;
+        handleSwitchApplication(&switchRequest, focusedAppClose);
+
+        return ActionProcessStatus::Skipped;
     }
 
     auto ApplicationManagerCommon::handleCustomActionOnBackgroundApp(ApplicationHandle *app, ActionEntry &action)
@@ -749,6 +750,7 @@ namespace app::manager
         app.startupReason = StartupReason::Launch;
 
         if (!actionsRegistry.hasPendingAction()) {
+            actionsRegistry.process();
             return;
         }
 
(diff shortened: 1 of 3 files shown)


module-services/service-appmgr/model/ApplicationManagerCommon.cpp
```diff
@@ -749,6 +749,7 @@ namespace app::manager
         app.startupReason = StartupReason::Launch;
 
         if (!actionsRegistry.hasPendingAction()) {
```
mjan-dev, line 751: basically !isCurrentlyProcessing. It would be more explicit.

pawel-mudita (approved):

mjan-dev (approved):

module-services/service-appmgr/model/ApplicationManagerCommon.cpp
```diff
@@ -560,8 +560,9 @@ namespace app::manager
                                        (actionParams && actionParams->disableAppClose));
         SwitchRequest switchRequest(
             service::name::appmgr, targetApp->name(), targetApp->switchWindow, std::move(targetApp->switchData));
-        return handleSwitchApplication(&switchRequest, focusedAppClose) ? ActionProcessStatus::Accepted
-                                                                        : ActionProcessStatus::Skipped;
+        handleSwitchApplication(&switchRequest, focusedAppClose);
```
pholat, line 563: why do we ignore app switch result now?

pholat (commented):
There are some UT for this, arent there? ( just for actions) if so - would you be so kind to add this flow to it?

Other than that - seems ok.

module-services/service-appmgr/model/ApplicationManagerCommon.cpp
Maciej-Mudita: first we check if the application is `ACTIVE_FORGROUND` or `ACTIVE_BACKGROUND` and then we give `ActionProcessStatus::Accepted`. My analysis shows that when we call application switching via `handleSwitchApplication`, we cannot give `ActionProcessStatus::Accepted` because the action will be called too soon - before the application starts.

Maciej-Mudita:
@pholat there is already a UT for the returned `ActionProcessStatus::Skipped` status in test-ActionsRegistry.cpp

module-services/service-appmgr/model/ApplicationManagerCommon.cpp
pholat: ok, it might be nice to add state, but can be cumbersone at the moment

PR #3825, opened 2022-05-11
[BH-1423] Enable tickless option

swichu91:
**Description**

Made required changes to be able to compile
freeRTOS using tickless idle.

Make sure that this PR:

### DIFF
--- a/module-bsp/board/rt1051/os/fsl_tickless_gpt.c
+++ b/module-bsp/board/rt1051/os/fsl_tickless_gpt.c
@@ -11,8 +11,9 @@
 #include "task.h"
 #include "fsl_tickless_generic.h"
 
-#if configUSE_TICKLESS_IDLE == 1
+#if configUSE_TICKLESS_IDLE == 2
 #include "fsl_gpt.h"
+#include "fsl_gpc.h"
 #else
 #include "fsl_common.h"
 #endif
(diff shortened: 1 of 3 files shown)


module-bsp/board/rt1051/os/fsl_tickless_gpt.c
```diff
@@ -11,8 +11,9 @@
 #include "task.h"
 #include "fsl_tickless_generic.h"
 
-#if configUSE_TICKLESS_IDLE == 1
+#if configUSE_TICKLESS_IDLE == 2
```
paweljonskim, line 14: why change from 1 to 2? is it changed everywhere?

paweljonskim (approved):

mjan-dev (approved):

module-bsp/board/rt1051/os/fsl_tickless_gpt.c
swichu91: It is described here: https://www.freertos.org/low-power-ARM-cortex-rtos.html 
Basically, set it to 2 if you want to use custom `vPortSuppressTicksAndSleep` and `vPortSetupTimerInterrupt`. It could be simply overwritten as these symbols are defined weak. Unfortunately, there is no way to overwrite weak symbols defined in different translation unit.

PR #3843, opened 2022-05-17
[BH-1489] Prevent popups from clearing settings screen

paweljonskim:
Prevent popups from clearing settings screen
Popups default switch reason is set to Popup instead of PhoneLock

**Description**

Make sure that this PR:

### DIFF
--- a/module-apps/apps-common/ApplicationCommon.hpp
+++ b/module-apps/apps-common/ApplicationCommon.hpp
@@ -210,10 +210,6 @@ namespace app
 
         std::list<std::unique_ptr<app::GuiTimer>> gui_timers;
         std::unordered_map<manager::actions::ActionId, OnActionReceived> receivers;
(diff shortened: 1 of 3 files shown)


pawel-mudita (approved):

module-apps/apps-common/ApplicationCommon.hpp
```diff
@@ -362,6 +358,10 @@ namespace app
         virtual void clearPendingPopups();
         virtual bool tryShowPopup();
         void abortPopup(gui::popup::ID id);
+        void switchWindowPopup(const std::string &windowName,
```
pholat, line 361: it doesn't have to be moved to public space, does it?

pholat: My bad, it protected was folded :)

pholat (approved):

PR #3849, opened 2022-05-19
[BH-1490] Prevent double deeppress from causing back behaviour

paweljonskim:
Prevent double deepress from causing back behaviour
By eliminating race in AbortPopup handling

**Description**

Make sure that this PR:


products/BellHybrid/apps/common/src/popups/AlarmActivatedWindow.cpp
```diff
@@ -62,10 +62,7 @@ namespace gui
 
     void AlarmActivatedWindow::returnToPreviousWindow()
     {
-        app::manager::Controller::sendAction(
-            application,
-            app::manager::actions::AbortPopup,
-            std::make_unique<gui::PopupRequestParams>(gui::popup::ID::AlarmDeactivated));
+        application->windowsStack().pop(popup::window::alarm_deactivated_window);
```
pholat, line 65: if we want to ignore window on stack - shouldn't we enter it with `ignoreWindowOnStack` option?

paweljonskim: I admit I didnt know  the `ignoreCurrentWindowOnStack` option. The thing is, we dont want to ignore ANY window on stack, just one, specific (alarm deactivated window in this case).

The idea is that if user pressed deep press multiple times (causing multiple activated/deactivated popups showing), only the latest will be presented to used and all the old ones will be deleted, so that after timeout user is back on the screen he pressed deeppress for the first time.

Do you see if that can be achieved by `ignoreCurrentWindowOnStack`?

pholat: It's ok then

pholat (approved):

mjan-dev (approved):

PR #3856, opened 2022-05-23
[MOS-473] Long time to refresh network coverage fix

Lefucjusz:
**Description**

Fix of the issue that after the system has started
the network coverage wasn't updated for a long time
on the lock screen.

Make sure that this PR:


pholat (approved):

module-gui/gui/widgets/StatusBar.hpp
```diff
@@ -221,7 +221,7 @@ namespace gui::status_bar
 
         /// Show/hide battery status widget
         /// @param enabled true to show false to hide the widget
-        void showBattery(bool enabled);
+        bool showBattery(bool enabled);
```
SP2FET, line 224: imo it would be worth to update the oxygen when you changed the function ;)

SP2FET (commented):
I quite don't get how battery level handling affects network refreshes to be honest :P

SP2FET (approved):
imo the whole battery singleton could be rewritten, but it's good enough right now

PR #3870, opened 2022-05-25
[MOS-494] PIN settings infinite loop fix

Lefucjusz:
**Description**

Fix of the bug that after selecting SIM slot
in which there's no card and trying to
enter PIN settings menu phone was getting
stuck in infinite looping between
PIN settings window and 'Cannot connect
to SIM card' popup.

Make sure that this PR:

### DIFF
--- a/module-apps/application-settings/windows/network/SimPINSettingsWindow.hpp
+++ b/module-apps/application-settings/windows/network/SimPINSettingsWindow.hpp
@@ -4,20 +4,24 @@
 #pragma once
 
 #include <application-settings/windows/BaseSettingsWindow.hpp>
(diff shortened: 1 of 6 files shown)


module-apps/application-settings/windows/network/SimPINSettingsWindow.hpp
```diff
@@ -18,6 +18,7 @@ namespace gui
         void changePinState(bool &currentState);
 
         bool pinIsOn = false;
+        bool simIsPresent                   = false;
```
pholat, line 21: can't we check is somehow better? i.e. from the hardware state?

pholat (changes requested):
doestn't seem right

SP2FET (approved):

PR #3885, opened 2022-06-01
[MOS-463] Fix no dialing state during BT turned on

SP2FET:
**Description**

When BT was connected (or just turned on) there was no dialing state -
the call went directly to ongoing call. Moreover, fixed no ringing
sound during ringing state - there was a bug in BT stack.

Make sure that this PR:

### DIFF
--- a/module-services/service-cellular/call/tests/test-CallMachine.cpp
+++ b/module-services/service-cellular/call/tests/test-CallMachine.cpp
@@ -41,6 +41,8 @@ namespace mocks
         fakeit::When(Method(gui, notifyCallEnded)).AlwaysReturn();
         fakeit::When(Method(gui, notifyCallActive)).AlwaysReturn();
         fakeit::When(Method(gui, notifyCallDurationUpdate)).AlwaysReturn();
+        fakeit::When(Method(gui, notifyOutgoingCallAnswered)).AlwaysReturn();
+
         return gui;
     }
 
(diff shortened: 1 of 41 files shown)


Lefucjusz (approved):

Maciej-Mudita (approved):

module-services/service-cellular/call/tests/test-CallMachine.cpp
```diff
@@ -362,6 +364,7 @@ TEST_CASE("call outgoing - answered")
     REQUIRE(machine->machine.process_event(call::event::StartCall{CallType::CT_OUTGOING, number.getView()}));
     REQUIRE(machine->machine.process_event(call::event::Answer{}));
     fakeit::Verify(Method(di.audio, play)).Exactly(0);
+    fakeit::Verify(Method(di.gui, notifyOutgoingCallAnswered)).Exactly(1);
```
pholat, line 367: it's already propagating:
```
            di.gui->notifyCallStarted(call.record.phoneNumber, call.record.type);
```
why do we have to have another state here? if we do - can't we remove its notification in HandleCallStart? So that it would be handled in a similar manner in UI and in Bluetooth?

Also: for some reason, we have multicast in GUI and in multicast.
By design - GUI was for actions ( first two calls in GUI dependency) then we have multicast. This might introduce redundancies - can you check it?

SP2FET: call started is called for both outgoing and incoming calls - that was the case why we needed the refactor TBH :)

For the placement - I just went with the scheme that was already in the code

PR #3895, opened 2022-06-03
[MOS-516] FIX: tethering while call disabled

pholat:
We discard tethering while the ongoing call now


kkleczkowski (approved):

products/PurePhone/sys/SystemManager.cpp
```diff
@@ -36,7 +36,13 @@ namespace sys
             return (Store::GSM::get()->selected == Store::GSM::SIM::SIM1 ||
                     Store::GSM::get()->selected == Store::GSM::SIM::SIM2);
         };
-        phoneModeSubject = std::make_unique<phone_modes::Subject>(this, simSelected);
+        auto isCallOngoing = [&]() {
```
swichu91, line 39: Please capture only what you use :)

pholat: yass :) done

PR #3915, opened 2022-06-13
[BH-1522] Meditation - new circular progress layout

paweljonskim:
Rework meditation progress layout

**Description**

Make sure that this PR:

### DIFF
--- a/module-apps/apps-common/widgets/ProgressTimer.cpp
+++ b/module-apps/apps-common/widgets/ProgressTimer.cpp
@@ -54,7 +54,7 @@ namespace app
 
     auto ProgressTimer::onTimerTimeout(sys::Timer &task) -> bool
     {
-        ++elapsed;
+        elapsed += baseTickInterval;
         update();
         if (isStopped() || isFinished()) {
             task.stop();
@@ -82,12 +82,17 @@ namespace app
 
     auto ProgressTimer::intervalReached() const noexcept -> bool
     {
-        return hasInterval && (elapsed.count() % interval.count()) == 0;
+        return hasInterval &&
+               (std::chrono::duration_cast<std::chrono::seconds>(elapsed).count() % interval.count()) == 0;
     }
 
     void ProgressTimer::stop()
     {
         isRunning = false;
+    }
+    std::chrono::milliseconds ProgressTimer::getElapsed()
+    {
+        return elapsed;
     }
 
     void ProgressTimer::registerOnFinishedCallback(std::function<void()> cb)
--- a/module-apps/apps-common/widgets/ProgressTimerWithBarGraphAndCounter.cpp
+++ b/module-apps/apps-common/widgets/ProgressTimerWithBarGraphAndCounter.cpp
@@ -27,7 +27,7 @@ namespace app
         if (text == nullptr) {
             return;
         }
-        const auto secondsRemaining = duration - elapsed;
+        const auto secondsRemaining = duration - std::chrono::duration_cast<std::chrono::seconds>(elapsed);
         const Duration remainingDuration{std::time_t{secondsRemaining.count()}};
         UTF8 timerText;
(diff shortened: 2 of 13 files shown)


module-apps/apps-common/widgets/ProgressTimer.cpp
```diff
@@ -82,12 +82,14 @@ namespace app
 
     auto ProgressTimer::intervalReached() const noexcept -> bool
     {
-        return hasInterval && (elapsed.count() % interval.count()) == 0;
+        return hasInterval &&
+               (std::chrono::duration_cast<std::chrono::seconds>(elapsed).count() % interval.count()) == 0;
     }
 
-    void ProgressTimer::stop()
+    std::chrono::milliseconds ProgressTimer::stop()
```
pholat, line 89: consider adding function `elapsed` instead? which will always return the current value?

products/BellHybrid/apps/application-bell-meditation-timer/presenter/MeditationProgressPresenter.cpp
```diff
@@ -81,10 +82,12 @@ namespace app::meditation
 
     void MeditationProgressPresenter::finish()
     {
-        timer->stop();
+        const auto elapsed = std::chrono::duration_cast<std::chrono::minutes>(timer->stop());
+        auto summaryText   = utils::translate("app_meditation_summary") + std::to_string(elapsed.count()) + " " +
```
swichu91, line 86:
```suggestion
        const auto summaryText   = utils::translate("app_meditation_summary") + std::to_string(elapsed.count()) + " " +
```

module-apps/apps-common/widgets/ProgressTimerWithBarGraphAndCounter.hpp
```diff
@@ -25,6 +25,8 @@ namespace app
         void update() override;
         void updateText();
         void updateProgress();
+        void reset(std::chrono::seconds _duration,
+                   std::chrono::seconds _interval = std::chrono::seconds::zero()) override;
```
swichu91, line 29:
```suggestion
                   std::chrono::seconds _interval = {}) override;
```

module-apps/apps-common/widgets/ProgressTimerWithBarGraphAndCounter.cpp
```diff
@@ -27,7 +27,7 @@ namespace app
         if (text == nullptr) {
             return;
         }
-        const auto secondsRemaining = duration - elapsed;
+        const auto secondsRemaining = duration - std::chrono::duration_cast<std::chrono::seconds>(elapsed);
```
swichu91, line 30: What will the result of such an operation be? Are we ok with rounding milliseconds down? For instance, 999ms will be converted to 0 seconds.

paweljonskim: I think its fine here. We are just deciding how many seconds to show to user when there is 1,999ms remaining. We can choose rounding up or down, I dont think it realy matters here. Especially since in this particular scenario we have 1s intervals, so it shouldnt be a problem.
Let me know if you think using rounding function will make this more clear to user, I have really no preference.

module-gui/gui/widgets/ProgressBar.cpp
```diff
@@ -149,4 +149,123 @@ namespace gui
...
+        };
+        createWidgets();
+        updateDrawArea();
+    }
+
+    void ArcProgressBar::createWidgets()
+    {
+        const auto progressArcRadius       = radius + 3;
+        const auto progressArcWidth        = penWidth + 7;
+        const auto progressIndicatorRadius = (progressArcWidth - 1) / 2;
```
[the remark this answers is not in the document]
paweljonskim: added a comment

PR #3951, opened 2022-07-05
[BH-1509] Frontlight wont flash on snooze

paweljonskim:
Frontlight wont flash on snooze when alarm light is off
and automatic frontlight mode is on

**Description**

Make sure that this PR:


disonek (approved):

swichu91 (approved):
It's not the most elegant solution, but I believe it works. I can't think of different solution that is more elegant and can be implemented in the same amount of time.

PR #3969, opened 2022-07-18
[MOS-589] Fix torch driver

dawidwojtas-mudita:
Perform turn-off torch in only one place.
Before writing i2c make sure the torch is power supplied.
Minor code refactor.

**Description**

Make sure that this PR:

### DIFF
--- a/module-bsp/board/rt1051/bsp/torch/torch.cpp
+++ b/module-bsp/board/rt1051/bsp/torch/torch.cpp
@@ -19,7 +19,8 @@ namespace bsp
 
     namespace torch
     {
-        static xQueueHandle qHandleIrq = NULL;
+        static xQueueHandle qHandleIrq = nullptr;
+        static constexpr auto delayMs  = 5;
 
         static I2CAddress addr = {.deviceAddress = 0x63, .subAddress = 0, .subAddressSize = 1};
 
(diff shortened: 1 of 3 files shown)


module-bsp/board/rt1051/bsp/torch/torch.cpp
```diff
@@ -42,11 +53,11 @@ namespace bsp
...
-                                              .pin      = static_cast<uint32_t>(BoardDefinitions::TORCH_DRIVER_EN)});
-            gpio->WritePin(static_cast<uint32_t>(BoardDefinitions::TORCH_DRIVER_EN), 0);
-            vTaskDelay(pdMS_TO_TICKS(5));
-            gpio->WritePin(static_cast<uint32_t>(BoardDefinitions::TORCH_DRIVER_EN), 1);
-            vTaskDelay(pdMS_TO_TICKS(5));
+                                              .pin = static_cast<std::uint32_t>(BoardDefinitions::TORCH_DRIVER_EN)});
+            powerOffTorch();
+            vTaskDelay(pdMS_TO_TICKS(delayTicks));
+            powerOnTorch();
+            vTaskDelay(pdMS_TO_TICKS(delayTicks));
```
SP2FET, line 60: those are not ticks but milliseconds, so please rename the variable ;)

SP2FET (approved):

module-bsp/board/rt1051/bsp/torch/torch.cpp
dawidwojtas-mudita: You're right, thanks :)

module-bsp/board/rt1051/bsp/torch/torch.cpp
```diff
@@ -56,17 +67,17 @@ namespace bsp
...
 
         bool isPresent(void)
         {
             al3644tt_device_id id;
             addr.subAddress = AL3644TT_DEVICE_ID_REG;
-            auto read       = i2c->Read(addr, (uint8_t *)(&id), 1);
+            auto read       = i2c->Read(addr, (std::uint8_t *)(&id), sizeof(al3644tt_device_id));
 
-            if (read != 1) {
+            if (read != sizeof(std::uint8_t)) {
```
awulkiew, line 80: `al3644tt_device_id` is a struct containing two `uint8_t` members. It is casted to `uint8_t`. So the number and order of member variables in memory has to be the same as data sent on I2C? This is error prone. What if someone adds a member there, then the number passed here is going to be different. It is not guaranteed that if this struct is changed then the I2C message will change as well. The message should be hardcoded as it was before or depend on some other settings, not on struct's size.

In Read `sizeof` the struct is passed which is 2 and below `sizeof` `uint8_t` is passed which is 1. Is that correct?

`uint8_t` is optional type and is not going to be defined on all platforms. I2C library in such cases would probably take `unsigned char` which not neccesarily would be 8-bit type.

What is the reason behind replacing hardcoded numbers with `sizeof`? Sizeof doesn't return the number of 8-bit bytes but the number of bytes of the size of `CHAR_BIT`. I2C message's size is not bound in any way by the platform. Why does the code suggest that it is?

dawidwojtas-mudita: Based on your clues I've changed the implementation. Now it uses proper structs and structs names. Our system has `uint8_t` so it is safe for i2c communication in this case.

Lefucjusz (approved):

awulkiew (approved):

PR #3989, opened 2022-07-27
[MOS-613] Fix keyboard doesn't respond

dawidwojtas-mudita:
Now the keyboard driver is checking the result of i2c
transmission and we get an error code if something fails.

Make sure that this PR:

### DIFF
--- a/module-bsp/board/rt1051/drivers/RT1051DriverI2C.cpp
+++ b/module-bsp/board/rt1051/drivers/RT1051DriverI2C.cpp
@@ -3,10 +3,54 @@
 
 #include "RT1051DriverI2C.hpp"
 #include <log/log.hpp>
-#include "../board.h"
+#include "board.h"
 
 namespace drivers
 {
+    namespace
+    {
+        ssize_t translateErrorCode(const status_t errorStatus)
+        {
+            I2CErrorCodes code;
+
+            switch (errorStatus) {
+            case kStatus_LPI2C_Busy:
+                code = I2CErrorCodes::busy;
+                break;
+            case kStatus_LPI2C_Idle:
+                code = I2CErrorCodes::idle;
+                break;
+            case kStatus_LPI2C_Nak:
+                code = I2CErrorCodes::nak;
+                break;
+            case kStatus_LPI2C_FifoError:
+                code = I2CErrorCodes::fifo;
+                break;
+            case kStatus_LPI2C_BitError:
+                code = I2CErrorCodes::bit;
+                break;
+            case kStatus_LPI2C_ArbitrationLost:
+                code = I2CErrorCodes::arbitrationLost;
+                break;
+            case kStatus_LPI2C_PinLowTimeout:
+                code = I2CErrorCodes::pinLowTimeout;
+                break;
+            case kStatus_LPI2C_NoTransferInProgress:
+                code = I2CErrorCodes::noTransferInProgress;
+                break;
+            case kStatus_LPI2C_DmaRequestFail:
+                code = I2CErrorCodes::dmaRequestFail;
+                break;
+            case kStatus_LPI2C_Timeout:
+                code = I2CErrorCodes::timeout;
+                break;
+            default:
+                code = I2CErrorCodes::unknown;
+                break;
+            }
+            return static_cast<ssize_t>(code);
+        }
+    } // namespace
 
     RT1051DriverI2C::RT1051DriverI2C(const I2CInstances inst, const DriverI2CParams &params) : DriverI2C(params, inst)
     {
@@ -64,7 +108,7 @@ namespace drivers
         auto ret = BOARD_LPI2C_Send(
             base, addr.deviceAddress, addr.subAddress, addr.subAddressSize, const_cast<uint8_t *>(txBuff), size);
         if (ret != kStatus_Success) {
-            return -1; // TODO:M.P: fix me
+            return translateErrorCode(ret);
         }
         else {
             return size;
@@ -76,7 +120,7 @@ namespace drivers
         cpp_freertos::LockGuard lock(mutex);
         auto ret = BOARD_LPI2C_Receive(base, addr.deviceAddress, addr.subAddress, addr.subAddressSize, rxBuff, size);
         if (ret != kStatus_Success) {
-            return -1; // TODO:M.P: fix me
+            return translateErrorCode(ret);
         }
         else {
             return size;
@@ -104,7 +148,7 @@ namespace drivers
         ret |= BOARD_LPI2C_Send(base, addr.deviceAddress, addr.subAddress, addr.subAddressSize, (uint8_t *)&rx, size);
 
         if (ret != kStatus_Success) {
-            return -1; // TODO:M.P: fix me
+            return translateErrorCode(ret);
         }
         else {
             return size;
--- a/module-bsp/board/rt1051/puretx/bsp/keyboard/keyboard.cpp
+++ b/module-bsp/board/rt1051/puretx/bsp/keyboard/keyboard.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2021, Mudita Sp. z.o.o. All rights reserved.
+// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
 // For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 #include "TCA8418.hpp"
@@ -10,7 +10,7 @@
 #include <bsp/keyboard/keyboard.hpp>
 #include <board/BoardDefinitions.hpp>
 #include <board.h>
-#include <fsl_common.h>
+#include <log/log.hpp>
 
 #include <stdio.h>
 #include <stdint.h>
@@ -29,7 +29,17 @@ namespace bsp::keyboard
             rightFnRelease = 0x04,
         };
 
+        enum class KeyboardPower
+        {
+            off = 0,
+            on
+        };
+
         constexpr auto keyboardContactOscillationTimeoutMS = 20;
+        constexpr auto maxTransmissionRetry                = 3;
+        constexpr auto chipResetMs                         = 1;
+        constexpr auto chipRecoveryMs                      = 1;
+        constexpr int retryMs[maxTransmissionRetry]        = {1, 3, 7};
 
         std::shared_ptr<DriverI2C> i2c;
         I2CAddress i2cAddr = {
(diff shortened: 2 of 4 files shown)


module-bsp/drivers/i2c/DriverI2C.hpp
```diff
@@ -49,6 +51,8 @@ namespace drivers
 
         virtual ssize_t Modify(const I2CAddress &addr, const uint32_t mask, bool setClr, const size_t size) = 0;
 
+        virtual status_t GetLastError() = 0;
```
marziel, line 54: This may be non-virtual as well, because returning the last error probably won't have to be a customizable operation

marziel (approved):

module-bsp/board/rt1051/puretx/bsp/keyboard/keyboard.cpp
```diff
@@ -67,24 +68,47 @@ namespace bsp::keyboard
...
+    bool write(const drivers::I2CAddress &addr, const std::uint8_t *txBuff, const std::size_t size)
+    {
+        for (auto i = 0; i < maxTransmissionRetry; ++i) {
+            auto ret = i2c->Write(addr, txBuff, size);
+            if (ret == size) {
+                return true;
+            }
         }
+        LOG_ERROR("Can't write data to keyboard. I2C error: %ld", i2c->GetLastError());
+        return false;
```
awulkiew, line 84: So at this point the keyboard will stop to respond, correct?
What should we do in this case?
Shouldn't we e.g. restart the phone if that happens since the phone is going to be unusable anyway?

module-bsp/board/rt1051/puretx/bsp/keyboard/keyboard.cpp
```diff
@@ -67,24 +68,47 @@ namespace bsp::keyboard
...
 
-        void clearAllIRQs()
-        {
-            std::uint8_t clearIRQs = 0xFF;
-            i2cAddr.subAddress     = REG_INT_STAT;
-            i2c->Write(i2cAddr, reinterpret_cast<std::uint8_t *>(&clearIRQs), 1);
+    bool write(const drivers::I2CAddress &addr, const std::uint8_t *txBuff, const std::size_t size)
+    {
+        for (auto i = 0; i < maxTransmissionRetry; ++i) {
+            auto ret = i2c->Write(addr, txBuff, size);
```
awulkiew, line 78: Is this call blocking i.e. we're waiting here for the end of transmission?

Let's consider a case when transmission failed. Does the write start just right after the previous one? If that's the case and the error was caused by some condition of the i2c could this condition be still present? If that's the case maybe we should wait a while between tries?

module-bsp/board/rt1051/puretx/bsp/keyboard/keyboard.cpp
dawidwojtas-mudita: First of all, I want to register transmission errors on i2c bus. I haven't considered restarting the phone yet, but I am afraid if i2c bus is broken, the firmware can restart all the time.

module-bsp/board/rt1051/puretx/bsp/keyboard/keyboard.cpp
dawidwojtas-mudita: Yes, we are waiting for the end of transmission.

In this case, next transmission starts immediately after the previous one. So that's a good tip. We can wait a bit and try to write/read again. How long should we wait 1 or 5ms?

module-bsp/board/rt1051/puretx/bsp/keyboard/keyboard.cpp
Lefucjusz: I'm not sure whether soft-restarting the phone is really a good solution here - when I encountered this bug, the only way to make phone responsive again was to remove battery, what physically "restarted" all ICs to initial state. I assume there was some unrecoverable condition on the bus that couldn't be fixed by OS restart, e.g. some IC was holding SDA low, what rendered further communication impossible.

dawidwojtas-mudita: It is also hard to reproduce so at this point I prefer to gather data and then try to find the reason for this fault.

module-bsp/board/rt1051/puretx/bsp/keyboard/keyboard.cpp
awulkiew: Your guess is as good as mine. :)
The times you propose looks reasonable.

I'd only add that we may consider waiting different ammounts of time for each call in case this condition was time-related. And because it may be caused by another part in the system sending messages (which is typically done in constant intervals) maybe these time values should be something more random? E.g. 1253us and 4723us?

dawidwojtas-mudita: I suppose that we can increase delay time in case when the operation fails eg. the next delay time will be two times longer.  Do you want to random delay every time or just hardcoded delays? I think that this is a bit overcomplicated. :)

awulkiew: I was thinking about something hardcoded. Increasing the time is ok too to play well with `maxTransmissionRetry`.

> I think that this is a bit overcomplicated.

Yes, proabably. It was just a question so if you think this is not needed then I'm ok with it. :)

swichu91: These are all valid points, but I want you to think about responsibility here for a second. Do you think it is a good idea to have this simple method perform retransmission logic? Imho it is not ;). The context here is too narrow to perform any meaningful action(why if in specific context 3 retransmission will be not enough?). Why not return a specific error code and then let the higher layers do something with it?

module-bsp/board/rt1051/drivers/RT1051DriverI2C.cpp
```diff
@@ -76,13 +77,19 @@ namespace drivers
...
             return -1; // TODO:M.P: fix me
         }
         else {
             return size;
         }
     }
 
+    int32_t RT1051DriverI2C::GetLastError()
+    {
+        return lastError;
```
swichu91, line 90: Could you please explain what the purpose of this method is? Wouldn't it be better to just return a correct retu code and then handle it later?

module-bsp/drivers/i2c/DriverI2C.hpp
```diff
@@ -49,6 +49,8 @@ namespace drivers
 
         virtual ssize_t Modify(const I2CAddress &addr, const uint32_t mask, bool setClr, const size_t size) = 0;
 
+        virtual int32_t GetLastError() = 0;
```
swichu91, line 52: I believe you missed adding simulator implementation? I can only see a `RT1051` one. In this case it won't compile for linux/simulator.

module-bsp/board/rt1051/drivers/RT1051DriverI2C.cpp
```diff
@@ -3,7 +3,7 @@
 
 #include "RT1051DriverI2C.hpp"
 #include <log/log.hpp>
-#include "../board.h"
+#include <board.h>
```
swichu91, line 6:
```suggestion
#include "board.h"
```
I think it is sufficient as `board.h` is the part of the same module/source directory `module-bsp`

dawidwojtas-mudita: You're right :)

module-bsp/board/rt1051/drivers/RT1051DriverI2C.cpp
dawidwojtas-mudita: I agree that GetLastError function isn't the best solution here. I added proper error codes for I2C and a function that translate generic error codes to I2C specific.

dawidwojtas-mudita: Done

module-bsp/drivers/i2c/DriverI2C.hpp
dawidwojtas-mudita: Done

module-bsp/board/rt1051/puretx/bsp/keyboard/keyboard.cpp
```diff
@@ -67,101 +77,140 @@ namespace bsp::keyboard
...
+        power(KeyboardPower::on);
+        // Recovery time is 120us
+        vTaskDelay(pdMS_TO_TICKS(chipRecoveryMs));
+    }
 
-        gpio->ConfPin(DriverGPIOPinParams{.dir      = DriverGPIOPinParams::Direction::Input,
-                                          .irqMode  = DriverGPIOPinParams::InterruptMode::IntRisingOrFallingEdge,
-                                          .defLogic = 0,
-                                          .pin      = static_cast<uint32_t>(BoardDefinitions::KEYBOARD_RF_BUTTON)});
+    bool write(const drivers::I2CAddress &addr, const std::uint8_t *txBuff, const std::size_t size)
```
SP2FET, line 101: just wondering - wouldn't it look nicer with returning `std::pair` with ret code and buffer pointer instead of C-style parameter passing?
The same goes to `read`

It's only a suggestion - if you don't find it useful - feel free to resolve it anytime :D

module-bsp/board/rt1051/puretx/bsp/keyboard/keyboard.cpp
```diff
@@ -67,101 +77,140 @@ namespace bsp::keyboard
...
         // Enable interrupts
         i2cAddr.subAddress = REG_CFG;
-        i2c->Write(i2cAddr, reinterpret_cast<std::uint8_t *>(&reg), 1);
+        write(i2cAddr, reinterpret_cast<std::uint8_t *>(&reg), 1);
 
         // Get key pressed/released count
         std::uint8_t val   = 0;
         i2cAddr.subAddress = REG_KEY_LCK_EC;
-        i2c->Read(i2cAddr, reinterpret_cast<std::uint8_t *>(&val), 1);
+        read(i2cAddr, reinterpret_cast<std::uint8_t *>(&val), 1);
```
SP2FET, line 171: we don't check the return codes on purpose?

module-bsp/board/rt1051/puretx/bsp/keyboard/keyboard.cpp
```diff
@@ -67,101 +77,140 @@ namespace bsp::keyboard
...
+        }
+        LOG_ERROR("Can't write data to keyboard. I2C error: %d", ret);
+        return false;
+    }
 
-        gpio->ConfPin(DriverGPIOPinParams{.dir      = DriverGPIOPinParams::Direction::Output,
-                                          .irqMode  = DriverGPIOPinParams::InterruptMode::NoIntmode,
-                                          .defLogic = 0,
-                                          .pin      = static_cast<uint32_t>(BoardDefinitions::KEYBOARD_RESET_PIN)});
+    bool read(const drivers::I2CAddress &addr, std::uint8_t *rxBuff, const std::size_t size)
```
SP2FET, line 115: consider adding [[nodiscard]]

swichu91 (approved):

module-bsp/board/rt1051/puretx/bsp/keyboard/keyboard.cpp
```diff
@@ -210,15 +259,15 @@ namespace bsp::keyboard
...
+            read(i2cAddr, reinterpret_cast<std::uint8_t *>(&retval), 1);
 
             std::uint8_t eventsCount = retval & keyCountRegisterMask;
             for (std::uint8_t i = 0; i < eventsCount; i++) {
                 std::uint8_t key      = 0;
                 std::uint8_t rel_pres = 0;
 
                 i2cAddr.subAddress = REG_KEY_EVENT_A;
-                i2c->Read(i2cAddr, reinterpret_cast<std::uint8_t *>(&retval), 1);
+                read(i2cAddr, reinterpret_cast<std::uint8_t *>(&retval), 1);
```
SP2FET, line 271: here I think we should check if the command succeeded, although as we don't have any mechanism to restart the I2C with success (as @Lefucjusz said) it's not critical for now

dawidwojtas-mudita: We don't know why the keyboard fails sometimes. The `read` function is trying to retransmit if something went wrong. If it doesn't help we just log this event. The resetting chip or other actions will be considered in the future when we find an exact problem.

module-bsp/board/rt1051/puretx/bsp/keyboard/keyboard.cpp
dawidwojtas-mudita: More or less, yes. After powerup, the keyboard chip is reset and then we assume that communication works fine. I've never seen any issue with the keyboard while initialization. 
I discussed with @Lefucjusz and @awulkiew about initialization and as far as we don't know the exact issue we don't reset the chip.

module-bsp/board/rt1051/puretx/bsp/keyboard/keyboard.cpp
dawidwojtas-mudita: Thanks for your suggestion. It is a good idea but most drivers use this same pattern of parameter passing. So I leave it now as it is.

module-bsp/board/rt1051/puretx/bsp/keyboard/keyboard.cpp
dawidwojtas-mudita: As we don't have any logic for handling errors we don't use always return code.

module-bsp/board/rt1051/puretx/bsp/keyboard/keyboard.cpp
SP2FET: alright, but I think that we should handle it in the future ;)

module-bsp/board/rt1051/puretx/bsp/keyboard/keyboard.cpp
SP2FET: fair enough

module-bsp/board/rt1051/puretx/bsp/keyboard/keyboard.cpp
SP2FET: alright

module-bsp/board/rt1051/puretx/bsp/keyboard/keyboard.cpp
dawidwojtas-mudita: Definitely, I agree :)

PR #3999, opened 2022-08-03
Fix Mudita Center synchronization fail

Maciej-Mudita:
**Description**

Sometimes the user was unable to connect
to the Mudita Center due to sync errors

Make sure that this PR:

### DIFF
--- a/module-services/service-desktop/WorkerDesktop.cpp
+++ b/module-services/service-desktop/WorkerDesktop.cpp
@@ -66,10 +66,12 @@ void WorkerDesktop::closeWorker(void)
     while (parser.getCurrentState() != sdesktop::endpoints::State::NoMsg && --maxcount > 0) {
         vTaskDelay(300);
     }
+
+    initialized = false;
     LOG_INFO("we can deinit worker now");
 
     /// additional wait to flush on serial - we should wait for data sent...
-    vTaskDelay(3000);
+    vTaskDelay(500);
 
     bsp::usbDeinit();
 
@@ -84,7 +86,6 @@ void WorkerDesktop::closeWorker(void)
     cpuSentinel.reset();
 
     LOG_DEBUG("deinit end");
-    initialized = false;
 }
 
 bool WorkerDesktop::reinit(const std::filesystem::path &path)
@@ -99,6 +100,7 @@ bool WorkerDesktop::reinit(const std::filesystem::path &path)
 
 void WorkerDesktop::reset()
 {
+    initialized = false;
     usbStatus = bsp::USBDeviceStatus::Disconnected;
     bsp::usbDeinit();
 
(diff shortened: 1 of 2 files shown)


module-services/service-desktop/WorkerDesktop.cpp
```diff
@@ -113,95 +115,126 @@ void WorkerDesktop::reset()
 
 bool WorkerDesktop::handleMessage(uint32_t queueID)
```
dawidwojtas-mudita, line 116: Use `std::uint32_t' instead. :)

dawidwojtas-mudita (approved):

Lefucjusz (approved):

SP2FET (approved):

pawel-mudita (approved):

module-services/service-desktop/WorkerDesktop.cpp
Maciej-Mudita: done

module-services/service-desktop/WorkerDesktop.cpp
```diff
@@ -113,95 +115,126 @@ void WorkerDesktop::reset()
...
 {
+    auto result       = false;
```
awulkiew, line 118: I think we should use `auto` more conservatively. This is a very useful language feature to improve the readability when some long type would otherwise be necessary but for short and known types I think we should refrain from using it. And here is the edge case where `auto` is used to replace `bool` so a known type with name of the same length as `auto`.

module-services/service-desktop/WorkerDesktop.cpp
```diff
@@ -113,95 +115,126 @@ void WorkerDesktop::reset()
...
 
-        parser.setMessageHandler(std::move(handler));
-        parser.processMessage(std::move(*receivedMsg));
+bool WorkerDesktop::handleReceiveQueueMessage(std::shared_ptr<sys::WorkerQueue> &queue)
+{
+    if (not initialized) {
+        return false;
+    }
+    std::string *receivedMsg = nullptr;
+    if (!queue->Dequeue(&receivedMsg, 0)) {
```
awulkiew, line 147: Logical not operators are not consistent. I'm not sure which one is better and whether or not there is a guideline about this. I have no preference. But we should at least use them consistently in the same function or maybe the entire file.

Side note: sometimes people put space after `!`, i.e.:
```
if (! queue->Dequeue(&receivedMsg, 0))
```
so this would be yet another style.

module-services/service-desktop/WorkerDesktop.cpp
Maciej-Mudita: done

module-services/service-desktop/WorkerDesktop.cpp
Maciej-Mudita: done

module-services/service-desktop/WorkerDesktop.cpp
```diff
@@ -113,95 +115,126 @@ void WorkerDesktop::reset()
...
 
-        delete receivedMsg;
+    using namespace sdesktop::endpoints;
+    auto factory = EndpointFactory::create(securityModel.getEndpointSecurity());
+    auto handler = std::make_unique<MessageHandler>(ownerService, messageProcessedCallback, std::move(factory));
+
+    parser.setMessageHandler(std::move(handler));
+    parser.processMessage(std::move(*receivedMsg));
+
+    delete receivedMsg;
```
awulkiew, line 159: This is a side note regarding the ownership of `receivedMsg`. It's unrelated to this PR specifically but to the overall design. Two different forms of ownership change are used here. Or in other words ownership of the content of `receivedMsg` is separated from the ownership of the memory of the object itself. This is confusing and error-prone.

I think that the queue should be refactored to return objects by moving them to the caller. Then here the message should be moved into `processMessage` and `delete` wouldn't be necessary. The ownership would be clear.

module-services/service-desktop/WorkerDesktop.cpp
```diff
@@ -113,95 +115,126 @@ void WorkerDesktop::reset()
...
-        std::string *sendMsg = nullptr;
-        if (!queue->Dequeue(&sendMsg, 0)) {
-            LOG_ERROR("handleMessage xQueueReceive failed for %s.", sdesktop::SEND_QUEUE_BUFFER_NAME);
-            return false;
-        }
 
-        bsp::usbCDCSend(sendMsg);
-        delete sendMsg;
+    bsp::usbCDCSend(sendMsg);
+    delete sendMsg;
```
awulkiew, line 174: Here is the same story, however here the message seems to be copied into `usbCDCSend`. Maybe it should be moved as well?

awulkiew (approved):

module-services/service-desktop/WorkerDesktop.cpp
Maciej-Mudita: Done

module-services/service-desktop/WorkerDesktop.cpp
Maciej-Mudita: A fair point, I fully agree. However, I think this is a topic for a separate task.

module-services/service-desktop/WorkerDesktop.cpp
awulkiew: Actually `bsp::usbCDCSend` probably takes it by `const&` so this change is probably only semantic.

awulkiew: Actually actually no, it seems that `bsp::usbCDCSend` expects a pointer to `std::string`. So it's not a matter of ownership but a matter of consistency of handling messages. Above it's taken by reference but here by pointer.

module-services/service-desktop/WorkerDesktop.cpp
```diff
@@ -111,97 +113,128 @@ void WorkerDesktop::reset()
...
-    else if (qname == sdesktop::SEND_QUEUE_BUFFER_NAME) {
-        std::string *sendMsg = nullptr;
-        if (!queue->Dequeue(&sendMsg, 0)) {
-            LOG_ERROR("handleMessage xQueueReceive failed for %s.", sdesktop::SEND_QUEUE_BUFFER_NAME);
-            return false;
-        }
 
-        bsp::usbCDCSend(sendMsg);
-        delete sendMsg;
+    bsp::usbCDCSend(std::move(sendMsg));
```
awulkiew, line 173: After applying my suggestion this will probably not compile as `bsp::usbCDCSend` takes a pointer. It would require to refactor `bsp::usbCDCSend` to take r-val reference. So maybe we should leave it as it was. ;)

awulkiew: So I think that `std::move` should be removed because it further suggest that this is an actual message object and this makes things even worse.

Maciej-Mudita: Done

PR #4002, opened 2022-08-03
MOS-641_read-active-sim-cardss-data-earlier-but-ask-for-unlocking-later-1

marziel:
Make sure that this PR:

### DIFF
--- a/module-apps/application-settings/presenter/network/SimPINSettingsPresenter.cpp
+++ b/module-apps/application-settings/presenter/network/SimPINSettingsPresenter.cpp
@@ -26,7 +26,7 @@ void SimPINSettingsPresenter::togglePinState()
     }
 }
 
-bool SimPINSettingsPresenter::getPinState() const noexcept
+bool SimPINSettingsPresenter::isPinEnabled() const noexcept
 {
     return isPinOn;
 }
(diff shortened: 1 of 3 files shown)


dawidwojtas-mudita (approved):

SP2FET:
Is this PR doing anything else than a rename? Maybe you missed to commit some more changes?

marziel:
> Is this PR doing anything else than a rename? Maybe you missed to commit some more changes?

For now, it isn't yet. When analysing the code, I found out that this name was confusing me, hence the renaming.

SP2FET:
> > Is this PR doing anything else than a rename? Maybe you missed to commit some more changes?
> 
> For now, it isn't yet. When analysing the code, I found out that this name was confusing me, hence the renaming.

So please either:
- rename the title of the PR
- wait with PR until any more changes appear

Or just turn it into a draft, as this PR doesn't seem to fix the MOS-641 issue

marziel:
Well, it looks like we ran into a following problem: there's a low hanging fruit to improve the code, even if slightly, and the changes would be related as kind of preparation work to some bigger changes. But, it's impossible to incorporate them into _master_ without:
a) merging alongside some bigger (related) changes,
b) creating a de facto dummy, new Jira ticket just to this purpose.
I thunk this should be somhow addressed.

marziel:
(ditto)

SP2FET:
> Well, it looks like we ran into a following problem: there's a low hanging fruit to improve the code, even if slightly, and the changes would be related as kind of preparation work to some bigger changes. But, it's impossible to incorporate them into _master_ without: a) merging alongside some bigger (related) changes, b) creating a de facto dummy, new Jira ticket just for this purpose. I think this should be somehow addressed.

That's why I suggested turning it into draft - you can continue your work then. If you wanted to do the rename, If I were you, I'd create a subtask in Jira and then create a PR under this subtask's ID - this way we have clean history of which commit fixed which ticket

marziel:
OK, I didn't know we allow creating PRs for subtasks - there are various policies "in the world" regarding this. So, this way seems reasonable to me.

PR #4017, opened 2022-08-10
MOS-101 Parse MMS notification and show [MMS] in message thread

awulkiew:
The `[MMS]` notification is shown in message thread of the correct number

- Add PDU WAP Push MMS Notification parser
- Use it in ServiceCellular to parse the raw message
- Refactor hexToBytes and bytesToHex utils
- Add endsWith util
- Add unit tests

**Description**

Make sure that this PR:

### DIFF
--- a/module-services/service-cellular/Pdu.cpp
+++ b/module-services/service-cellular/Pdu.cpp
@@ -1,0 +1,982 @@
+// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
+// For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
+
+#include "Pdu.hpp"
+
+#include <Utils.hpp>
+
+#include <algorithm>
+#include <array>
+#include <cassert>
+#include <cstdint>
+#include <variant>
+
+namespace pdu::constants
+{
+
+    // PDU Types
+    // http://www.openmobilealliance.org/release/Browser_Protocol_Stack/V2_1-20110315-A/OMA-WAP-TS-WSP-V1_0-20110315-A.pdf,
+    // p98
+    enum Type : std::uint8_t
+    {
+        Reserved = 0x00,
+        Connect,
+        ConnectReply,
+        Redirect,
+        Reply,
+        Disconnect,
+        Push,
+        ConfirmedPush,
+        Suspend,
+        Resume,
+        Get = 0x40,
+        Options,
+        Head,
+        Delete,
+        Trace,
+        Post = 0x60,
+        Put,
+        DataFragment = 0x80
+    };
+
+    // Well-known Header Field Names
+    // http://www.openmobilealliance.org/release/Browser_Protocol_Stack/V2_1-20110315-A/OMA-WAP-TS-WSP-V1_0-20110315-A.pdf,
+    // p103
+    enum FieldName : std::uint8_t
+    {
+        /* v1.1 */
+        Accept = 0x00,
+        AcceptCharset11,  // deprecated
+        AcceptEncoding11, // deprecated
+        AcceptLanguage,
+        AcceptRanges,
+        Age,
+        Allow,
+        Authorization,
+        CacheControl11, // deprecated
+        Connection,
+        ContentBase11, // deprecated
+        ContentEncoding,
+        ContentLanguage,
+        ContentLength,
+        ContentLocation,
+        ContentMD5,
+        ContentRange11, // deprecated
+        ContentType,
+        Date,
+        Etag,
+        Expires,
+        From,
+        Host,
+        IfModifiedSince,
+        IfMatch,
+        IfNoneMatch,
+        IfRange,
+        IfUnmodifiedSince,
+        Location,
+        LastModified,
+        MaxForwards,
+        Pragma,
+        ProxyAuthenticate,
+        ProxyAuthorization,
+        Public,
+        Range,
+        Referer,
+        RetryAfter,
+        Server,
+        TransferEncoding,
+        Upgrade,
+        UserAgent,
+        Vary,
+        Via,
+        Warning,
+        WWWAuthenticate,
+        ContentDisposition11, // deprecated
+        /* v1.2 */
+        XWapApplicationId, // X-Wap-Application-ID
+        XWapContentURI,
+        XWapInitiatorURI,
+        AcceptApplication,
+        BearerIndication,
+        PushFlag,
+        Profile,
+        ProfileDiff,
+        ProfileWarning12, // deprecated
+        /* v1.3 */
+        Expect13,
+        TE,
+        Trailer,
+        AcceptCharset,
+        AcceptEncoding,
+        CacheControl13, // deprecated
+        ContentRange,
+        XWapTod,
+        ContentID,
+        SetCookie,
+        Cookie,
+        EncodingVersion,
+        /* v1.4 */
+        ProfileWarning,
+        ContentDisposition,
+        XWAPSecurity,
+        CacheControl,
+        /* v1.5 */
+        Expect,
+        XWapLocInvocation,
+        XWapLocDelivery
+    };
+
+    // Basic Well-Known Content Types
+    // http://www.openmobilealliance.org/release/Browser_Protocol_Stack/V2_1-20110315-A/OMA-WAP-TS-WSP-V1_0-20110315-A.pdf,
+    // p105
+    // http://www.wapforum.org/wina/wsp-content-type.htm
+    enum ContentType : std::uint8_t
+    {
+        AnyAny = 0x00,                         // "*/*"
+        TextAny,                               // "text/*"
+        TextHtml,                              // "text/html"
+        TextPlain,                             // "text/plain"
+        TextXHdml,                             // "text/x-hdml"
+        TextXTtml,                             // "text/x-ttml"
+        TextXVCalendar,                        // "text/x-vCalendar"
+        TextXVCard,                            // "text/x-vCard"
+        TextVndWapWml,                         // "text/vnd.wap.wml"
+        TextVndWapWmlScript,                   // "text/vnd.wap.wmlscript"
+        TextVndWapWtaEvent,                    // "text/vnd.wap.wta-event"
+        MultipartAny,                          // "multipart/*"
+        MultipartMixed,                        // "multipart/mixed"
+        MultipartFormData,                     // "multipart/form-data"
+        MultipartByteranges,                   // "multipart/byteranges"
+        MultipartAlternative,                  // "multipart/alternative"
+        ApplicationAny,                        // "application/*"
+        ApplicationJavaVm,                     // "application/java-vm"
+        ApplicationXWwwFormUrlencoded,         // "application/x-www-form-urlencoded"
+        ApplicationXHdmlc,                     // "application/x-hdmlc"
+        ApplicationVndWapWmlc,                 // "application/vnd.wap.wmlc"
+        ApplicationVndWapWmlscriptc,           // "application/vnd.wap.wmlscriptc"
+        ApplicationVndWapWtaEventc,            // "application/vnd.wap.wta-eventc"
+        ApplicationVndWapUaprof,               // "application/vnd.wap.uaprof"
+        ApplicationVndWapWtlsCaCertificate,    // "application/vnd.wap.wtls-ca-certificate"
+        ApplicationVndWapWtlsUserCertificate,  // "application/vnd.wap.wtls-user-certificate"
+        ApplicationXX509CaCert,                // "application/x-x509-ca-cert"
+        ApplicationXX509UserCert,              // "application/x-x509-user-cert"
+        ImageAny,                              // "image/*"
+        ImageGif,                              // "image/gif"
+        ImageJpeg,                             // "image/jpeg"
+        ImageTiff,                             // "image/tiff"
+        ImagePng,                              // "image/png"
+        ImageVndWapWbmp,                       // "image/vnd.wap.wbmp"
+        ApplicationVndWapMultipartAny,         // "application/vnd.wap.multipart.*"
+        ApplicationVndWapMultipartMixed,       // "application/vnd.wap.multipart.mixed"
+        ApplicationVndWapMultipartFormData,    // "application/vnd.wap.multipart.form-data"
+        ApplicationVndWapMultipartByteranges,  // "application/vnd.wap.multipart.byteranges"
+        ApplicationVndWapMultipartAlternative, // "application/vnd.wap.multipart.alternative"
+        ApplicationXml,                        // "application/xml"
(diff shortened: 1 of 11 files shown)


module-utils/utility/Utils.hpp
```diff
@@ -13,26 +13,106 @@
...
 #include <tuple>
+#include <type_traits>
 #include <functional>
 
 namespace utils
 {
     inline constexpr auto WHITESPACE = " \n\r\t\f\v";
-    constexpr unsigned int secondsInMinute =
-        std::chrono::duration_cast<std::chrono::seconds>(std::chrono::minutes(1)).count();
+    constexpr unsigned int secondsInMinute = 60;
```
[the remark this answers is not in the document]
awulkiew: I hope it's correct. :)

module-services/service-cellular/ServiceCellular.cpp
```diff
@@ -949,6 +950,11 @@ std::optional<std::shared_ptr<sys::Message>> ServiceCellular::identifyNotificati
     return urcHandler.getResponse();
 }
 
+inline std::string numberFromAddress(std::string const &address)
```
awulkiew, line 953: Since this is cpp file `inline` keyword is probably not needed.

module-utils/utility/Utils.hpp
```diff
@@ -13,26 +13,106 @@
...
+
+    // WARNING: Unsafe
+    template <typename T, std::enable_if_t<is_byte_v<T>, int> = 0>
+    inline void byteToHexAt(T c, std::string &result, std::size_t offset)
+    {
+        result[offset]     = halfByteToHex<std::uint8_t>((c & 0xF0) >> 4);
+        result[offset + 1] = halfByteToHex<std::uint8_t>(c & 0x0F);
+    }
+
+    template <typename T, std::enable_if_t<is_byte_v<T>, int> = 0> inline std::string byteToHex(T c)
```
awulkiew, line 43: Some of these functions are not used anywhere. Do we need them or should I removed them?

module-utils/utility/Utils.hpp
```diff
@@ -13,26 +13,106 @@
...
+    inline std::string bytesToHex(const Bytes &bytes)
+    {
+        std::string result(bytes.size() * 2, '0');
+        for (std::size_t i = 0; i < bytes.size(); ++i) {
+            byteToHexAt(bytes[i], result, i * 2);
+        }
+        return result;
+    }
+
+    template <typename T = std::uint8_t, std::enable_if_t<is_byte_v<T>, int> = 0> inline T halfHexToByte(char c)
```
awulkiew, line 60: This formatting was done with our clang-format. I think this is harder to read than e.g. a version with line break. What do you think about changing clang-format option `AlwaysBreakTemplateDeclarations` from `MultiLine` to `Yes` here: https://github.com/mudita/MuditaOS/blob/master/.clang-format#L23 ?

module-services/service-cellular/Pdu.cpp
```diff
@@ -0,0 +1,972 @@
...
+#include <Utils.hpp>
+
+#include "Pdu.hpp"
+
+namespace pdu::constants
+{
+
+    // PDU Types
+    // http://www.openmobilealliance.org/release/Browser_Protocol_Stack/V2_1-20110315-A/OMA-WAP-TS-WSP-V1_0-20110315-A.pdf,
+    // p98
```
SP2FET, line 19: really nice that you linked the source!

module-services/service-cellular/Pdu.cpp
```diff
@@ -0,0 +1,972 @@
...
+        bool next(std::uint8_t &result)
+        {
+            if (m_begin != m_end) {
+                result = *m_begin++;
+                return true;
+            }
+            return false;
+        }
+
+        void eat()
```
SP2FET, line 416: just wondering - shouldn't eat/consume method somehow take out the data? as it's being eaten/consumed. Here you just move the pointer further, without altering the data

module-services/service-cellular/Pdu.cpp
```diff
@@ -0,0 +1,972 @@
...
+            std::array<bool, N> headersFound;
+            headersFound.fill(false);
+            std::size_t foundCount = 0;
+
+            while (!octets.empty()) {
+                std::size_t nameId      = N;
+                HeaderValue headerValue = std::uint8_t(0);
+
+                auto pageShiftPred = [&](std::uint8_t page) {
+                    // TODO: Support page shifts
```
SP2FET, line 798: please don't leave TODO in the code - create a followup task in jira:
https://github.com/mudita/MuditaOS/blob/master/doc/MuditaCppCodingGuidelines.md#m-9-there-shall-be-no-comments-like-todo-fixme-etc

module-services/service-cellular/Pdu.cpp
```diff
@@ -0,0 +1,972 @@
...
+                };
+                auto namePred = [&](auto &&name) {
+                    for (std::size_t i = 0; i < N; ++i) {
+                        if (headersFound[i]) {
```
SP2FET, line 802: wouldn't `.at()` method be safer than brackets? I know that it probably won't go out of index, but you know :D

module-services/service-cellular/Pdu.cpp
```diff
@@ -0,0 +1,972 @@
...
+
+        // https://www.openmobilealliance.org/release/MMS/V1_3-20110913-A/OMA-TS-MMS_ENC-V1_3-20110913-A.pdf, p55
+        Text address;
+        if (std::holds_alternative<Octets>(mmsNotificationFields.fromAddress)) {
+            Octets octets      = std::get<Octets>(mmsNotificationFields.fromAddress);
+            std::uint8_t token = 0;
+            if (!octets.next(token)) {
+                return std::nullopt;
+            }
+            static const std::uint8_t AddressPresentToken = 128;
```
SP2FET, line 950: why this has to be static? can't it be constexpr?

module-services/service-cellular/ServiceCellular.cpp
```diff
@@ -1075,7 +1081,20 @@ auto ServiceCellular::receiveSMS(std::string messageNumber) -> bool
...
 
-                    const auto decodedMessage = UCS2(messageRawBody).toUTF8();
+                    UTF8 decodedMessage;
+
+                    const std::string decodedStr  = utils::hexToBytes<std::string>(messageRawBody);
+                    const auto mmsNotificationOpt = pdu::parse(decodedStr);
+                    if (mmsNotificationOpt) {
+                        std::string number = numberFromAddress(mmsNotificationOpt->fromAddress);
+                        // NOTE: number may be empty
+                        decodedMessage = UTF8("[MMS]");
```
SP2FET, line 1091: wondering if MMS in other languages has different spelling - if so, we'd need to put it through our translation system

module-utils/utility/Utils.hpp
SP2FET: IMO those can be removed - if there will be need for them, then they will be created/reverted ;)

module-utils/utility/Utils.hpp
```diff
@@ -13,23 +13,109 @@
...
+            return c - 'A' + 10;
+        }
+        else if (c >= 'a' && c <= 'f') {
+            return c - 'a' + 10;
+        }
+        throw std::invalid_argument("Unexpected hex digit");
+        return 0;
+    }
+
+    // WARNING: Unsafe
```
SP2FET, line 75: could you please elaborate? 
Maybe there is a way to make it more safe?

SP2FET (approved):

module-utils/utility/Utils.hpp
awulkiew: I double-checked and it seems that:
- `byteToHex` is used in `QMBNManager.cpp`
- `bytesToHex` is not used
- `hexToByte` is not used
- `hexToBytes` is used in `QMBNManager.cpp` and `ServiceCellular.cpp` (added in this PR)

So only two of them would be removed.

module-services/service-cellular/Pdu.cpp
awulkiew: Yes, `next()` advances the iterator and returns the data. `eat()` is there to only advance the iterator after `peek()`. There is also `eatN()` used to advance the iterator by N bytes if e.g. a header is detected that is not supported and can be ignored. If you think some other naming would be preferable I'm open to suggestion.

module-services/service-cellular/Pdu.cpp
awulkiew: Old habit. I'll change it. ;)

module-utils/utility/Utils.hpp
awulkiew: There is no check of the length of `hex` or value of `offset`. This is a function implementing common part of other functions and which should not be called directly.

module-services/service-cellular/ServiceCellular.cpp
awulkiew: Yes, it seems there is. I wanted to avoid localization but since we have to do this anyway we could as well think about more elaborate message, e.g.: `MMS not supported`.

module-services/service-cellular/Pdu.cpp
awulkiew: In general yes, but here `N` is compile-time constant. I don't think that `at()` throwing an exception is needed here. Replacing the loop with recursive compile-time metafunction would be more complex than it's needed as well.

module-utils/utility/Utils.hpp
awulkiew: If I remove 2 of these functions then there is no need for these unsafe common parts to exist.

module-services/service-cellular/ServiceCellular.cpp
awulkiew: However AFAIS in all languages that we support it's called MMS. Just like it's SMS in all supported languages. So I think it can stay as it is. I'm not an expert though.

module-services/service-cellular/Pdu.cpp
awulkiew: Or we could use C++17 fold expressions instead of metafunction and check index at compile-time instead of throwing an exception, e.g.:
```
auto namePred = [&](auto &&name) {
    return any(
        [&](auto index) {
            static_assert(index < N);
            // Header was not found before and is required
            if (!headersFound[index] && isKnown(name, headerNames[index])) {
                headersFound[index] = true;
                ++foundCount;
                nameId = index;
                return true;
            }
            return false;
        },
        std::make_index_sequence<N>());
};
                
...

template <typename IndexPred, std::size_t... N>
static bool any(IndexPred indexPred, std::index_sequence<N...>)
{
    return (indexPred(std::integral_constant<std::size_t, N>()) || ...);
}
```

But I think this is not needed.

If you insist on checking `i` I'd rather use `assert(i < N)` instead of `at()` throwing and exception because if `i` was out of range it'd be programmer's fault, not an exceptional case. But I don't think this is needed either in this case.

module-services/service-cellular/Pdu.cpp
SP2FET: I'd say that `proceed` or something like this would be more appropriate here - when I see `eat` function I'd assume, that the data is somehow lost

module-services/service-cellular/Pdu.cpp
SP2FET: sure, you're right ;)

module-services/service-cellular/Pdu.cpp
awulkiew: What do you think about `ignore()` which is a name used in `istream`? I could also rename `next()` with `get()` to be consistent.

module-services/service-cellular/Pdu.cpp
```diff
@@ -0,0 +1,978 @@
...
+                octets.eat();
+                std::uint32_t length = lengthOctet;
+                if (lengthOctet == UintvarEscape) {
+                    if (!parseUintvar(octets, length)) {
+                        return false;
+                    }
+                }
+                if (octets.size() < length) {
+                    return false;
+                }
```
awulkiew, line 606: I've added this check and similar one below to return 'false' if an incomplete message was sent. Or several partial messages were not correctly concatenated into one notification. Before this change the returned value contained as many characters as possible and parsing was considered as successful.
There is an edge case however. It's impossible to detect incomplete message if it ends exactly where one of the headers ends. In such case `true` will still be returned and incomplete number of headers will be parsed.

module-utils/utility/Utils.hpp
awulkiew: Done

module-services/service-cellular/Pdu.cpp
awulkiew: Done

module-utils/utility/Utils.hpp
awulkiew: Done

module-services/service-cellular/Pdu.cpp
SP2FET: as discussed on private channel - ignore and next will be fine ;)

module-services/service-cellular/ServiceCellular.cpp
SP2FET: fine ;)

module-utils/utility/Utils.hpp
awulkiew: I'll create a separate PR for this.

SP2FET:
please add a changelog entry before merging ;)

module-services/service-cellular/Pdu.cpp
```diff
@@ -0,0 +1,982 @@
...
+
+#include <algorithm>
+#include <array>
+#include <cassert>
+#include <cstdint>
+#include <variant>
+
+#include <Utils.hpp>
+
+#include "Pdu.hpp"
```
dawidwojtas-mudita, line 12: According to our guidelines, the `Pdu.hpp` header should be first. https://github.com/mudita/MuditaOS/blob/master/doc/MuditaCppCodingGuidelines.md#m-3-sort-the-include-files

dawidwojtas-mudita (approved):

module-services/service-cellular/Pdu.cpp
awulkiew: Done

pure_changelog.md
```diff
@@ -12,6 +12,7 @@
 * Fixed turning on loudspeaker before outgoing call is answered
 * Fixed PLAY label translation in German
 * Fixed USB connection/disconnection detection
+* Fixed MMS handling
```
SP2FET, line 15: I'd say something like "Added simple MMS handling"

awulkiew: Done

kkleczkowski (approved):

PR #4018, opened 2022-08-11
[MOS-658] Make windows switch accordingly

marziel:
**Description**

Make sure that this PR:

### DIFF
--- a/.gdbinit
+++ b/.gdbinit
@@ -1,3 +1,3 @@
-source .gdb_macros
+source ~/.gdbinit
 
 handle SIGUSR1 nostop noprint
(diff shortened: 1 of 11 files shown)


marziel:
The remaining thing to do in this subtask is to make the window flow return to the `SIM cards` window after the user has committed too many mistakes and now the SIM card needs PUK instead of PIN (notifying this fact to the user will be handled separately). This is hard because the object `locks::SimLockHandler simLockHandler` is a part of `ApplicationManager` and `ApplicationSettings` has no direct influence on it. One way to resolve this may be memorizing the pop-up's order in `ApplicationSettings` and deciding upon it.

marziel:
Still doesn't work entirely as expected.

PR #4020, opened 2022-08-11
[MOS-259] Fix no calllog entries in DND mode

SP2FET:
**Description**

Fixed no calllog entries in DND mode, which were caused
by not starting the call in DB interface, so no entry was created.

Moreover, small optimizing of the DND handling.

Make sure that this PR:

### DIFF
--- a/module-services/service-cellular/call/CallMachine.hpp
+++ b/module-services/service-cellular/call/CallMachine.hpp
@@ -38,6 +38,8 @@ namespace call
     {
         bool operator()(CallData &call)
         {
+            LOG_DEBUG("Called RING, status: %s",
+                      call.mode == sys::phone_modes::PhoneMode::Connected ? "true" : "false");
             return call.mode == sys::phone_modes::PhoneMode::Connected;
         }
     } constexpr RING;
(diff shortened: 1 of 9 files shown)


SP2FET:
This will probably have to wait with merging until we'll fix MOS-602, as it will create multiple calllog entries while rejecting..
.

module-services/service-cellular/call/CallMachine.hpp
```diff
@@ -143,10 +148,16 @@ namespace call
     {
         void operator()(Dependencies &di, const call::event::CLIP &clip, CallData &call)
         {
+            // the order here is crucial - rejecting by modem is first, because of how DND works -
+            // we have to reject incomming call ASAP.
```
Lefucjusz, line 152: "incoming" ;)

Lefucjusz (approved):

module-services/service-cellular/call/CallMachine.hpp
SP2FET: thanks, fixed

SP2FET:
closing due to duplicate

PR #4024, opened 2022-08-12
[BH-1542] Pure meditation fix and cetner TimeFixedWidget

disonek:
*DigitsContainer definicions moved to cpp file(for proper linkage)
*DigitsContainer changed container to std::vector
*Fixed a bug in ProgressTimerWithBarGraphAndCounter where timerText was
treated as timeWidget
*MeditationCountdownWindow timer is now custom size
*Minus removed form MeditationRunningWindow timer

Make sure that this PR:

### DIFF
--- a/module-apps/apps-common/widgets/DigitsContainer.cpp
+++ b/module-apps/apps-common/widgets/DigitsContainer.cpp
@@ -1,0 +1,49 @@
+// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
+// For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
+
+#include "DigitsContainer.hpp"
+
+#include <Utils.hpp>
+
+namespace gui
+{
+
+    void setDigits(std::string text, const DigitsContainer &container, DimensionsParams params)
+    {
+        std::for_each(std::crbegin(container.digits),
+                      std::crend(container.digits),
+                      [text = std::move(text), &params](Label *label) mutable {
+                          if (text.empty()) {
+                              label->setText("");
+                              label->setSize(0, 0);
+                          }
+                          else {
+                              label->setText(std::string{text.back()});
+                              label->setSize(params.digitMaxWidth, params.mainBoxHeight);
+                              text.pop_back();
+                          }
+                      });
+    }
+
+    DigitsContainer::DigitsContainer(uint32_t digitsCount) : digits(digitsCount, nullptr)
+    {}
+
+    void DigitsContainer::setMinutesBox(std::uint32_t minutes, DimensionsParams params)
+    {
+        constexpr auto rangeGuard = 1000;
+        minutes %= rangeGuard;
+
+        setDigits(std::to_string(minutes), *this, params);
+    }
+
+    void DigitsContainer::setSecondsBox(std::uint32_t seconds, DimensionsParams params)
+    {
+        constexpr auto rangeGuard = 100;
+        seconds %= rangeGuard;
+
+        std::string text = utils::addLeadingZeros(std::to_string(seconds), digits.size());
+
+        setDigits(std::move(text), *this, params);
+    }
+
+} // namespace gui
--- a/module-apps/apps-common/widgets/DigitsContainer.hpp
+++ b/module-apps/apps-common/widgets/DigitsContainer.hpp
@@ -18,45 +18,13 @@ namespace gui
         std::uint32_t rightBoxWidth;
     };
 
-    template <size_t N> struct DigitsContainer
+    struct DigitsContainer
     {
-        std::array<Label *, N> digits;
+        explicit DigitsContainer(std::uint32_t digitsCount);
+        std::vector<Label *> digits;
+
         void setMinutesBox(std::uint32_t minutes, DimensionsParams params);
         void setSecondsBox(std::uint32_t seconds, DimensionsParams params);
     };
-    template <size_t N> void setDigits(std::string text, const DigitsContainer<N> &container, DimensionsParams params)
-    {
-        std::for_each(std::crbegin(container.digits),
-                      std::crend(container.digits),
-                      [text = std::move(text), &params](Label *label) mutable {
-                          if (text.empty()) {
-                              label->setText("");
-                              label->setSize(0, 0);
-                          }
-                          else {
-                              label->setText(std::string{text.back()});
-                              label->setSize(params.digitMaxWidth, params.mainBoxHeight);
-                              text.pop_back();
-                          }
-                      });
-    }
 
-    template <size_t N> void DigitsContainer<N>::setMinutesBox(std::uint32_t minutes, DimensionsParams params)
-    {
-        auto rangeGuard = 1000;
-        minutes %= rangeGuard;
-
-        setDigits(std::to_string(minutes), *this, params);
-    }
-
-    template <size_t N> void DigitsContainer<N>::setSecondsBox(std::uint32_t seconds, DimensionsParams params)
-    {
-        auto rangeGuard = 100;
-        seconds %= rangeGuard;
-
-        bool shouldBeDisplayedAsOneDigit = seconds < 10;
-        std::string text = shouldBeDisplayedAsOneDigit ? "0" + std::to_string(seconds) : std::to_string(seconds);
-
-        setDigits(std::move(text), *this, params);
-    }
 } // namespace gui
--- a/module-apps/apps-common/widgets/ProgressTimerWithBarGraphAndCounter.cpp
+++ b/module-apps/apps-common/widgets/ProgressTimerWithBarGraphAndCounter.cpp
@@ -18,12 +18,32 @@ namespace app
     {
         updateText();
         updateProgress();
+        updateTimeWidget();
         ProgressTimer::update();
     }
 
     void ProgressTimerWithBarGraphAndCounter::updateText()
     {
-        const auto secondsRemaining = duration - std::chrono::duration_cast<std::chrono::seconds>(elapsed);
+        using utils::time::Duration;
+        if (text == nullptr) {
+            return;
+        }
+        const auto secondsRemaining = std::chrono::duration_cast<std::chrono::seconds>(duration - elapsed);
+        const Duration remainingDuration{secondsRemaining};
+        UTF8 timerText;
+        if (countdownMode == ProgressCountdownMode::Increasing && secondsRemaining != std::chrono::seconds::zero()) {
+            timerText += increasingModePrefix;
+        }
+        timerText += remainingDuration.str(displayFormat);
+        text->setText(std::move(timerText));
+    }
+
+    void ProgressTimerWithBarGraphAndCounter::updateTimeWidget()
+    {
+        if (timeWidget == nullptr) {
+            return;
+        }
+        const auto secondsRemaining = std::chrono::duration_cast<std::chrono::seconds>(duration - elapsed);
         timeWidget->setMinutesBox(secondsRemaining.count() / 60);
         timeWidget->setSecondsBox(secondsRemaining.count() % 60);
     }
--- a/module-apps/apps-common/widgets/TimeFixedWidget.hpp
+++ b/module-apps/apps-common/widgets/TimeFixedWidget.hpp
@@ -17,15 +17,21 @@ namespace gui
       public:
         struct LeftBox
         {
+            explicit LeftBox(std::uint32_t digitsCount);
+            LeftBox();
+            static constexpr std::uint32_t defaultSize{3};
             HBox *box    = nullptr;
             Label *minus = nullptr;
-            DigitsContainer<3> container{};
+            DigitsContainer container;
         };
 
         struct RightBox
         {
+            explicit RightBox(std::uint32_t digitsCount);
+            RightBox();
+            static constexpr std::uint32_t defaultSize{2};
             HBox *box = nullptr;
-            DigitsContainer<2> container{};
+            DigitsContainer container;
         };
 
         TimeFixedWidget(Item *parent,
@@ -33,13 +39,24 @@ namespace gui
                         const std::uint32_t &y,
                         const std::uint32_t &w,
                         const std::uint32_t &h,
-                        const bool minus = false);
+                        bool minus = false);
+
+        TimeFixedWidget(Item *parent,
+                        const std::uint32_t &x,
+                        const std::uint32_t &y,
+                        const std::uint32_t &w,
+                        const std::uint32_t &h,
+                        bool minus,
+                        std::uint32_t leftBoxSize,
+                        std::uint32_t rightBoxSize);
+
+        void init(std::uint32_t w, std::uint32_t h);
 
         void setMinutesBox(std::uint32_t first);
         void setSecondsBox(std::uint32_t second);
         void setFontAndDimensions(const UTF8 &fontName) const;
 
-        DimensionsParams getDimensions() const;
+        DimensionsParams getInitialDimensions() const;
 
       private:
         void attachLabelToBox(Label *&label, HBox *&box) const;
--- a/module-apps/tests/widgets/test-DigitsContainer.cpp
+++ b/module-apps/tests/widgets/test-DigitsContainer.cpp
@@ -11,26 +11,27 @@
 namespace
 {
 
-    template <size_t N> void checkDisplayedText(std::string text, const gui::DigitsContainer<N> &container)
+    void checkDisplayedText(std::string &&text, const gui::DigitsContainer &container)
     {
-        std::for_each(std::crbegin(container.digits), std::crend(container.digits), [&text](gui::Label *label) {
-            if (text.empty()) {
-                EXPECT_STREQ((label->getText()).c_str(), "");
-            }
-            else {
-                EXPECT_STREQ(label->getText().c_str(), &text.back());
-                text.pop_back();
-            }
-        });
+        std::for_each(std::crbegin(container.digits),
+                      std::crend(container.digits),
+                      [text = std::move(text)](gui::Label *label) mutable {
+                          if (text.empty()) {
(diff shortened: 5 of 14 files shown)


module-apps/apps-common/widgets/DigitsContainer.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+namespace gui
+{
+
+    DigitsContainer::DigitsContainer(uint32_t digitsCount) : digits(digitsCount, nullptr)
+    {}
+
+    void DigitsContainer::setMinutesBox(std::uint32_t minutes, DimensionsParams params)
+    {
+        auto rangeGuard = 1000;
```
SP2FET, line 14: constexpr

module-apps/apps-common/widgets/DigitsContainer.cpp
```diff
@@ -0,0 +1,48 @@
...
+        minutes %= rangeGuard;
+
+        setDigits(std::to_string(minutes), *this, params);
+    }
+
+    void DigitsContainer::setSecondsBox(std::uint32_t seconds, DimensionsParams params)
+    {
+        auto rangeGuard = 100;
```
SP2FET, line 22: constexpr

module-apps/apps-common/widgets/DigitsContainer.hpp
```diff
@@ -16,47 +16,20 @@ namespace gui
         std::uint32_t minusWidth;
         std::uint32_t leftBoxWidth;
         std::uint32_t rightBoxWidth;
+        std::uint32_t avaliableEvenWidth;
+        bool notCentral{false};
```
SP2FET, line 20: what this flag does? Basing on the name, I can't figure it out at the first look. Consider renaming to `isNotCentered` or something like that

module-apps/apps-common/widgets/DigitsContainer.hpp
```diff
@@ -16,47 +16,20 @@ namespace gui
...
     };
 
-    template <size_t N> struct DigitsContainer
+    struct DigitsContainer
     {
-        std::array<Label *, N> digits;
+        explicit DigitsContainer(std::uint32_t digitsCount);
+        ;
```
SP2FET, line 26: doubled semicolon?

module-apps/apps-common/widgets/ProgressTimerWithBarGraphAndCounter.cpp
```diff
@@ -18,11 +18,31 @@ namespace app
...
         ProgressTimer::update();
     }
 
     void ProgressTimerWithBarGraphAndCounter::updateText()
     {
+        using utils::time::Duration;
+        if (text == nullptr) {
+            return;
+        }
+        const auto secondsRemaining = duration - std::chrono::duration_cast<std::chrono::seconds>(elapsed);
```
SP2FET, line 31: As both are `std::chrono`'s durations, why not do something like:
```Cpp
    const auto secondsRemaining = std::chrono::duration_cast<std::chrono::seconds>(duration - elapsed);

File: module-apps/apps-common/widgets/ProgressTimerWithBarGraphAndCounter.cpp [hunk 6]
```diff
@@ -18,11 +18,31 @@ namespace app
...
+        const Duration remainingDuration{std::time_t{secondsRemaining.count()}};
```
SP2FET, line 32: consider adding constructor in the `Duration` class for `std::chrono`'s object

module-apps/tests/widgets/test-DigitsContainer.cpp
```diff
@@ -11,7 +11,7 @@
 namespace
 {
 
-    template <size_t N> void checkDisplayedText(std::string text, const gui::DigitsContainer<N> &container)
+    void checkDisplayedText(std::string text, const gui::DigitsContainer &container)
```
SP2FET, line 14:
```suggestion
    void checkDisplayedText(const std::string& text, const gui::DigitsContainer &container)
```

module-apps/apps-common/widgets/DigitsContainer.cpp
```diff
@@ -0,0 +1,48 @@
...
+        seconds %= rangeGuard;
+
+        bool shouldBeDisplayedAsOneDigit = seconds < 10;
+        std::string text = shouldBeDisplayedAsOneDigit ? "0" + std::to_string(seconds) : std::to_string(seconds);
```
SP2FET, line 26: consider using:
```utils::addLeadingZeros```

File: module-apps/apps-common/widgets/ProgressTimerWithBarGraphAndCounter.cpp (hunk 5 above)
Comment (disonek) [reply module-apps/apps-common/widgets/ProgressTimerWithBarGraphAndCounter.cpp:31]: The code was restored form inital commit introducing(deleting this code) TimeFixedWidget so I did not want to change anything but its ok change I will change to:   
   const auto secondsRemaining = std::chrono::duration_cast<std::chrono::seconds>(duration - elapsed);
  
as you suggested

File: module-apps/apps-common/widgets/ProgressTimerWithBarGraphAndCounter.cpp (hunk 6 above)
Comment (disonek) [reply module-apps/apps-common/widgets/ProgressTimerWithBarGraphAndCounter.cpp:32]: The code was restored form inital commit introducing(deleting this code) TimeFixedWidget so I did not want to change anything. The benefit from new constructor would be lack of casting to time_t however I would like to not introduce changes in Duration as its working and tested

File: module-apps/tests/widgets/test-DigitsContainer.cpp (hunk 7 above)
Comment (disonek) [reply module-apps/tests/widgets/test-DigitsContainer.cpp:14]: checkDisplayedText function modifies text so text cant be passed ass const & but I wll change to &&

Comment (SP2FET) [reply module-apps/tests/widgets/test-DigitsContainer.cpp]: so then:
- function name is incorrect
- something is `no yes`

**Comment by SP2FET** [comment]
image
https://github.com/mudita/MuditaOS/blob/master/doc/development_workflow.md#pr-review---answer-questions ;)

File: module-utils/time/test/unittest_time.cpp [hunk 9]
```diff
@@ -519,6 +519,12 @@ TEST_CASE("Duration - creation")
         REQUIRE(duration.get() == 333);
     }
 
+    SECTION("std::chrono::seconds constructor")
+    {
+        utils::time::Duration duration(std::chrono::seconds(333));
+        REQUIRE(duration.get() == 333);
```
SP2FET, line 525: what will happen, when you pass e.g. minutes? duration will be presented in minutes, or casted to seconds?
I mean - should we return the count, or absolute duration?

module-apps/tests/widgets/test-DigitsContainer.cpp
disonek: What would be your proposal than for "ideal" name?

module-utils/time/test/unittest_time.cpp
disonek: As Duration now takes second I would suppouse that minues will be implictly converted by chrono library

module-apps/tests/widgets/test-DigitsContainer.cpp
SP2FET: I'd move the `pop` out of the function - if it has a checking purpose, it shouldn't alter the parameter - somehow related to single responsibility rule :P

module-utils/time/test/unittest_time.cpp
SP2FET: I won't agree there - looking what you wrote, duration will be feeded with argument's `count` - so it will have count only, not absolute duration in seconds. If that's desired (that we have duration in seconds) you should modify the `get` method and add casting to seconds there

SP2FET: that's why both good and bad path in unit tests are useful ;)

module-apps/tests/widgets/test-DigitsContainer.cpp
disonek: If you would remove pop out of the function please add comment related to pop move out of this function. I would disagree this function takse a Rvalue as parameter to check that has proper value. How it brakes single responsibility  ?

module-apps/apps-common/widgets/ProgressTimerWithBarGraphAndCounter.cpp
```diff
@@ -18,11 +18,31 @@ namespace app
...
+        timerText += remainingDuration.str(displayFormat);
+        text->setText(std::move(timerText));
+    }
+
+    void ProgressTimerWithBarGraphAndCounter::updateTimeWidget()
+    {
+        if (timeWidget == nullptr) {
+            return;
+        }
         const auto secondsRemaining = duration - std::chrono::duration_cast<std::chrono::seconds>(elapsed);
```
SP2FET, line 46: and please, do me a favor and rewrite this one too ;)

module-utils/time/test/unittest_time.cpp
disonek: test like this would pass:
 SECTION("std::chrono::seconds by minutes constructor")
    {
        auto minutes = std::chrono::minutes(2);
        utils::time::Duration duration(minutes);
        REQUIRE(duration.get() == 120);
    }
    Could  you please  give me real example how to break it ?

SP2FET: my bad - I didn't see that you made the constructor requires `std::chrono::seconds` instead of generic duration: https://en.cppreference.com/w/cpp/chrono/duration

So the casting/conversion is going under the hood.

module-apps/tests/widgets/test-DigitsContainer.cpp
SP2FET:
```
The idea behind the SRP is that every class, module, or function in a program should have one responsibility/purpose in a program```
If the function is named `checkSomething` I'd say that it will do some testing and leave the parameters untouched. If it do anything on the arguments which are passed by reference/pointer, then it shouldn't be named this way, as it's confusing.

More over, I believe that this pop should be moved outside the function with the corresponding `std::for_each` - then you can have a function with testing-only, and rest of the test logic, which is responsible of shifting the text/popping it out/whatever

Comment (disonek) [reply module-apps/tests/widgets/test-DigitsContainer.cpp]: You mentioned that SRP is broken but you did not mentioned these plural responsibilities. I assume that in your understanding this function does **check** text variable and  **modifies** text variable. However this function is not about checking text variable its about checking container on basis of text variable so in my opinion this solution is efficient and for sure does not break SRP. In another words in test environment checks what would be displayed on real device. I will not try to convenient you because we have different opinions/understanding in this topic and thats perfectly fine. 

Despite IMO this whole discussion is out of scope as in this commit  just changed function signature as DigitsContainer is not longer template (name, funcion parameters and body does not changed)and I have different opinion, I'm very grateful for your effort regarding aspiration for better quality code so plase provide the better implementaion containing your ideas. I will paste them in to this commit. It would speedup the review process and everyone will be satisfied.

Comment (dawidwojtas-mudita) [reply module-apps/tests/widgets/test-DigitsContainer.cpp]: Apart from the `SRP` rule, it looks weird when we have the function `check` something and then modifies parameters inside. I agree with @SP2FET that you should at least change the function name. This small change can make this code  more readable.

Comment (disonek) [reply module-apps/tests/widgets/test-DigitsContainer.cpp]: OK lets change the name than. @SP2FET and @dawidwojtas-mudita function name   doesDigitalsContainerWouldDislpayText would be more readable and would not confuse you? Please give your ideas if this name would not be good enough or lacking some way?

Comment (SP2FET) [reply module-apps/tests/widgets/test-DigitsContainer.cpp]: TBH I can't find a proper name for this function, as it behaves quite strange for me - I'd say, that if we want to compare if the container's data is equal to the expected ones, we should just iterate through the container and verify it - the function just pops out the data if ithere's any, and then verifies if it's emptied out...

Comment (SP2FET) [reply module-apps/tests/widgets/test-DigitsContainer.cpp]: or as a compromise - pass it via value and then modify the parameter inside the function only - I dug into the code more, and I see that it might have sense, so that's one of the solutions

Comment (disonek) [reply module-apps/tests/widgets/test-DigitsContainer.cpp]: The naming is quite hard especially there is no precisely rules. Thanks for your try. Once again I encourage you to make proposals for better implementation allowing you easly name this function.

Comment (disonek) [reply module-apps/tests/widgets/test-DigitsContainer.cpp]: I haven't seen your last comment while I posted my last comment. Currently in this review I did change signature to Rvalue reference it fits better because represents binding to temporary object(by default can be modified), but if you would like to pass text variable by copy its perfectly fine for me.

File: module-apps/apps-common/widgets/DigitsContainer.hpp [hunk 11]
```diff
@@ -16,47 +16,19 @@ namespace gui
         std::uint32_t minusWidth;
         std::uint32_t leftBoxWidth;
         std::uint32_t rightBoxWidth;
+        std::uint32_t avaliableEvenWidth;
+        bool isNotCentered{false};
```
swichu91, line 20: Remove it as it is not used.

module-apps/apps-common/widgets/DigitsContainer.hpp
```diff
@@ -16,47 +16,19 @@ namespace gui
...
-    }
-
-    template <size_t N> void DigitsContainer<N>::setSecondsBox(std::uint32_t seconds, DimensionsParams params)
-    {
-        auto rangeGuard = 100;
-        seconds %= rangeGuard;
 
-        bool shouldBeDisplayedAsOneDigit = seconds < 10;
-        std::string text = shouldBeDisplayedAsOneDigit ? "0" + std::to_string(seconds) : std::to_string(seconds);
+    void setDigits(std::string text, const DigitsContainer &container, DimensionsParams params);
```
swichu91, line 32: Move it to cpp file? As far as I know, it is not used anywhere else?

swichu91: Passing by value should be okay thanks to the SSO(short strings optimization).

module-apps/apps-common/widgets/TimeFixedWidget.hpp
```diff
@@ -35,11 +41,22 @@ namespace gui
                         const std::uint32_t &h,
                         const bool minus = false);
 
+        TimeFixedWidget(Item *parent,
+                        const std::uint32_t &x,
+                        const std::uint32_t &y,
+                        const std::uint32_t &w,
+                        const std::uint32_t &h,
+                        const bool minus,
```
swichu91, line 49: Remove `const`?

module-apps/apps-common/widgets/TimeFixedWidget.hpp
```diff
@@ -35,11 +41,22 @@ namespace gui
...
+                        std::uint32_t leftBoxSize,
+                        std::uint32_t rightBoxSize);
+
+        void init(const std::uint32_t &w, const std::uint32_t &h);
```
swichu91, line 53: Do we really want to pass `uint32_t`s by reference here? There is no difference in size in this case but it makes the code more complex than necessary.

module-utils/time/test/unittest_time.cpp
```diff
@@ -519,6 +519,19 @@ TEST_CASE("Duration - creation")
...
```
swichu91, line 525: What's the deal with these methods? Doesn't `std::chrono` provide this functionality already?

module-apps/tests/widgets/test-DigitsContainer.cpp
```diff
@@ -26,9 +26,9 @@ namespace
         EXPECT_TRUE(text.empty());
     }
 
-    template <size_t N> auto initContainer(gui::HBox &box) -> gui::DigitsContainer<N>
+    template <size_t N> auto initContainer(gui::HBox &box) -> gui::DigitsContainer
     {
-        auto container = gui::DigitsContainer<N>{};
+        auto container = gui::DigitsContainer{N};
         for (auto &digit : container.digits) {
```
swichu91, line 33: Unnecessary blank line.

swichu91 (commented):
Please check the comments. Apart from that, please generate the image by invoking `BellHybrid-StandaloneImage` and send it for further testing to @patrycja-paczkowska and @sofianemudita . We need to make sure it doesn't introduce regressions both to the Pure and Harmony as it already happened previously. Thanks in advance!

module-apps/apps-common/widgets/DigitsContainer.hpp
disonek: Its declaration in hpp definition is in cpp, so I can remove definicion in hpp file

module-apps/apps-common/widgets/TimeFixedWidget.hpp
disonek: The origin ctor had const here so I did also add const here. I can remove from both ctors

module-apps/apps-common/widgets/DigitsContainer.hpp
disonek: Its declaration in hpp definition is in cpp, so I can remove definicion in hpp file asd

module-apps/apps-common/widgets/TimeFixedWidget.hpp
disonek: I will remove &

module-apps/apps-common/widgets/DigitsContainer.hpp
swichu91: Yeah, if it is not used by another module, there is no need to have it in public API.

swichu91 (approved):

module-apps/tests/widgets/test-DigitsContainer.cpp
SP2FET: let's leave it as it is - with the move instead of passing by ref - at least it's not confusing so much

SP2FET (approved):

PR #4030, opened 2022-08-18
[BH-1543] I want to set a date

disonek:
*DateSetSpinner and DateSetListItem added (based od TimeSetSpinner and
TimeSetListItem)
*BellSideListItem ctor without title added
*TimeUnitsModel changed to more accurate DateTimeUnitsModel name

Make sure that this PR:


module-apps/apps-common/widgets/DateSetSpinner.cpp
```diff
@@ -0,0 +1,267 @@
...
+        setFont();
+
+        focusChangedCallback = [&](Item &) {
+            if (editMode != EditMode::Edit) {
+                setFocusItem(nullptr);
+                return false;
+            }
+
+            if (lastFocus != nullptr) {
+                updateFocus(lastFocus);
```
SP2FET, line 69: here you pass `lastFocus` as a parameter, but in DateSetSpinner.cpp:235 you set it to `lastFocus`, so basically 
```C
lastFocus = lastFocus?
```

module-apps/apps-common/widgets/DateSetSpinner.cpp
```diff
@@ -0,0 +1,267 @@
...
+        setMinimumSize(getWidgetMinimumAreaWidth(), getFontHeight(noFocusFontName));
+        setMaximumWidth(widgetMaximumArea.w);
+
+        HBox::informContentChanged();
+    }
+
+    std::uint32_t DateSetSpinner::getWidgetMinimumAreaWidth()
+    {
+        return day->widgetMinimumArea.w + firstSlash->widgetMinimumArea.w + month->widgetMinimumArea.w +
+               secondSlash->widgetMinimumArea.w + year->widgetMinimumArea.w;
```
SP2FET, line 112: are those slashes different in width?

module-apps/apps-common/widgets/DateSetSpinner.cpp
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    bool DateSetSpinner::onInput(const InputEvent &inputEvent)
+    {
+        if (editMode != EditMode::Edit) {
+            return false;
+        }
+
+        if (auto ret = this->focusItem->onInput(inputEvent)) {
+            setMaxDay();
```
SP2FET, line 170: are you sure we have to set the last day of the month?
Please rename the function to `setLastDay` if so.

module-apps/apps-common/widgets/DateSetSpinner.cpp
```diff
@@ -0,0 +1,267 @@
...
+            default:
+                break;
+            }
+        }
+        return false;
+    }
+
+    void DateSetSpinner::setMaxDay()
+    {
+        auto dayCoutInMounth =
```
SP2FET, line 190:
```suggestion
        auto dayCountInMonth =
```

module-apps/apps-common/widgets/DateSetSpinner.cpp
```diff
@@ -0,0 +1,267 @@
...
+        else {
+            title->setRichText(utils::translate("app_settings_title_day"));
+            updateFont(day, noFocusFontName);
+            updateFont(month, noFocusFontName);
+            updateFont(year, noFocusFontName);
+            setFocusItem(nullptr);
+        }
+
+        applySizeRestrictions();
+    }
```
SP2FET, line 265: I believe we can do this better than with this bunch of `if`'s
image

products/BellHybrid/apps/application-bell-settings/models/DateTimeUnitsModel.cpp
```diff
@@ -12,37 +13,48 @@
...
 #include <widgets/TimeSetFmtSpinner.hpp>
 
 #include <ctime>
 
 namespace app::bell_settings
 {
-    TimeUnitsModel::TimeUnitsModel(app::ApplicationCommon *app) : application(app)
+    namespace
+    {
+        constexpr auto factoryResetDate   = date::year(2022) / date::month(8) / date::day(18);
```
SP2FET, line 25: As @patrycja-paczkowska said, the default date should be 2022/01/01

products/BellHybrid/apps/application-bell-settings/models/DateTimeUnitsModel.cpp
```diff
@@ -12,37 +13,48 @@
...
+        constexpr int factoryResetTime    = 12 * 60 * 60; // 12:00 PM
```
SP2FET, line 26: can't we write it in a more verbose way?

products/BellHybrid/apps/application-bell-settings/models/DateTimeUnitsModel.cpp
```diff
@@ -12,37 +13,48 @@
...
+    auto DateTimeUnitsModel::getItem(gui::Order order) -> gui::ListItem *
     {
         return getRecord(order);
     }
 
-    void TimeUnitsModel::createData()
+    void DateTimeUnitsModel::createData()
     {
+        dateSetListItem = new gui::DateSetListItem();
+        internalData.push_back(dateSetListItem);
```
SP2FET, line 57: who has ownership of the underlying memory? if internalData, then consider using `std::move`

SP2FET (commented):
Reviewed by @Lefucjusz and @patrycja-paczkowska

module-apps/apps-common/widgets/BellSideListItem.hpp
```diff
@@ -22,10 +16,12 @@ namespace gui
         BellBaseLayout *body = nullptr;
 
         explicit BellSideListItem(const std::string &description);
+        BellSideListItem();
```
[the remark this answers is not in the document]
disonek: not, becaues we need to call  init() functuin

module-apps/apps-common/widgets/DateSetSpinner.cpp
```diff
@@ -0,0 +1,267 @@
...
+    date::year_month_day DateSetSpinner::getDate()
+    {
+        return date::year(year->get_value()) / date::month(month->get_value()) / date::day(day->get_value());
+    }
+
+    void DateSetSpinner::setDate(const date::year_month_day date)
+    {
+        day->set_value(static_cast<unsigned>(date.day()));
+        month->set_value(static_cast<unsigned>(date.month()));
+        year->set_value(static_cast<int>(date.year()));
```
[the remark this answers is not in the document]
disonek: can we cast date::year to unsigned ? :P

module-apps/apps-common/widgets/DateSetSpinner.cpp
disonek: I did copy this part from TimeSetSpinner. In this particular call It would be a self assigment as you have mentioned. But is it a problem here?

module-apps/apps-common/widgets/DateSetSpinner.cpp
disonek: no, would be better for you firstSlash->widgetMinimumArea.w * 2  ?

module-apps/apps-common/widgets/DateSetSpinner.cpp
disonek: Yes Im sure, ok will rename to setLastDay

products/BellHybrid/apps/application-bell-settings/models/DateTimeUnitsModel.cpp
disonek: ok I did not remember the date I will change to 2022/1/1

module-apps/apps-common/widgets/DateSetSpinner.cpp
disonek: This code is copied from TimeSetSpinner. The meme is very funny but could you please be more precisely what would you change here ? Whats wrong in your opinion, the nested If level? code repetition? Would  swith-case for example better here?

products/BellHybrid/apps/application-bell-settings/models/DateTimeUnitsModel.cpp
disonek: dateSetListItem is used in another places in cpp file, using moved form object after move can cause undefined behaviour

module-apps/apps-common/widgets/BellSideListItem.hpp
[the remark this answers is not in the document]
awulkiew: What's the semantic difference between default-constructed item and an item with empty description?

module-apps/apps-common/widgets/BellSideListItem.cpp
```diff
@@ -10,9 +10,7 @@ namespace gui
 {
     BellSideListItem::BellSideListItem(const std::string &description) : ListItemWithDescription(description)
     {
-        setMinimumSize(style::sidelistview::list_item::w, style::sidelistview::list_item::h);
-        setEdges(RectangleEdge::None);
-        body = new BellBaseLayout(this);
+        init();
```
awulkiew, line 13: This is more of a side note because AFAIU this code was here before this PR but I'd like to point out a few things.

There are two `new` calls in the constructor. If an exception is thrown by any other call, e.g. by the second `new` then the allocated memory will not be freed.

Who owns `topMessage`? It's a local variable. The memory is allocated and is not freed anywhere. The same actually applies for `body` and `bottomText`. Am I missing something?

What is the reason of not using smart pointers? Or even better would it be possible to avoid pointers altogether and store these objects as members of this item?

The calls in this constructor are more or less the same as the ones in `setupBottomDescription()`. Shouldn't this logic be encapsulated in some common function called in both places?

products/BellHybrid/apps/application-bell-settings/models/DateTimeUnitsModel.cpp
awulkiew: @SP2FET It's owned by `app::InternalModel<gui::ListItem *>`. They are deleted in the desctuctor of this base class.

I have a different but related question though. Why are raw pointers used instead of some smart pointers? If an exception is thrown by `push_back` then there will be a memory leak. Why not use `unique_ptr`? And here we're circling back to @SP2FET question. With `unique_ptr` the ownership would be clear. You either have to keep it as a member of `DateTimeUnitsModel` or move it to `internalData` and use it from there. Or you could use `shared_ptr` and the ownership would also be clear.

Why is lack of clear ownership a problem? Consider calling `DateTimeUnitsModel::clearData()` which calls `InternalModel::eraseInternalData()` which calls delete for all of the items. So at this point the items were deleted but the pointers inside `DateTimeUnitsModel` still are not `null`. So now we could try to call something from a place of memory which isn't ours. Etc. With this design you should probably have to remember to set these pointers to `null`, maybe add an `assert` somewhere to detect calling null pointers, etc. So basiclaly you have to care about them in two places at once. Smart pointers and/or clearly defined ownership would solve this automatically.

disonek: The owner of  dateSetListItem is internalData. As we discussed it internally, memory exception would crash system anyway so its acceptable :P . Generally gui design is based on raw pointers and memory is managed manually. The destructor is called in base/parent/Item class

module-apps/apps-common/widgets/BellSideListItem.cpp
disonek: **topMessage** and **bottomMessage** are owned by **body** , **body** is owned by **this**, as we discussed internally there is not memory leakage. The reason why we cant use smart pointers was already mentioned earlier _"Generally gui design is based on raw pointers and memory is managed manually. The destructor is called in base/parent/Item class"_

awulkiew: Thanks for the explanation. Note that the memory is going to be leaked here if an exception is thrown but currently the whole system will crash if that happens so memory leak is less important I guess.

module-apps/apps-common/widgets/BellSideListItem.hpp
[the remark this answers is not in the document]
awulkiew: I see what is done internally but what does it mean?
Is it simply an item without top message / description or is it something else?
If `topMessage` being null is a valid state then why is it initialized if description is empty?
Why in this case the base class doesn't need a description?
What is the meaning of an `ListItemWithDescription` without description?

products/BellHybrid/apps/application-bell-settings/models/DateTimeUnitsModel.cpp
awulkiew: Thanks for the explanation.

module-apps/apps-common/widgets/BellSideListItem.hpp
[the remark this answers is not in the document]
disonek: _"I see what is done internally but what does it mean?"_
It is quite general question but answers for following questions would give you overview.

_"Is it simply an item without top message / description or is it something else?"_
Yes it is simply an item without top message.

_"If topMessage being null is a valid state then why is it initialized if description is empty?"_
Is this question why **BellSideListItem(const std::string &description)** where description is an empty string would not be chceck by if(description.empty())?
Probably because no one have done as detailed analysis as you and naming in this class is a bit off.

_"Why in this case the base class doesn't need a description?"_
Because topMessage(body->first) has to be controlled by the child(DateSetListItem) as the topMerssage has to change "on fly".

_"What is the meaning of an ListItemWithDescription without description?"_
The class ListItemWithDescription introduces the description member but its not used anywhere and its not needed at all. For me its ListItemWithDescription without description its just LIstItem.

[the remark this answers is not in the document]
disonek: Inspired by your questions I did remove ListItemWithDescription class entirely from code base as not needed and misleading. BellSideListItem derives form ListItem now. Check now how it looks like.

module-apps/apps-common/widgets/DateSetSpinner.cpp
```diff
@@ -0,0 +1,263 @@
...
+
+    DateSetSpinner::DateSetSpinner(Item *parent, VBox *titleBox, Length x, Length y, Length w, Length h)
+        : HBox(parent, x, y, w, h)
+    {
+        setAlignment(Alignment(Alignment::Horizontal::Center, Alignment::Vertical::Center));
+        setEdges(RectangleEdge::None);
+
+        title = new TextFixedSize(titleBox);
+        title->setMaximumSize(style::bell_base_layout::w, style::bell_base_layout::outer_layouts_h);
+        title->setFont(style::bell_sidelist_item::title_font);
```
swichu91, line 43: Is this a generic widget that can be used both for Pure/Harmony? If yes, then these definitions cannot be used as they inject harmony-related dependencies. I can see two solutions to this problem:
* move this module to the bell product directory
* make it truly generic(remove harmony dependencies)
Whatever you choose, I'm ok with that.

module-apps/apps-common/widgets/BellSideListItem.cpp
```diff
@@ -1,29 +1,38 @@
...
 
 #include "TextFixedSize.hpp"
-#include "ListItemWithDescription.hpp"
 #include "BellSideListItem.hpp"
 #include "Style.hpp"
 
 namespace gui
 {
-    BellSideListItem::BellSideListItem(const std::string &description) : ListItemWithDescription(description)
+    BellSideListItem::BellSideListItem(const std::string &tile)
```
swichu91, line 10:
```suggestion
    BellSideListItem::BellSideListItem(const std::string &title)
```

swichu91 (commented):
Apart from comments, has the issue with the incorrect font size been solved or at least discussed/approved by the design team?

module-apps/apps-common/widgets/DateSetSpinner.cpp
disonek: I did some refactor, please chceck if its better now

disonek:
> 

Yes its was approved by design team

module-apps/apps-common/widgets/DateSetSpinner.cpp
disonek: I did choose option number 2. I added 2 setters and these are set in DateSetListItem. I did remove not needed :
#include "BellSideListItemStyle.hpp"
and also
#include <apps-common/widgets/BellBaseLayout.hpp>

swichu91 (approved):

module-apps/apps-common/widgets/DateSetSpinner.cpp
disonek: currently as the automatic date correction works different the setLasDay is called in handleEnterKey()

awulkiew (approved):

module-apps/apps-common/widgets/DateSetSpinner.cpp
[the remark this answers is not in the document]
swichu91: It looks a bit quirky but it is the only way to access day/month/yeah data. For more info please check:
`third-party/date/include/date/date.h` lines: 418,375,337

products/BellHybrid/apps/application-bell-settings/models/DateTimeUnitsModel.cpp
swichu91: Fixed

module-apps/apps-common/widgets/DateSetSpinner.cpp
swichu91: In this case it's not a problem. Apart from that I simplified focus update logic a bit ;)

module-apps/apps-common/widgets/DateSetSpinner.cpp
swichu91: I believe it's good enough solution.

module-apps/apps-common/widgets/DateSetSpinner.cpp
swichu91: I did some further refactor based on Onufry's work.

SP2FET (approved):
merge like there's no tomorrow :D

PR #4043, opened 2022-08-26
MOS-685 make small improvements in various places of the code

awulkiew:
- Forward arguments
- Remove unused code
- Remove auto return type when it's not needed
- Add const to member functions
- Use startsWith and endsWith from utils
- Allow instantiation of operators only for specific types
- Caluclate key mask in compile time

Make sure that this PR:
- [X] Has unit tests if possible

### DIFF
--- a/module-gui/gui/Common.hpp
+++ b/module-gui/gui/Common.hpp
@@ -79,60 +79,63 @@ namespace gui
         Horizontal
     };
 
-    template <class T>
-    bool operator&(const T &lhs, const T &rhs)
+    // This namespace exists to prevent instantiation of operators for other classes from namespace gui
+    namespace rectangle_enums
     {
-        return static_cast<uint32_t>(lhs) & static_cast<uint32_t>(rhs);
-    }
-    template <class T>
-    T operator|(const T &lhs, const T &rhs)
-    {
-        return static_cast<T>(static_cast<uint32_t>(lhs) | static_cast<uint32_t>(rhs));
-    }
-    template <class T>
-    T operator|=(const T &lhs, const T &rhs)
-    {
-        lhs = lhs | rhs;
-        return lhs;
-    }
+        enum class RectangleEdge
+        {
+            None   = 0x00,
+            Top    = 0x01,
+            Bottom = 0x02,
+            Left   = 0x04,
+            Right  = 0x08,
+            All    = 0x0F
+        };
 
-    enum class RectangleEdge
-    {
-        None   = 0x00,
-        Top    = 0x01,
-        Bottom = 0x02,
-        Left   = 0x04,
-        Right  = 0x08,
-        All    = 0x0F
-    };
+        enum class RectangleFlatEdge
+        {
+            None        = 0x00,
+            TopLeft     = 0x01,
+            TopRight    = 0x02,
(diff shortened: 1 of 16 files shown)


swichu91 (approved):
:)

module-gui/gui/Common.hpp
```diff
@@ -79,60 +79,72 @@ namespace gui
...
+        enum class RectangleYap
+        {
+            None        = 0x00,
+            TopLeft     = 0x10,
+            TopRight    = 0x20,
+            BottomLeft  = 0x40,
+            BottomRight = 0x80,
+        };
+
+        // If the above scoped enums require operators then maybe they should be regular enums?
```
swichu91, line 123: Let's resolve this issue instead of leaving comment ;)

module-gui/gui/widgets/Item.hpp
```diff
@@ -79,7 +79,7 @@ namespace gui
         BoundingBox drawArea; // drawableArea would be more accurate
                               // maximal bounding box size
         /// gets bounding box for selected area
-        auto area(Area which = Area::Normal) -> BoundingBox &
```
SP2FET, line 82: just asking, why did you change the trailing return to a standard one?

module-utils/utility/tests/unittest_utils.cpp
```diff
@@ -526,6 +526,39 @@ TEST_CASE("Byte to hex")
...
+
+    SECTION("No")
+    {
+        REQUIRE((utils::startsWith("Abcde", "bcd") == false));
+    }
+
+    SECTION("Yes")
+    {
+        REQUIRE((utils::startsWith("Abcde", "Ab") == true));
+    }
```
SP2FET, line 554: I know it might be sufficient, but I think that it may be more verbose (I mean section names :D )

SP2FET (approved):

module-gui/gui/widgets/Item.hpp
awulkiew: Because:
1. For consistency, the standard return type is used for majority of functions in this class.
2. Trailing return type is not required in this case
3. The line is shorter

I don't have a strong opinion about it though. If you prefer the trailing type I can revert it.

But I think that if this is prefered then it should probably be done for all of the functions in the codebase or at least in this file. This way all of the function names would be aligned and the return type would always be at the end. So a reader would know where to look in all of the places. The function names would be easier to read but return types especially for functions with long list of parameters would be harder to read. Depends what we prefer.

module-gui/gui/Common.hpp
awulkiew: I mean this solution works and limits the scope of these operators. My point was that if we use these enums like integers performing bitwise operations and implicitly casting them to `bool` then maybe these shouldn't be class enums but regular enums instead? Since they are used as such.

awulkiew: I removed the comment and one operator that was not used anywhere. I think this is still better than it was, i.e. the scope of the operators is limited, and can stay as it is for now.

module-utils/utility/tests/unittest_utils.cpp
awulkiew: Done.

PR #4045, opened 2022-08-29
[MOS-662][MOS-301] Fix Eink shutdown while restoring data

dawidwojtas-mudita:
While restoring data the timer can turn off power of eink to
save the power. To avoid this situation we need to power on
the eink before wiping out the display. The restore procedure
also performs reversing servises list to close. In this case
we do it twice so the order is wrong.

Make sure that this PR:

### DIFF
--- a/pure_changelog.md
+++ b/pure_changelog.md
@@ -7,6 +7,8 @@
 * Separated system volume from Bluetooth device volume for A2DP
 
 ### Fixed
+* Fixed order of the services while closing system
+* Fixed crash of the E-ink service while restoring system data
 * Fixed removing wrong sentinels
 * Fixed music player behaviour when connecting/disconnecting audio devices
 * Fixed dropping the call during the DND mode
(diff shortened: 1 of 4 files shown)


Maciej-Mudita (approved):

pure_changelog.md
```diff
@@ -3,6 +3,8 @@
 ## Unreleased
 
 ### Fixed
+* Fixed order services while closing the system
```
SP2FET, line 6:
```suggestion
* Fixed order of the services while closing the system
```

pure_changelog.md
```diff
@@ -3,6 +3,8 @@
...
+* Fixed crash of the Eink service while restoring data
```
SP2FET, line 7: what data? :P if OS, maybe it would be better to write it explicit ;)

SP2FET (approved):

pure_changelog.md
dawidwojtas-mudita: Ok, you're right :)

dawidwojtas-mudita: Done :)

pure_changelog.md
dawidwojtas-mudita: Done :)

PR #4050, opened 2022-09-01
[MOS-692] Update slider after tethering is off

dawidwojtas-mudita:
When tethering is on we can change the slider position,
however, we can't change the slider mode. If we change
the slider position after tethering is off we need to update
the slider mode. Fix typo in the Y register.

Make sure that this PR:

### DIFF
--- a/module-bsp/board/rt1051/bsp/magnetometer/magnetometer.cpp
+++ b/module-bsp/board/rt1051/bsp/magnetometer/magnetometer.cpp
@@ -35,8 +35,8 @@ namespace bsp
 
             xQueueHandle qHandleIrq = nullptr;
 
-            bsp::KeyCodes current_parsed = bsp::KeyCodes::Undefined;
-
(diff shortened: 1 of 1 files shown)


Maciej-Mudita (approved):

pure_changelog.md
```diff
@@ -7,6 +7,7 @@
 * Separated system volume from Bluetooth device volume for A2DP
 
 ### Fixed
+* Fixed update slider mode after tethering is off
```
SP2FET, line 10: isn't that a regression fix? if so, I don't think if the entry is necessary

SP2FET (approved):

pure_changelog.md
dawidwojtas-mudita: Yes, it is. I just thought that we need to add everything to the changelog according to the development even if it is a regression :P
I'm about to remove that.

Lefucjusz (approved):

PR #4059, opened 2022-09-09
[MOS-694] Move DB initialization from OS to build scripts

SP2FET:
* Switched to DB initialization at compile time
* Organized and cleaned up db files directories(not finished completely)
* Fixed DB related unit tests
* Fixed two URC-related unit tests that caused memory corruption
* Fixed `ThreadsTable::getBySMSQuery` implementation(memory corruption & returning incorrect data)
* Removed `USE_FS` dependency from db unit tests
* Minor improvements to CMake
* Small fixes for GCC12 build

Make sure that this PR:


tools/init_databases.py
```diff
@@ -0,0 +1,66 @@
...
+
+# import required module
+import os
+import sqlite3
+import argparse
+
+
+# this helper script creates DBs from SQL schema files
+
+def process_file(path: os.path, dst_directory: os.path) -> None:
```
julia-okuniewska, line 13: meaningless function name. Process what? Process for what?

tools/init_databases.py
```diff
@@ -0,0 +1,66 @@
...
+    file_name = os.path.basename(path)
+    print(file_name)
```
julia-okuniewska, line 15: clear out print statements or use `logging` package with `DEBUG` level if this script is meant to use on CI

tools/init_databases.py
```diff
@@ -0,0 +1,66 @@
...
+    print(f"Database name: {db_name}\n")
+    print(f"Executing {path} script into {dst_db_path} database")
+
+    try:
+        connection = sqlite3.connect(dst_db_path)
+        with open(path) as fp:
+            connection.executescript(fp.read())
+        connection.commit()
+    except OSError as e:
+        print(e)
```
julia-okuniewska, line 29: Add meaningful exception message like f"Failed on parsing {db_name}. Error: {e}" or something

tools/init_databases.py
```diff
@@ -0,0 +1,66 @@
...
+    finally:
+        if connection:
+            connection.close()
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Create databases from schema scripts')
```
julia-okuniewska, line 36: There is pattern among python users to define `main` function and run only main method in `if __name__ = "__main__" ` section.

It gives the advantage of returning errors in C-like style (`sys.exit(main())`)
Reference here
https://docs.python.org/3/library/__main__.html#idiomatic-usage

julia-okuniewska (changes requested):
Lack of documentation about new script (and old scripts as well) in `tools` folder.

tools/init_databases.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+# this helper script creates DBs from SQL schema files
+
+def process_file(path: os.path, dst_directory: os.path) -> None:
+    file_name = os.path.basename(path)
+    print(file_name)
+
+    db_name = file_name.split("_0")[0]
+    db_name = db_name + ".db"
```
julia-okuniewska, line 18: Keep in mind that using f-strings is "more pythonic" than "+" sign concatenation & it's far more efficient when there are more strings arguments to join

Reference: https://stackoverflow.com/questions/59180574/string-concatenation-with-vs-f-string

SP2FET: fixed

tools/init_databases.py
SP2FET: added main function + return codes

tools/init_databases.py
SP2FET: fixed

tools/init_databases.py
SP2FET: added logging

tools/init_databases.py
SP2FET: fixed

julia-okuniewska: Ok

tools/init_databases.py
julia-okuniewska: Ok

tools/init_databases.py
julia-okuniewska: Ok

tools/init_databases.py
julia-okuniewska: Ok

tools/init_databases.py
julia-okuniewska: Ok

module-vfs/paths/filesystem_paths.cpp
```diff
@@ -42,6 +44,10 @@ namespace purefs
         {
             return std::filesystem::path{eMMC_disk} / PATH_USER;
         }
+        std::filesystem::path getDataBasePath() noexcept
+        {
+            return std::filesystem::path{eMMC_disk} / PATH_USER / PATH_DB;
```
swichu91, line 49:
```suggestion
            return getUserDiskPath() / PATH_DB;
```

SP2FET: true, fixed ;)

module-vfs/paths/include/purefs/filesystem_paths.hpp
```diff
@@ -14,6 +14,7 @@ namespace purefs
         std::filesystem::path getRootDiskPath() noexcept;
         std::filesystem::path getMfgConfPath() noexcept;
         std::filesystem::path getUserDiskPath() noexcept;
+        std::filesystem::path getDataBasePath() noexcept;
```
swichu91, line 17: How about this?
```suggestion
        std::filesystem::path getDatabasesPath() noexcept;
```

module-db/Database/Database.cpp
```diff
@@ -95,23 +88,9 @@ Database::Database(const char *name, bool readOnly)
...
-    auto filePath = (purefs::dir::getCurrentOSPath() / "user/db");
-    if (!std::filesystem::exists(filePath)) {
-        filePath = (purefs::dir::getUserDiskPath() / "db");
-    }
-
-    LOG_INFO("Running scripts: %s", filePath.c_str());
-    isInitialized_ = initializer->run(filePath.c_str(), InitScriptExtension);
-
-    if (isInitialized_) {
+    if (not isInitialized_) {
```
pawel-mudita, line 91: It is a little bit confusing, that we have` isInitialized_` flag `true` here, but before we print _database init done_ info, we do some additional operations.

pawel-mudita (approved):

module-db/Database/Database.cpp
swichu91: True, fixed.

module-db/tests/Helpers.cpp
```diff
@@ -0,0 +1,150 @@
...
+        const auto fileContent = readContent(filePath.c_str());
+        std::string currentStatement{};
+        std::vector<std::string> statements{};
+
+        std::string line{};
+        for (const auto &c : fileContent) {
+            if (c != '\n') {
+                line += c;
+            }
+            else {
```
marziel, line 46: A minor thing for the future: consider an early `break` instead of an `else`.

module-db/tests/Helpers.cpp
```diff
@@ -0,0 +1,150 @@
...
+        const auto name    = filename.substr(0, filename.find("."));
+        const auto prefix  = name.substr(0, name.find_last_of("_"));
+        const auto postfix = name.substr(name.find_last_of("_") + 1, std::string::npos);
+
+        return {name, prefix, postfix};
+    }
+
+    bool isOnExcludedList(const std::string &name, const std::vector<std::string> &excludedList)
+    {
+        return std::any_of(excludedList.begin(), excludedList.end(), [&name](const auto &entry) {
```
marziel, line 76: Usage of an STL algorithm, a good thing! :)

module-db/tests/Helpers.cpp
```diff
@@ -0,0 +1,150 @@
...
+    /// After proper module-db and service-db split it wont' be needed as every product-specific db unit test will be
+    /// placed in the correct directory.
+    std::filesystem::path getPurePhoneScriptsPath()
+    {
+        return currentFileLocation() / "../../products/PurePhone/services/db/databases/scripts";
+    }
+
+    bool DatabaseScripts::operator()(Database &db)
+    {
+        const std::filesystem::path dbpath = db.getName();
```
marziel, line 139: Not for this PR, but a general remark by the way: I think we could get rid of this unnecessary assignments "padding"...

module-db/tests/Helpers.hpp
```diff
@@ -0,0 +1,88 @@
...
+            Database::deinitialize();
+        }
+
+        Db &get()
+        {
+            return *db;
+        }
+
+      private:
+        void construct(const std::filesystem::path &directory,
```
marziel, line 53: Why this `construct()` function instead of implementing directly in the first constructor?

module-db/tests/Helpers.hpp
```diff
@@ -0,0 +1,88 @@
...
+                       const std::filesystem::path &scriptsDirectory,
+                       bool withDevelopment)
+        {
+            Database::initialize();
+
+            if (std::filesystem::exists(directory)) {
```
marziel, line 59: Checking if the file exists is redundant here: https://en.cppreference.com/w/cpp/filesystem/remove

module-db/tests/Helpers.hpp
swichu91: It's due to the having two constructors taking different parameters :)

module-db/tests/Helpers.cpp
swichu91: Yeah, in this case it might be applicable. I used already implemented solution, and tried to not modify it to save some time.

module-db/tests/Helpers.hpp
marziel: I see :) But why not reuse the more generic constructor inside the less generic one?

swichu91: Ah, that's what you mean. I decided it will be clearer to implement it that way but of course I can change it.

swichu91: Fixed

marziel (approved):

PR #4062, opened 2022-09-12
[MOS-718] Fix part of SIM unlocking

marziel:
In the course of working on MOS-640, fixed a potential bug in message handling and refactored unclear naming and logic there and in some other places.

Make sure that this PR:

### DIFF
--- a/module-apps/apps-common/ApplicationCommon.cpp
+++ b/module-apps/apps-common/ApplicationCommon.cpp
@@ -927,7 +927,7 @@ namespace app
         windowsStack().push(newWindow, windowsFactory.build(this, newWindow), d);
     }
 
-    std::optional<std::string> ApplicationCommon::getPrevWindow(uint32_t count) const
+    std::optional<std::string> ApplicationCommon::getPreviousWindow(std::uint32_t count) const
     {
         return windowsStack().get(count);
     }
--- a/module-apps/apps-common/ApplicationCommon.hpp
+++ b/module-apps/apps-common/ApplicationCommon.hpp
@@ -377,7 +377,7 @@ namespace app
         void pushWindow(const std::string &newWindow, const gui::popup::Disposition &d = gui::popup::WindowDisposition);
         /// getter for previous window name
         /// @ingrup AppWindowStack
-        std::optional<std::string> getPrevWindow(uint32_t count = 1) const;
+        std::optional<std::string> getPreviousWindow(std::uint32_t count = 1) const;
         /// clears windows stack
         /// @ingrup AppWindowStack
         void cleanPrevWindw();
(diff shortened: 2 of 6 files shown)


module-apps/apps-common/locks/handlers/SimLockHandler.cpp
```diff
@@ -113,7 +113,7 @@ namespace locks
         lock.attemptsLeft = attempts;
         lock.lockName     = utils::enumToString(Store::GSM::get()->selected);
 
-        if (simUnlockBlockOnLockedPhone) {
```
SP2FET, line 116: you have a point there - it's much more readable right now, but I'm concerned a little bit with the underscore - I assume that you wanted to highlight that's an internal variable, but I don't know if we use this convention anywhere else?

marziel: No, we don't, but we can start ;) This convention is probably better `mSomething` and surely better than putting leading underscore(s) because they can introduce an UB in some cases. But, I won't insist if a majority of the reviewers don't like it.

module-apps/apps-common/ApplicationCommon.cpp
```diff
@@ -927,7 +927,7 @@ namespace app
         windowsStack().push(newWindow, windowsFactory.build(this, newWindow), d);
     }
 
-    std::optional<std::string> ApplicationCommon::getPrevWindow(uint32_t count) const
+    std::optional<std::string> ApplicationCommon::getPreviousWindow(uint32_t count) const
```
dawidwojtas-mudita, line 930: What about `std::uint32_t`?

module-apps/apps-common/ApplicationCommon.hpp
```diff
@@ -377,7 +377,7 @@ namespace app
         void pushWindow(const std::string &newWindow, const gui::popup::Disposition &d = gui::popup::WindowDisposition);
         /// getter for previous window name
         /// @ingrup AppWindowStack
-        std::optional<std::string> getPrevWindow(uint32_t count = 1) const;
+        std::optional<std::string> getPreviousWindow(uint32_t count = 1) const;
```
dawidwojtas-mudita, line 380: And here `std::uint32_t`?

dawidwojtas-mudita (approved):

module-apps/apps-common/ApplicationCommon.cpp
marziel: Good point, I'll change it :)

module-apps/apps-common/locks/handlers/SimLockHandler.cpp
SP2FET: sure - I'm not stubborn on this particular solution - but we should find one and the only one solution and write it down in a guideline ;)

marziel: OK, will do in the future :) So, for now, I'll let myself leave this change as it is here

PR #4063, opened 2022-09-12
[MOS-550] Implement eink partial refresh

awulkiew:
Detect parts of the screen changed since last update and merge them into
bigger regions. These regions defines parts of the context sent to the
display.
Refresh the region covering all of the parts since this is the most time
consuming part and the size of the refreshed region doesn't change the
time much.
Refresh the whole screen if deep refresh is requested and previously
fast refresh was used. This is needed to prevent unwanted artifacts in
some cases.

Refactor some parts of the gui and display code.

Make sure that this PR:
- [X] Has unit tests if possible


module-gui/gui/core/Renderer.hpp
```diff
@@ -34,10 +34,16 @@ namespace gui
          */
 
       public:
-        virtual ~Renderer() = default;
+        // virtual ~Renderer() = default;
```
Maciej-Mudita, line 37: If it is not needed, delete it

module-services/service-eink/ServiceEink.cpp
```diff
@@ -189,12 +237,77 @@ namespace service::eink
...
+            //         ss << '{' << f.pos_x << ' ' << f.pos_y << ' ' << f.size.width << ' ' << f.size.height << "} ";
+            //     LOG_INFO("Frames: %s", ss.str().c_str());
+            //     ss.str("");
+            //     ss << '{' << refreshFrame.pos_x << ' ' << refreshFrame.pos_y << ' ' << refreshFrame.size.width << ' '
+            //     << refreshFrame.size.height << "} ";
+            //     LOG_INFO("Refresh frame: %s", ss.str().c_str());
+            // }
+        }
+
+        // LOG_INFO("Refresh: %d of context: %d %d", (int)refreshMode, (int)ctx.getW(), (int)ctx.getH());
```
Maciej-Mudita, line 307: If it is not needed, delete it

Maciej-Mudita (approved):

module-services/service-eink/ServiceEink.cpp
awulkiew: It still is. E.g. there are some problems with Harmony which I have to address. What do you think about putting them in an `#ifdef`? Is there a naming convention for macros enabling debug logs?

module-gui/gui/core/Renderer.hpp
awulkiew: Done

module-services/service-eink/ServiceEink.cpp
Maciej-Mudita: It will be a better option. Take a look at similar solutions in module-utils/log/api/log/debug.hpp

module-gui/test/test-catch/test-context.cpp
```diff
@@ -138,13 +124,13 @@ TEST_CASE("Draw vector image test")
...
+    drawCommand.draw(&context);
     std::string dump;
 
     uint32_t offset = 0;
-    for (uint32_t y = 0; y < context->getH(); y++) {
-        for (uint32_t x = 0; x < context->getW(); x++) {
-            uint32_t value = *(context->getData() + offset);
+    for (uint32_t y = 0; y < context.getH(); y++) {
+        for (uint32_t x = 0; x < context.getW(); x++) {
+            uint32_t value = *(context.getData() + offset);
```
dawidwojtas-mudita, line 133: You can use `std::uint32_t` :)

dawidwojtas-mudita (approved):

module-services/service-eink/ServiceEink.cpp
awulkiew: Done

module-gui/test/test-catch/test-context.cpp
awulkiew: Done

PR #4074, opened 2022-09-16
[MOS-666] Fix disappearing setting bars in German

Lefucjusz:
Fix of the issue that with German language
selected manual volume setting bar and
vibration volume setting bar would
disappear when focused.

Make sure that this PR:

### DIFF
--- a/module-apps/application-settings/windows/display-keypad/DisplayLightWindow.cpp
+++ b/module-apps/application-settings/windows/display-keypad/DisplayLightWindow.cpp
@@ -119,6 +119,7 @@ namespace gui
 
         auto spinner = std::make_unique<gui::SpinBoxOptionSettings>(
             utils::translate("app_settings_display_light_brightness"),
+            utils::translate("app_settings_display_light_brightness"),
             std::ceil(brightnessValue / brightnessStep),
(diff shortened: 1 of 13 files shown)


module-apps/application-settings/windows/display-keypad/DisplayLightWindow.cpp
```diff
@@ -118,6 +118,7 @@ namespace gui
         };
 
         auto spinner = std::make_unique<gui::SpinBoxOptionSettings>(
+            utils::translate("app_settings_display_light_brightness"),
             utils::translate("app_settings_display_light_brightness"),
```
Maciej-Mudita, line 122: Should it be duplicated?

Maciej-Mudita (approved):

awulkiew (approved):

module-apps/application-settings/windows/display-keypad/DisplayLightWindow.cpp
Lefucjusz: As we discussed offline - yes ;) After my changes the `SpinBoxOptionSettings` constructor takes two text parameters - first one displayed when the option is not focused, second one when focused. In this case in code there's no need to show different texts, so I just passed the same translation twice.

PR #4075, opened 2022-09-20
[MOS-713] Fix access to the phone before going onboarding

Maciej-Mudita:
Before accepting the EULA,
the user may fully use the Mudita-Center.

Make sure that this PR:

### DIFF
--- a/image/user/db/settings_v2_002-devel.sql
+++ b/image/user/db/settings_v2_002-devel.sql
@@ -21,7 +21,7 @@ INSERT OR IGNORE INTO settings_tab (path, value) VALUES
     ('gs_lock_screen_passcode_is_on', '1'),
     ('gs_display_language', 'English'),
     ('gs_input_language', 'English'),
-    ('gs_eula_accepted', '0'),
+    ('gs_eula_accepted', '1'),
     ('\ApplicationManager\\gs_onboarding_done', '1'),
     ('gs_usb_security', '1'),
     ('gs_os_update_version', '0.00.0'),
--- a/module-services/service-desktop/endpoints/security/SecurityEndpointHelper.cpp
+++ b/module-services/service-desktop/endpoints/security/SecurityEndpointHelper.cpp
@@ -50,14 +50,26 @@ namespace sdesktop::endpoints
 
     auto SecurityEndpointHelper::processStatus(Context & /*context*/) -> http::Code
     {
+        auto result         = http::Code::NoContent;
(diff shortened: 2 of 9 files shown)


Maciej-Mudita:
@pawel-mudita I tested https://github.com/mudita/MuditaOS/blob/master/test/pytest/service-desktop/test_security.py and it works fine.

image/user/db/settings_v2_002-devel.sql
```diff
@@ -21,7 +21,7 @@ INSERT OR IGNORE INTO settings_tab (path, value) VALUES
     ('gs_lock_screen_passcode_is_on', '1'),
     ('gs_display_language', 'English'),
     ('gs_input_language', 'English'),
-    ('gs_eula_accepted', '0'),
```
pawel-mudita, line 24: Why set EULA accepted = true as default?
I see it is then set to false at the beginning of onboarding.

module-services/service-desktop/endpoints/security/SecurityEndpointHelper.cpp
```diff
@@ -50,14 +50,26 @@ namespace sdesktop::endpoints
...
-
-        return security == EndpointSecurity::Allow ? http::Code::NoContent : http::Code::Forbidden;
+        else {
+            switch (security.reason) {
+            case BlockReason::NoReason:
+            case BlockReason::DeviceLocked:
+                result = http::Code::Forbidden;
+                break;
+            case BlockReason::EulaNotAccepted:
+                result = http::Code::Locked;
```
pawel-mudita, line 68: The code name for `EulaNotAccepted` block reason looks a bit confusing here, but I understand it is not easy to find a proper HTTP error code for our needs.

pawel-mudita (approved):

image/user/db/settings_v2_002-devel.sql
Maciej-Mudita: @pawel-mudita is only set to true when compiling with the WITH_DEVELOPMENT_FEATURES flag because then onboarding is automatically skipped

module-services/service-desktop/endpoints/security/SecurityEndpointHelper.cpp
Maciej-Mudita: @pawel-mudita I think I chose the most suitable code anyway from the available and unused ones

kkleczkowski (approved):

PR #4076, opened 2022-09-20
[MOS-504] Mixed SMS messages

dawidwojtas-mudita:
Refactor the old part of SMS handler and
parse the concatenated messages using
std::unordered_map and std::map. Added
unittest to the SMS parser.

Make sure that this PR:


module-services/service-cellular/SMSParser.cpp
```diff
@@ -0,0 +1,112 @@
...
+#include <ucs2/UCS2.hpp>
+
+#include <algorithm>
+#include <ctime>
+#include <map>
+#include <sstream>
+#include <unordered_map>
+#include <vector>
+
+namespace SMSParser
```
kkleczkowski, line 16: :+1:

kkleczkowski (approved):
Lokks good to me. Please think about clearing duplicated message. It can be done in separate task.

module-services/service-cellular/SMSParser.cpp
```diff
@@ -0,0 +1,112 @@
...
+        std::unordered_map<std::uint16_t, std::map<std::uint8_t, std::string>> concatenatedMessage;
+        std::string message;
+        UTF8 number;
+    } // namespace
+
+    bool parse(const std::string *smsData)
+    {
+        bool parsed = false;
+
+        if (smsData[0].find(SMSReadMsgName) != std::string::npos) {
```
Lefucjusz, line 33: What about inverting the condition and making early return here instead of enclosing all the code inside this `if`? Just a suggestion though, it's a matter of personal preference I suppose, feel free to close if you don't agree with me ;)

pure_changelog.md
```diff
@@ -8,6 +8,7 @@
 * Separated system volume from Bluetooth device volume for A2DP
 
 ### Fixed
+* Fixed mixed sms messages
```
Lefucjusz, line 11:
```suggestion
* Fixed mixed SMS messages
```

Lefucjusz (approved):

module-services/service-cellular/SMSParser.cpp
dawidwojtas-mudita: I think that your suggestion is correct. It is done :)

PR #4089, opened 2022-09-29
[MOS-242] Fix windows flow after PIN mistakes

marziel:
After too many PIN mistakes during a PIN change,
the user saw distorted windows instead of a PUK
request. Now, this has been fixed.

Also, a typo in a translation has been corrected.

Make sure that this PR:


module-services/service-cellular/src/SimCard.cpp
```diff
@@ -225,11 +225,17 @@ namespace cellular
                 if (auto state = simState(); state) {
                     handleSimState(*state);
                 }
-            }
-            else if (result != sim::Result::OK) {
-                if (onUnhandledCME)
+            } else if (result != sim::Result::OK) {
+		if (result == sim::Result::SIM_PUKRequired) {
+			// MOS-242: silence this case to avoid a misleading "wrong PUK" pop-up
```
marziel, line 230: This is a hack/workaround for today's cellular message handling: the true cause of this issue lies in doubled PUK request /"need PUK" error messages passed down our system. By making this change we silence one of them, allowing the second one to do its job so that the proper "enter PUK" pop-up is displayed instead of the distorted windows.

dawidwojtas-mudita (approved):

PR #4107, opened 2022-10-10
BH-1581 Prevent meditation settings reset with deep press

awulkiew:
Set all settings values each time a value is changed instead of doing it after pressing enter and moving to the next settings page.

Make sure that this PR:

### DIFF
--- a/harmony_changelog.md
+++ b/harmony_changelog.md
@@ -6,6 +6,8 @@
 
 ### Fixed
 
+* Fixed resetting meditation settings with deep press.
+
 ### Added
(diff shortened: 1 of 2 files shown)


awulkiew:
An alternative would be to check deep press in `app::meditation::SettingsWindow::onInput()` and call `app::meditation::SettingsPresenter::saveData()`. This solution would probably require less saves but I think it would be less clean and less consistent with handling of `chimeVolume`.

Maciej-Mudita (approved):

PR #4120, opened 2022-10-21
[CP-1427] Cleanup leftovers after old backup restore implementation

pawel-mudita:
The synchronization mechanism has been freed from the constraints of the old backup and restore implementation.

Make sure that this PR:


module-services/service-desktop/Sync.cpp
```diff
@@ -0,0 +1,205 @@
...
+            return false;
+        }
+
+        LOG_INFO("Deleting file ...");
+
+        if (!std::filesystem::remove(direntry.path(), errorCode)) {
+            LOG_ERROR("Deleting file failed, error: %d, quitting...", errorCode.value());
+            mtar_close(&tarFile);
+            Sync::RemoveSyncDir(path);
+            return false;
```
piotr-tanski, line 185: Why don't delete all the files at the end of the operation in a bulk?
Is this really a good reason to return false?

module-services/service-desktop/Sync.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+    return true;
+}
+
+bool Sync::CreateSyncDir(const std::filesystem::path &path)
+{
+    LOG_INFO("Creating sync directory %s...", path.c_str());
+    std::error_code errorCode;
+
+    if (!std::filesystem::exists(path)) {
```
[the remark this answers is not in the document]
pawel-mudita: I agree, fixed.

module-services/service-desktop/Sync.cpp
```diff
@@ -0,0 +1,205 @@
...
+{
+    return direntry.path() != "." && direntry.path() != ".." && direntry.path() != "...";
+}
+
+Sync::CompletionCode Sync::PrepareSyncPackage(sys::Service *ownerService, std::filesystem::path &path)
+{
+    assert(ownerService != nullptr);
+    LOG_INFO("Sync package preparation started...");
+
+    if (!Sync::RemoveSyncDir(path)) {
```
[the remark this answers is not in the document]
pawel-mudita: An error will be returned as a response to the sync package request, but yes - a new request can be received then, to start the sync process again.

module-services/service-desktop/Sync.cpp
```diff
@@ -0,0 +1,205 @@
...
+    if (!DBServiceAPI::DBPrepareSyncPackage(ownerService, path)) {
+        LOG_ERROR("Sync package preparation, quitting...");
+        Sync::RemoveSyncDir(path);
+        return CompletionCode::DBError;
+    }
+
+    LOG_DEBUG("Packing files");
+    if (!Sync::PackSyncFiles(path)) {
+        LOG_ERROR("Failed pack sync files");
+        Sync::RemoveSyncDir(path);
```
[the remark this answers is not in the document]
pawel-mudita: Sync dir should be removed only in case of these two errors, not always in PrepareSyncPackage() function.
Not sure if using gsl::finally will help here.

module-services/service-desktop/Sync.cpp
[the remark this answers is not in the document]
piotr-tanski: But - if I understand the code correctly - the same directory will be passed again to that function, and most probably it fill fail again, right? If we aren't able to remove the directory for some reason, we probably won't be able to do so in the next request.
Can we work around such an error somehow? E.g. pass another path?

module-services/service-desktop/Sync.cpp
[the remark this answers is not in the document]
piotr-tanski: OK, agreed. It looked like a missing call to `RemoveSyncDir` below for me. :)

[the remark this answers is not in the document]
pawel-mudita: We can't because the tar file is in this directory :)
It needs to be downloaded by Mudita Center first.
After the file is downloaded, then MC requests to delete this directory using the filesystem endpoint.

module-services/service-desktop/Sync.cpp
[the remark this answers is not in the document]
pawel-mudita: No, it will be different dir, because `taskId` will change: `syncStatus.tempDir = purefs::dir::getTemporaryPath() / syncStatus.taskId;`
If this function fails, we have a filesystem error and we cannot do much more here.

module-services/service-desktop/Sync.cpp
pawel-mudita: Now all files are deleted at the end of the operation.

module-services/service-desktop/Sync.cpp
```diff
@@ -0,0 +1,205 @@
...
+    setvbuf(tarFile.stream, streamBuffer.get(), _IOFBF, streamBufferSize);
+    std::error_code errorCode;
+
+    for (auto &direntry : std::filesystem::directory_iterator(path)) {
+        if (!isValidDirentry(direntry)) {
+            continue;
+        }
+
+        LOG_INFO("Archiving file ...");
+        auto *file = std::fopen(direntry.path().string().c_str(), "r");
```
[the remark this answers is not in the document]
pawel-mudita: Changed to `ifstream`

module-services/service-desktop/Sync.cpp
```diff
@@ -0,0 +1,205 @@
...
+            std::fclose(file);
+            mtar_close(&tarFile);
+            Sync::RemoveSyncDir(path);
+            return false;
+        }
+
+        uintmax_t filesize = std::filesystem::file_size(direntry.path(), errorCode);
+        if (errorCode) {
+            LOG_ERROR("Failed to get size for file %s, error: %d", direntry.path().c_str(), errorCode.value());
+            Sync::RemoveSyncDir(path);
```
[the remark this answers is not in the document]
pawel-mudita: Context cleanup moved to `gsl::finally`

kkleczkowski (approved):

module-services/service-desktop/Sync.cpp
```diff
@@ -0,0 +1,180 @@
...
+        return false;
+    }
+    isTarFileOpen = true;
+
+    auto fileStreamBuffer = std::make_unique<char[]>(purefs::buffer::tar_buf);
+    auto tarStreamBuffer  = std::make_unique<char[]>(purefs::buffer::tar_buf);
+    setvbuf(tarFile.stream, tarStreamBuffer.get(), _IOFBF, purefs::buffer::tar_buf);
+    std::error_code errorCode;
+
+    for (auto &direntry : std::filesystem::directory_iterator(path)) {
```
awulkiew, line 118: Would `const auto &` make sense?

module-services/service-desktop/Sync.cpp
```diff
@@ -0,0 +1,180 @@
...
+            return false;
+        }
+
+        uintmax_t filesize = std::filesystem::file_size(direntry.path(), errorCode);
+        if (errorCode) {
+            LOG_ERROR("Failed to get size for file %s, error: %d", direntry.path().c_str(), errorCode.value());
+            return false;
+        }
+        auto loopCount = (filesize / purefs::buffer::tar_buf) + 1u;
+        uint32_t readSize;
```
awulkiew, line 139: `readSize` is not used outside for loop below so it could be defined inside this loop.

module-services/service-desktop/Sync.cpp
```diff
@@ -0,0 +1,180 @@
...
+            LOG_ERROR("Writing tar header failed");
+            return false;
+        }
+
+        uintmax_t filesize = std::filesystem::file_size(direntry.path(), errorCode);
+        if (errorCode) {
+            LOG_ERROR("Failed to get size for file %s, error: %d", direntry.path().c_str(), errorCode.value());
+            return false;
+        }
+        auto loopCount = (filesize / purefs::buffer::tar_buf) + 1u;
```
awulkiew, line 138: `const`

module-services/service-desktop/Sync.cpp
```diff
@@ -0,0 +1,180 @@
...
+        uint32_t readSize;
+        std::ifstream fileStream;
+        fileStream.open(direntry.path().string().c_str());
```
awulkiew, line 141: Could be written in one line.

module-services/service-desktop/Sync.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+        for (auto i = 0u; i < loopCount; i++) {
+            if (fileStream.bad()) {
+                LOG_ERROR("File stream error.");
+                return false;
+            }
+            if (i + 1u == loopCount) {
+                readSize = filesize % purefs::buffer::tar_buf;
+            }
```
awulkiew, line 150: So this is the code executed only for the last iteration. Neither `filesize` nor `purefs::buffer::tar_buf` change in this loop. What do you think about not checking this and instead simply read `loopCount-1` blocks of size `tar_buf`and one block of size `filesize % tar_buf`?

awulkiew: e.g.:
```
auto readBlock = [&](std::size_t readSize) {
   ...
};
for (auto i = 0u; i < loopCount - 1; i++) {
    readBlock(purefs::buffer::tar_buf);
}
readBlock(filesize % purefs::buffer::tar_buf;
```

module-services/service-desktop/Sync.cpp
awulkiew: Would it make sense to defined it as `uintmax_t` instead of `auto`?

module-services/service-desktop/Sync.cpp
pawel-mudita: Yes, changed

module-services/service-desktop/Sync.cpp
pawel-mudita: I agree about `const`, but I do not see any benefit of changing `auto` to `uintmax_t`.

module-services/service-desktop/Sync.cpp
pawel-mudita: True, changed

module-services/service-desktop/Sync.cpp
awulkiew: I personally prefer to use explicit types if they are simple and non-generic because it gives more information while reading the code. But I'm ok with `auto` if you prefer it.

module-services/service-desktop/Sync.cpp
pawel-mudita: Your suggestion would work, but then I would have to add two additional checks for values returned from lambdas, making one additional check overall.

awulkiew (approved):

PR #4132, opened 2022-11-07
[MOS-790] Workaround for ext4 booting

Lefucjusz:
Workaround required to make MuditaOS
able to boot from ext4-formatted OS
partition.
Previously OS partition was mounted
on /sys and user partition on
/sys/user. This worked
when two separate FS drivers
(FATFS and lwext4) were used.
Changing OS partition to ext4
created an ambiguity whether
/sys/user is user's partition
mountpoint or folder 'user'
on OS partition. This
workaround still may create
ambiguities, so the issue
should be fixed in one of
the further releases by
completely removing 'sys'
prefix.

Make sure that this PR:

### DIFF
--- a/board/linux/libiosyscalls/src/iosyscalls-internal.hpp
+++ b/board/linux/libiosyscalls/src/iosyscalls-internal.hpp
@@ -2,7 +2,6 @@
 // For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 #pragma once
-#include <stddef.h>
 #include <stdio.h>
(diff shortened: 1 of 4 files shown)


SP2FET (commented):
Do we need to adjust scripts that are generating os image?

Lefucjusz:
We don't have to, this workaround only changes the mount point path, the structure of the directories and partition layout stays the same. I've tested it by flashing the image to the Pure and everything seems to be working properly.

SP2FET (approved):

PR #4151, opened 2022-11-22
[BH-1598] Add clock faces with dates.

awulkiew:
Add classic and vertical face with dates for both 12h and 24h formats.

dates

Change placement of battery and time format indicators on ClassicWithAmPm screen while charging.

battery_placement

Fix centering of battery indicator when level is smaller than 100%.

battery_fix

Fix incorrect time format of the alarm in clock face settings.

face_settings

Make sure that this PR:

### DIFF
--- a/products/BellHybrid/apps/common/include/common/layouts/HomeScreenLayoutVerticalWithAmPm.hpp
+++ b/products/BellHybrid/apps/common/include/common/layouts/HomeScreenLayoutVerticalWithAmPm.hpp
@@ -9,7 +9,7 @@ namespace style::homescreen_vertical_ampm
 {
     constexpr inline auto alarm_margin_top   = 83U;
     constexpr inline auto battery_margin_bot = 68U;
-    constexpr inline auto fmt_margin_bot     = 77U;
+    constexpr inline auto ampm_margin_bot    = 77U;
 } // namespace style::homescreen_vertical_ampm
 
 namespace gui
@@ -22,6 +22,6 @@ namespace gui
 
         auto setTime(std::time_t newTime) -> void override;
 
-        TextFixedSize *fmt{};
+        TextFixedSize *ampm{};
     };
 }; // namespace gui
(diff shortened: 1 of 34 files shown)


SP2FET (approved):

products/BellHybrid/apps/common/src/layouts/HomeScreenLayoutVertical.cpp
```diff
@@ -129,6 +135,17 @@ namespace gui
...
 
+    bool HomeScreenLayoutVertical::isAlarmTimeVisibilityAllowed()
+    {
+        return false;
+    }
+
+    bool HomeScreenLayoutVertical::isBatteryCharging(const Store::Battery::State state)
+    {
+        using State = Store::Battery::State;
+        return state == State::Charging or state == State::ChargingDone;
```
Lefucjusz, line 146:
```suggestion
        return (state == State::Charging) || (state == State::ChargingDone);
```

AFAIR in some discussion the team decided to consistently use symbol-like logical operators instead of keyword-like ones in our codebase ;)

Lefucjusz (approved):

products/BellHybrid/apps/common/src/layouts/HomeScreenLayoutVertical.cpp
awulkiew: I agree.

products/BellHybrid/apps/common/include/common/layouts/HomeScreenLayoutVerticalWithAmPm.hpp
```diff
@@ -22,6 +22,6 @@ namespace gui
 
         auto setTime(std::time_t newTime) -> void override;
 
-        TextFixedSize *fmt{};
+        TextFixedSize *ampm{};
```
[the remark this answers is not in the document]
awulkiew: My intention was to improve the current name and at the same time be consistent with other names like `time`, `battery` and `alarm`. If we think that these are not intuitive we should probably consider changing all of them. But I'd prefer to avoid doing this in this PR.

PR #4159, opened 2022-11-24
[MOS-811] Make SMS modem domain in line with voice

marziel:
Unit tests and documentation update will be done later - a distinct ticket for that: MOS-822.

Make sure that this PR:

### DIFF
--- a/module-services/service-cellular/src/VolteHandler.hpp
+++ b/module-services/service-cellular/src/VolteHandler.hpp
@@ -42,18 +42,29 @@ namespace cellular::service
 
             if (enable) {
                 // according to Quectel, this setting doesn't have to be reset when disabling
-                constexpr std::uint8_t voiceDomainPacketSwitchedPreferred = 0x03;
                 auto voiceDomainAnswer =
-                    channel.cmd(factory(at::AT::QNVFW) + "\"/nv/item_files/modem/mmode/voice_domain_pref\"," +
-                                utils::byteToHex<std::uint8_t>(voiceDomainPacketSwitchedPreferred));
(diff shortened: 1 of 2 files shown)


kkleczkowski (approved):

Lefucjusz (approved):

SP2FET (approved):

module-services/service-cellular/src/VolteHandler.hpp
```diff
@@ -103,6 +114,47 @@ namespace cellular::service
...
+            operator std::string()
+            {
+                return "\"/nv/item_files/modem/mmode/" + domain_ + "_domain_pref\"," +
+                       utils::byteToHex<std::uint8_t>(preference_);
+            }
+
+          private:
+            std::string domain_;
+            Type type_;
+            std::uint8_t preference_;
```
awulkiew, line 155: Are there guidelines regarding the names of member variables? I can see many different styles in the code base but haven't seen this one yet. We should probably talk about this and define a clear standard.

awulkiew (approved):

module-services/service-cellular/src/VolteHandler.hpp
marziel: That's right. This one probably already exists in some pieces of code written by me. It's clearer than `mSomething` and the underscore is trailing because a leading one can mean an UB in some cases.

marziel: @awulkiew So, for now I'm closing this conversation - please arrange a meeting or a Slack discussion as this topic is recurring sometimes.

PR #4165, opened 2022-11-28
[MOS-827] Fix looping on the SIM card selection screen

Maciej-Mudita:
It was not possible to finish onboarding properly
when we selected a SIM card without an active PIN

Make sure that this PR:

### DIFF
--- a/module-apps/application-onboarding/ApplicationOnBoarding.cpp
+++ b/module-apps/application-onboarding/ApplicationOnBoarding.cpp
@@ -19,9 +19,11 @@
 
 #include <application-settings/windows/system/ChangeTimeZone.hpp>
 #include <apps-common/locks/data/PhoneLockMessages.hpp>
+#include <apps-common/locks/data/SimLockMessages.hpp>
 #include <service-appmgr/Constants.hpp>
 #include <service-appmgr/messages/GetCurrentDisplayLanguageResponse.hpp>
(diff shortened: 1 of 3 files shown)


module-apps/application-onboarding/ApplicationOnBoarding.cpp
```diff
@@ -93,8 +95,17 @@ namespace app
             return sys::msgHandled();
         });
 
+        connect(typeid(locks::SimSwitched), [&](sys::Message *msg) {
+            auto simState = Store::GSM::get()->sim;
+            if (simState == Store::GSM::SIM::SIM1 || simState == Store::GSM::SIM::SIM2) {
+                phoneLockSubject.setPhoneLock();
+                return sys::msgHandled();
+            }
+            return sys::msgNotHandled();
```
SP2FET, line 104: shouldn't the sim handling be rewritten to don't let the message with no sim arrived?

Maciej-Mudita: In this case, it won't change anything. And rewriting the logic will not be so easy because when we do not have the "no SIM" status, pop-ups with errors from the modem will appear at the start if the user does not have any card inserted.

SP2FET: alright then

SP2FET (approved):

marziel (approved):

PR #4171, opened 2022-11-30
Mos 803 backup restore

swichu91:
* DeviceInfo endpoint update
* Added backup/restore scripts

Make sure that this PR:


image/user/boot.json
```diff
@@ -3,12 +3,12 @@
         "slots": {
             "a": {
                 "bootable": true,
-                "successful": false,
+                "successful": true,
```
SP2FET, line 6: IMO it should be set to false - first successful boot will make this `true` - in case of bootloop after flashing

SP2FET (approved):

image/user/boot.json
swichu91: That depends: if we assume that there is a possibility we ship broken image to the factory, then yes, you're right.

swichu91: Let's revert this change as the previous version provides a bit more safety.

PR #4178, opened 2022-12-02
[BH-1415] Front light intensity for pre-wake up and main alarm

dawidwojtas-mudita:
The user can manage the intensity
of the front light during the pre-wake up
and main alarm. The intensity can be selected
from a 10-degree scale, where 1 is the lightest
front light and 10 the brightest.

Make sure that this PR:


products/BellHybrid/alarms/src/actions/FrontlightAction.cpp
```diff
@@ -149,14 +161,22 @@ namespace alarms
     screen_light_control::LinearProgressModeParameters LinearProgressFrontlightAction::prepareParameters()
     {
         constexpr auto firstTargetDuration = std::chrono::seconds{5};
+
+        std::string brightnessString =
+            settings.getValue(bell::settings::PrewakeUp::brightness, settings::SettingsScope::Global);
+        float brightness              = 50.0f;
+        if (!brightnessString.empty()) {
+            brightness = frontlight_utils::fixedValToPercentage(std::stoi(brightnessString));
+        }
```
awulkiew, line 170: You could consider putting this inside a function since it's repeated in `ManualFrontlightAction`.

products/BellHybrid/apps/application-bell-settings/models/alarm_settings/AbstractPrewakeUpSettingsModel.hpp
```diff
@@ -45,10 +49,16 @@ namespace app::bell_settings
             return *lightDuration;
         }
 
+        gui::AbstractSettingsModel<frontlight_utils::Brightness> &getBrightness()
```
awulkiew, line 52: I'd expect that a function called `getBrightness` returns brightness value. Shouldn't this function rather be called `getXXXModel` e.g. `getBrightnessModel`?

products/BellHybrid/apps/application-bell-settings/models/alarm_settings/AlarmSettingsListItemProvider.cpp
```diff
@@ -69,8 +73,44 @@ namespace app::bell_settings
 
         internalData.emplace_back(alarmVolume);
 
-        internalData.emplace_back(
-            new OnOffListItem(model.getAlarmLightOnOff(), utils::translate("app_bell_settings_alarm_settings_light")));
+        auto onOffLight = new OnOffListItem(settingsModel.getAlarmLightOnOff(),
+                                            utils::translate("app_bell_settings_alarm_settings_light"));
+
+        internalData.emplace_back(onOffLight);
```
awulkiew, line 79: This is not exception-safe but since when an exception is thrown the device is restarted anyway I guess it may be left as it is.

Another thing is that I don't like the fact that `internalData` is modified here directly when this is a container of pointers managed by one of the base classes. It's because someone may do anything with this container, e.g. remove some pointers and the objects will not be destroyed when base class is destroyed. But changing it is probably outside of the scope of this PR so I'm ok with leaving it as it is.

awulkiew (approved):

products/BellHybrid/alarms/src/actions/FrontlightAction.cpp
```diff
@@ -103,11 +106,25 @@ namespace alarms
...
 
+    screen_light_control::ManualModeParameters ManualFrontlightAction::prepareParameters()
+    {
+        std::string brightnessString =
+            settings.getValue(bell::settings::Alarm::brightness, settings::SettingsScope::Global);
+        float brightness = 50.0f;
+        if (!brightnessString.empty()) {
+            brightness = frontlight_utils::fixedValToPercentage(std::stoi(brightnessString));
+        }
+        return screen_light_control::ManualModeParameters{.manualModeBrightness = brightness};
```
marziel, line 121: Please avoid designated _initializers_: they're perfectly allowed only in C (since C99) or since C++20.

products/BellHybrid/apps/common/include/common/data/FrontlightUtils.hpp
```diff
@@ -0,0 +1,36 @@
...
+        static constexpr auto minimumLightOnPercentOffsetValue = 16.0f;
+        static constexpr auto minBrightness                    = 1U;
+        static constexpr auto maxBrightness                    = 10U;
+        static constexpr float multiplier = (maxPercent - minimumLightOnPercentOffsetValue) / maxBrightness;
+    } // namespace
+
+    /// 1-10 range
+    using Brightness = std::uint8_t;
+
+    static inline float fixedValToPercentage(Brightness value)
```
marziel, line 24: Maybe those two functions could be simplified with `std::clamp()` or something alike?

products/BellHybrid/alarms/src/actions/FrontlightAction.cpp
dawidwojtas-mudita: I just simplified this repeated code.

products/BellHybrid/alarms/src/actions/FrontlightAction.cpp
dawidwojtas-mudita: I see that GCC supports `designated initializers` in C++17. However, it is not part of the standard in C++17 so I removed it from the code. :)

products/BellHybrid/apps/application-bell-settings/models/alarm_settings/AlarmSettingsListItemProvider.cpp
dawidwojtas-mudita: I totally agree with you. However, I tried to follow the old base of code and don't change a lot. 
As you said before - it is not part of this PR, but definitely, it should be addressed and changed in the feature.
On the other hand, almost every application on our system has an MVC pattern so we need to redesign every `provider`.

products/BellHybrid/apps/common/include/common/data/FrontlightUtils.hpp
dawidwojtas-mudita: In this case `std::clamp()` is not useful at all, but thanks for your advice :) I think that `std::min` and `std::max` is fine here so I'm leaving it as it is.

PR #4186, opened 2022-12-06
Release 1.8.0 bell

awulkiew:

Make sure that this PR:


tools/generate_update_image.sh
```diff
@@ -67,11 +67,31 @@ function linkInStageing(){
...
+    pushd bell 1> /dev/null
+    mkdir bg_sounds
+    pushd bg_sounds 1> /dev/null
+    ln -s ../../../../../sysroot/sys/current/assets/audio/bg_sounds/Ancient_Greek.mp3
+    ln -s ../../../../../sysroot/sys/current/assets/audio/bg_sounds/Woodland_Ambiance.mp3
+    popd 1> /dev/null # bg_sounds
+    popd 1> /dev/null # bell
+    popd 1> /dev/null # audio
+
+    popd 1> /dev/null # assets
```
SP2FET, line 94: please describe those changes in some document at it will affect UDM

SP2FET (changes requested):

tools/generate_update_image.sh
awulkiew: This is done only in Harmony release branch and should not be done when UDM is implemented. After UDM is implemented bell directory should be deleted and mp3 files from new directory should be indexed after the update.

SP2FET: so please write it down - what has to be done during udm version ;)

awulkiew: Sure, my point was that UDM is unrelated to this change so this is not the correct place to discuss this. ;)

SP2FET (approved):

dawidwojtas-mudita (approved):

PR #4242, opened 2023-01-02
[MOS-812] Improved CUSD message handling

marziel:
* Added handling for shorter +CUSD formats, thus got rid of a crash.
* Added handling for the "push session"+"action required" state of a USSD interaction with the network.

Make sure that this PR:

### DIFF
--- a/module-cellular/at/UrcCusd.hpp
+++ b/module-cellular/at/UrcCusd.hpp
@@ -3,22 +3,24 @@
 
 #pragma once
 
-#include "Urc.hpp"
-
 #include <optional>
+#include "Urc.hpp"
 
 namespace at::urc
 {
-
     class Cusd : public Urc
     {
-
         enum Tokens
         {
             Status,
             Response,
             DCS
         };
+
+        bool valid_ = true;
+
+        auto split(const std::string &str) -> void override;
+        [[nodiscard]] auto getDCS() const noexcept -> std::optional<int>;
 
       public:
         enum class StatusType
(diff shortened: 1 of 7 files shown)


module-cellular/test/unittest_URC.cpp
```diff
@@ -279,78 +279,6 @@ TEST_CASE("+Qind: FOTA")
     }
 }
 
-TEST_CASE("+Cusd")
```
SP2FET, line 282: maybe it's worth adding some test cases for the new implementation? especially that you're using regex :P

marziel: Yeah, I'm already working on them...

marziel: Decided to temporarily withdraw the UTs, see https://appnroll.atlassian.net/browse/MOS-812?focusedCommentId=69817.

module-cellular/at/UrcCusd.hpp
```diff
@@ -30,26 +32,17 @@ namespace at::urc
...
         static constexpr std::string_view head = "+CUSD";
-        static auto isURC(const std::string &uHead) -> bool
-        {
-            return uHead.find(Cusd::head) != std::string::npos;
-        }
 
-        using Urc::Urc;
+        explicit Cusd(const std::string &urcBody, const std::string &urcHead = {});
 
+        static auto isURC(const std::string &uHead) -> bool;
```
rrandomsky, line 40: Sugestion:
make method  `const`
`static auto isURC(const std::string &uHead) const -> bool;`

marziel: A `static` method can't be `const` at the same time as it isn't bound to any instance of this class.

kkleczkowski (approved):

module-services/service-cellular/call/tests/CMakeLists.txt
```diff
@@ -10,6 +10,7 @@ add_catch2_executable(
         sml::sml
         sml::utils::logger
         test::fakeit
+	libphonenumber
```
marziel, line 13: Added to fix linkage problems caused by messy dependencies

PR #4258, opened 2023-01-10
[MOS-872] Add support for directories in filesystem endpoint's listdir

SP2FET:
Added support for listing directories with nested ones

Make sure that this PR:

### DIFF
--- a/module-services/service-desktop/endpoints/filesystem/FS_Helper.cpp
+++ b/module-services/service-desktop/endpoints/filesystem/FS_Helper.cpp
@@ -1,23 +1,52 @@
-// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
+// Copyright (c) 2017-2023, Mudita Sp. z.o.o. All rights reserved.
 // For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 #include <endpoints/filesystem/FS_Helper.hpp>
-#include <endpoints/Context.hpp>
-#include <service-desktop/DesktopMessages.hpp>
 #include <service-desktop/ServiceDesktop.hpp>
-#include <endpoints/JsonKeyNames.hpp>
 #include <endpoints/message/Sender.hpp>
-#include <purefs/filesystem_paths.hpp>
 
 #include <sys/statvfs.h>
-#include <filesystem>
 
 namespace sdesktop::endpoints
 {
     namespace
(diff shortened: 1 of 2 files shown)


swichu91 (approved):
Approved but please fix minor issues that we've discussed on priv conversation.

module-services/service-desktop/endpoints/filesystem/FS_Helper.cpp
```diff
@@ -1,23 +1,52 @@
...
+        {
+            directory,
+            regularFile,
+            symlink,
+            other,
+        };
+
+        auto parseFileEntry(const std::filesystem::directory_entry &entry) -> json11::Json
+        {
+            int size = 0;
```
swichu91, line 26: `std::size_t` or in this case `uintmax_t` would be better in my opinion.

module-services/service-desktop/endpoints/filesystem/FS_Helper.cpp
```diff
@@ -329,28 +357,21 @@ namespace sdesktop::endpoints
...
+        json11::Json::object const response({{directory, jsonArr}});
         return ResponseContext{.status = http::Code::OK, .body = response};
     }
 
     auto FS_Helper::getFreeSpaceForUserFilesMiB() const -> unsigned long
     {
         const auto userDiskPath = purefs::dir::getUserDiskPath();
 
-        std::unique_ptr<struct statvfs> vfstat = std::make_unique<struct statvfs>();
+        auto vfstat = std::make_unique<struct statvfs>();
```
swichu91, line 374: How about simple allocation on stack? `struct vfs` should easily fit local stack :)

module-services/service-desktop/endpoints/filesystem/FS_Helper.cpp
SP2FET: yes, but JSON (our library) does not support anything larger than int, so it will be either way narrowed down to int

module-services/service-desktop/endpoints/filesystem/FS_Helper.cpp
SP2FET: I can check that, but we'll have to test thoroughly the whole feature :)

Maciej-Mudita (approved):

Lefucjusz (approved):

PR #4259, opened 2023-01-10
[CP-1579] Fix correct password inserted during Pure boot is not accepted

lukasmas:
Fix the issue when the USB cable connected to the device during boot up and Center requested a passcode before the device was fully up. What caused a 'Wrong passcode' message after typing a passcode via Center.

Make sure that this PR:

### DIFF
--- a/module-services/service-desktop/ServiceDesktop.cpp
+++ b/module-services/service-desktop/ServiceDesktop.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
+// Copyright (c) 2017-2023, Mudita Sp. z.o.o. All rights reserved.
 // For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 #include <service-desktop/ServiceDesktop.hpp>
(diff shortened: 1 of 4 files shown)


module-services/service-desktop/ServiceDesktop.cpp
```diff
@@ -195,6 +200,7 @@ auto ServiceDesktop::usbWorkerInit() -> sys::ReturnCodes
     auto caseColour   = getCaseColour();
 
     LOG_DEBUG("usbWorkerInit Serial Number: %s, Case Colour: %s", serialNumber.c_str(), caseColour.c_str());
+    vTaskDelay(3000);
```
Maciej-Mudita, line 203: How will the system behave when we quickly connect and disconnect the USB plug?

module-services/service-desktop/ServiceDesktop.cpp
```diff
@@ -52,7 +57,7 @@ sys::ReturnCodes ServiceDesktop::InitHandler()
     connectHandler<sdesktop::usb::USBDisconnected>();
     connectHandler<sevm::USBPlugEvent>();
 
-    checkChargingCondition();
+    checkChargingConditionTimer.start();
```
Maciej-Mudita, line 60: Please consider instead of the initialization delay the response delay in OS -> Center communication

module-services/service-desktop/ServiceDesktop.cpp
lukasmas: This delay shouldn't be there

module-services/service-desktop/ServiceDesktop.cpp
lukasmas: Good idea, I will check it

Maciej-Mudita (approved):

SP2FET (approved):

PR #4269, opened 2023-01-17
[MOS-767] Fix full path displayed for invalid audio file

Lefucjusz:
Fix of the issue that whole file path was
displayed instead of just filename in music 
player for invalid audio file for which Taglib 
failed to obtain tags.

Make sure that this PR:

### DIFF
--- a/module-audio/tags_fetcher/TagsFetcher.cpp
+++ b/module-audio/tags_fetcher/TagsFetcher.cpp
@@ -1,79 +1,76 @@
-// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
+// Copyright (c) 2017-2023, Mudita Sp. z.o.o. All rights reserved.
 // For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 #include "TagsFetcher.hpp"
-
-#include <gsl/util>
 #include <Utils.hpp>
-
 #include "fileref.h"
-#include "tag.h"
-#include "tfilestream.h"
 
 namespace tags::fetcher
 {
-    std::optional<Tags> fetchTagsInternal(std::string filePath)
+    namespace
     {
-        TagLib::FileRef const tagReader(filePath.c_str());
-        if (!tagReader.isNull() && (tagReader.tag() != nullptr)) {
-            TagLib::Tag *tags                   = tagReader.tag();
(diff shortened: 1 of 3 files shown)


module-audio/tags_fetcher/TagsFetcher.cpp
```diff
@@ -1,79 +1,76 @@
...
     {
-        TagLib::FileRef const tagReader(filePath.c_str());
-        if (!tagReader.isNull() && (tagReader.tag() != nullptr)) {
-            TagLib::Tag *tags                   = tagReader.tag();
-            TagLib::AudioProperties *properties = tagReader.audioProperties();
+        std::string getTitleFromFilePath(const std::string &path)
+        {
+            const auto pos = path.rfind('/');
+            if (pos != std::string::npos) {
+                return &path[pos + 1];
```
marziel, line 16: Please either use the `c_str()` string method or `substr()`

module-audio/tags_fetcher/TagsFetcher.cpp
```diff
@@ -1,79 +1,76 @@
...
-    std::optional<Tags> fetchTagsInternal(std::string filePath)
+    namespace
     {
-        TagLib::FileRef const tagReader(filePath.c_str());
-        if (!tagReader.isNull() && (tagReader.tag() != nullptr)) {
-            TagLib::Tag *tags                   = tagReader.tag();
-            TagLib::AudioProperties *properties = tagReader.audioProperties();
+        std::string getTitleFromFilePath(const std::string &path)
+        {
+            const auto pos = path.rfind('/');
```
marziel, line 14: Is this always an absolute path and thus always contains a `/`?

module-audio/tags_fetcher/TagsFetcher.cpp
Lefucjusz: Fixed

module-audio/tags_fetcher/TagsFetcher.cpp
Lefucjusz: It seems so, but the case when `/` is not found is also handled, whole content of `path` will be displayed then.

marziel: Right, I forgot about it.

marziel (approved):

PR #4298, opened 2023-01-26
[MOS-174] Fix broken events counter for 99+ events

Lefucjusz:
Fix of the issue that events counter
was displayed incorrectly for more
than 99 events.

Make sure that this PR:

### DIFF
--- a/module-apps/apps-common/notifications/NotificationListItem.cpp
+++ b/module-apps/apps-common/notifications/NotificationListItem.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2021, Mudita Sp. z.o.o. All rights reserved.
+// Copyright (c) 2017-2023, Mudita Sp. z.o.o. All rights reserved.
 // For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 #include "NotificationListItem.hpp"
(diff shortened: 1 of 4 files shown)


Maciej-Mudita (approved):

module-apps/apps-common/notifications/NotificationListItem.cpp
```diff
@@ -57,30 +58,35 @@ namespace
...
     }
 
-    constexpr auto maxNotificationValue = "99+";
-    constexpr auto singleNotification   = "1";
+    constexpr auto singleNotification             = 1;
+    constexpr auto maxNotificationValue           = 99;
+    constexpr auto maxNotificationReplacementText = "99+";
 
-    auto buildNotificationCountText(const UTF8 &indicator) -> gui::Text *
+    auto buildNotificationCountText(unsigned count) -> gui::Text *
```
SP2FET, line 65: consider using fixed-size type

SP2FET (approved):

marziel (approved):

module-apps/apps-common/notifications/NotificationListItem.cpp
Lefucjusz: The only place this method is used is in the `NotificationWithEventCounter` constructor, which gets its argument only from `getValue()` method from `NotificationData.cpp`. That method explicitly returns `unsigned` type, so I just used the same type here.

PR #4315, opened 2023-02-02
[MOS-578] Fix incorrect logic with SMS notifications

SP2FET:
Fixed case, when there's only one thread unread and there's no info about the thread, only a generic message

Tests are deleted because the fix would require rewritting ALL
of those tests - those will be written with new implementation
of notification table (MOS-899)

Make sure that this PR:


module-db/Interface/NotificationsRecord.hpp
```diff
@@ -84,6 +84,7 @@ class NotificationsRecordInterface : public RecordInterface<NotificationsRecord,
   private:
     NotificationsDB *notificationsDb   = nullptr;
     ContactRecordInterface *contactsDb = nullptr;
+    std::vector<uint32_t> unreadMsgNumberIDs;
```
Maciej-Mudita, line 87:
```suggestion
    std::vector<std::uint32_t> unreadMsgNumberIDs;
```

module-db/Interface/NotificationsRecord.hpp
```diff
@@ -98,5 +99,5 @@ class NotificationsRecordInterface : public RecordInterface<NotificationsRecord,
     [[nodiscard]] bool processIncrement(NotificationsRecord::Key key,
                                         std::optional<utils::PhoneNumber::View> &&number,
                                         size_t size);
-    [[nodiscard]] bool processDecrement(NotificationsRecord::Key key, size_t size);
+    [[nodiscard]] bool processDecrement(NotificationsRecord::Key key, uint32_t numberID, size_t size);
```
Maciej-Mudita, line 102:
```suggestion
    [[nodiscard]] bool processDecrement(NotificationsRecord::Key key, std::uint32_t numberID, size_t size);
```

module-db/Interface/NotificationsRecord.cpp
```diff
@@ -256,17 +261,32 @@ bool NotificationsRecordInterface::processIncrement(NotificationsRecord::Key key
     return ret;
 }
 
-bool NotificationsRecordInterface::processDecrement(NotificationsRecord::Key key, size_t delta)
+bool NotificationsRecordInterface::processDecrement(const NotificationsRecord::Key key,
+                                                    const uint32_t numberID,
```
Maciej-Mudita, line 265:
```suggestion
                                                    const std::uint32_t numberID,
```

Maciej-Mudita (approved):

module-db/Interface/NotificationsRecord.hpp
SP2FET: fixed

module-db/Interface/NotificationsRecord.hpp
SP2FET: fixed

module-db/Interface/NotificationsRecord.cpp
SP2FET: fixed

Lefucjusz (approved):

PR #4317, opened 2023-02-03
[MOS-196] Add days of the week to the list of SMS, calls and notes

Maciej-Mudita:
If an event (e.g. receiving a call or SMS) happened in the current week, we display the days of the week instead of the date on the list.

Make sure that this PR:

### DIFF
--- a/module-utils/time/test/unittest_time.cpp
+++ b/module-utils/time/test/unittest_time.cpp
@@ -41,6 +41,10 @@ const std::regex regexDDMMYYYY("^([0-2]\\d|3[0-1])\\.(0[1-9]|1[0-2])\\.\\d{4}$")
 const std::regex regexDDMM("^([0-2]\\d|3[0-1])\\.(0[1-9]|1[0-2])$");
 const std::regex regexMMDDYYYY("^(0[1-9]|1[0-2])\\.([0-2]\\d|3[0-1])\\.\\d{4}$");
 const std::regex regexMMDD("^(0[1-9]|1[0-2])\\.([0-2]\\d|3[0-1])$");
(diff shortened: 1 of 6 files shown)


module-utils/time/test/unittest_time.cpp
```diff
@@ -342,6 +346,55 @@ TEST_CASE("DateTime")
...
+            auto newTimeTimeinfo = currentTimeTimeinfo;
+            SECTION("different day")
+            {
+                auto i = GENERATE(range(7, 32));
+                newTimeTimeinfo.tm_mday -= i;
+                auto newTime = std::mktime(&newTimeTimeinfo);
+                DateTime datetime(timeSettings, newTime, currentTime);
+                REQUIRE(!datetime.isCurrentWeek());
+            }
+            SECTION("different month")
```
marziel, line 371: Test names like this don't tell much about what's the test supposed to do :( I can see here we have generating months January-December and... what happens next? I mean the test name should tell what's checked without diving into the code under test

Maciej-Mudita: If you look at the whole test, can't you see that we change the month in the date to test whether the isCurrentWeek function works correctly?
```
TEST_CASE("DateTime")
...
	SECTION("isCurrentWeek")
	...
		SECTION("not current week")
		...
			SECTION("different month")
			...
```

marziel: No :D Anyway, I won't insist - leave it as is if you prefer

Maciej-Mudita: Fixed :)

SP2FET (approved):

Lefucjusz (approved):

PR #4319, opened 2023-02-03
[MOS-871] Loudspeaker icon not updated when audio routing has changed

lukasmas:

Make sure that this PR:

### DIFF
--- a/module-services/service-audio/AudioServiceAPI.cpp
+++ b/module-services/service-audio/AudioServiceAPI.cpp
@@ -1,4 +1,4 @@
-﻿// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
+﻿// Copyright (c) 2017-2023, Mudita Sp. z.o.o. All rights reserved.
 // For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 #include <AudioMessage.hpp>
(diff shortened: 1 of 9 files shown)


Lefucjusz (approved):

module-services/service-audio/AudioServiceAPI.cpp
```diff
@@ -90,16 +90,18 @@ namespace AudioServiceAPI
...
+        // return true;
     }
 
-    bool SendEvent(sys::Service *serv, audio::EventType eType, audio::Event::DeviceState state)
+    void SendEvent(sys::Service *serv, audio::EventType eType, audio::Event::DeviceState state)
     {
         auto msg = std::make_shared<AudioEventRequest>(eType, state);
-        return serv->bus.sendUnicast(msg, service::name::audio);
+        serv->bus.sendMulticast(msg, sys::BusChannel::ServiceAudioNotifications);
+        // return true;
```
SP2FET, line 104: why did you change it to void?

module-services/service-audio/ServiceAudio.cpp
```diff
@@ -492,7 +497,7 @@ std::unique_ptr<AudioResponseMessage> ServiceAudio::HandleSendEvent(std::shared_
         input.audio->SendEvent(evt);
     }
 
-    return std::make_unique<AudioEventResponse>(RetCode::Success);
+    return sys::msgHandled();
```
SP2FET, line 500: doesn't it break anything?

module-services/service-audio/AudioServiceAPI.cpp
SP2FET: nevermind - I didn't noticed the multicast

SP2FET (approved):

module-services/service-audio/AudioServiceAPI.cpp
lukasmas: Two reasons:
1. It is not checked anywhere what was the outcome of this function
2. Multicast doesn't return anything, so I cannot check if the message was actually sent or not

module-services/service-audio/ServiceAudio.cpp
lukasmas: seems not. 
This message is now multicast, so there is no need to send a response to a sender. Also, I've checked and multicast messages are handled like this.

PR #4332, opened 2023-02-09
[MOS-871] Loudspeaker icon is not updated when audio routing has changed

lukasmas:
Fix for the loudspeaker becomes active again when a headset is unplugged if it was on while the headset was connected to the phone.

Make sure that this PR:

### DIFF
--- a/module-apps/application-call/ApplicationCall.cpp
+++ b/module-apps/application-call/ApplicationCall.cpp
@@ -6,6 +6,7 @@
 #include "CallWindow.hpp"
 #include "EmergencyCallWindow.hpp"
 #include "EnterNumberWindow.hpp"
+#include "presenter/CallPresenter.hpp"
 
 #include <apps-common/messages/DialogMetadataMessage.hpp>
 #include <apps-common/windows/Dialog.hpp>
@@ -19,6 +20,7 @@
 #include <service-cellular/CellularServiceAPI.hpp>
 #include <time/time_conversion.hpp>
 #include <WindowsPopupFilter.hpp>
+#include <service-audio/AudioServiceAPI.hpp>
 
 #include <cassert>
 #include <memory>
@@ -114,7 +116,8 @@ namespace app
             return actionHandled();
         });
 
-        callModel = std::make_shared<app::call::CallModel>(this);
+        callModel     = std::make_shared<app::call::CallModel>(this);
+        callPresenter = std::make_unique<call::CallWindowContract::Presenter>(this->callModel);
     }
 
     bool ApplicationCall::conditionalReturnToPreviousView()
@@ -181,9 +184,9 @@ namespace app
             return sys::MessageNone{};
         });
 
-        connect(typeid(AudioEventRequest), [&](sys::Message *request) {
-            auto message = static_cast<AudioEventRequest *>(request);
-            return handleAudioMessageEvent(message);
+        connect(typeid(AudioRoutingNotification), [this](sys::Message *request) {
+            auto message = static_cast<AudioRoutingNotification *>(request);
+            return handleRoutingNotification(message);
         });
 
         createUserInterface();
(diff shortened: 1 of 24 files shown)


module-apps/application-call/windows/CallWindow.hpp
```diff
@@ -29,6 +29,14 @@ namespace gui
 
         std::unique_ptr<app::call::CallWindowContract::Presenter> presenter;
 
+        std::set<audio::EventType> devices_connected{};
+        bool isLouderSpeakerON = false;
```
Lefucjusz, line 33:
```suggestion
        bool isLoudspeakerOn = false;
```

Why not just like that?

Lefucjusz (approved):

module-apps/application-call/windows/CallWindow.cpp
```diff
@@ -299,34 +299,61 @@ namespace gui
...
             }
             break;
         }
+        case audio::EventType::CallLoudspeakerOn: {
+            isLouderSpeakerON = true;
+            break;
+        }
+        case audio::EventType::CallLoudspeakerOff: {
+            isLouderSpeakerON = false;
+            break;
```
swichu91, line 321: These cases should be passed to the presenter and handled there too:) Try to avoid adding boolean variables holding different kinds of state, and instead implement logic using proper state machines or use already existing mechanisms to do the job:)

module-apps/application-call/windows/CallWindow.cpp
```diff
@@ -299,34 +299,61 @@ namespace gui
...
+        devices_connected.erase(eventType);
+        if (not devices_connected.empty()) {
+            turnOffSpeakerIconIfNeeded();
+        }
+        else {
+            turnOnSpeakerIconIfNeeded();
+        }
+    }
+
+    void CallWindow::turnOffSpeakerIconIfNeeded()
```
swichu91, line 346: Same as above, ought to be implemented inside presenter, not in the window class. By doing this, we are following fundamental rule of SRP(Single responsibility rule), i.e. window(view) is responsible only for displaying data, presenter is responsible for business logic and orchestration between models and view and so on.

swichu91 (commented):
Please check comments.

kkleczkowski (approved):

swichu91 (approved):

module-apps/application-call/ApplicationCall.cpp
```diff
@@ -6,6 +6,7 @@
 #include "CallWindow.hpp"
```
lukasmas, line 6: One case still needs to be fixed. Speaker Icon gets lost when Bluetooth headphones are connected before the call starts.

module-apps/application-call/windows/CallWindow.cpp
```diff
@@ -297,36 +301,14 @@ namespace gui
...
-                if (not devices_connected.empty()) {
-                    changeSpeakerIconIfNeeded();
-                }
-            }
-            break;
-        }
-        default: {
-            break;
-        }
-        }
```
swichu91, line 322: Is this logic not needed anymore?

module-apps/application-call/ApplicationCall.cpp
```diff
@@ -181,10 +184,7 @@ namespace app
             return sys::MessageNone{};
         });
 
-        connect(typeid(AudioEventRequest), [&](sys::Message *request) {
-            auto message = static_cast<AudioEventRequest *>(request);
-            return handleAudioMessageEvent(message);
-        });
+        connect(typeid(AudioEventRequest), [&](sys::Message *request) { return handleAudioMessageEvent(); });
```
swichu91, line 187:
```suggestion
        connect(typeid(AudioEventRequest), [this](sys::Message *request) { return handleAudioMessageEvent(); });
```
Try to capture only what you need/use.

module-apps/application-call/ApplicationCall.cpp
```diff
@@ -281,10 +280,11 @@ namespace app
         }
     }
 
-    sys::MessagePointer ApplicationCall::handleAudioMessageEvent(AudioEventRequest *message)
+    sys::MessagePointer ApplicationCall::handleAudioMessageEvent()
```
swichu91, line 283: It would be much better if we have a specific AudioEvent that correctly notifies us about audio profile changes.

module-apps/application-call/ApplicationCall.cpp
lukasmas: fixed

PR #4335, opened 2023-02-10
[MOS-903] Remove reboot to USB MSC mode

Lefucjusz:
Remove mechanics enabling to perform reboot
to MSC mode by sending request to one of the
endpoints via serial port, as it is not
implemented in bootloader and may potentially
be unsafe.

Additionally added commit with listdir support
in harness repository that was previously
omitted by mistake.

Make sure that this PR:


module-sys/SystemManager/SystemManagerCommon.cpp
```diff
@@ -597,9 +578,13 @@ namespace sys
             return MessageNone{};
         });
 
+        /* Workaround for hardware issues with power supply. With USB cable connected the phone cannot
+         * restart, so detect such state and perform reboot when right function button is pressed.
+         * This way for the user the behavior looks as if the phone was normally turned on from shutdown
+         * state. */
```
swichu91, line 584: 👍

module-sys/SystemManager/SystemManagerCommon.cpp
```diff
@@ -597,9 +578,13 @@ namespace sys
...
         connect(sevm::KbdMessage(), [&](Message *) {
-            // we are in shutdown mode - we received that there was red key pressed -> we need to reboot
-            if (state == State::Shutdown) {
+            if ((state == State::Shutdown) && (Store::Battery::get().state != Store::Battery::State::Discharging)) {
```
swichu91, line 586: Is it possible to be in such a state where USB cable is plugged and system is reporting 'Discharging' status? The obvious one that comes to mind is some kind of error condition but is there anything else?

swichu91: Apart from that, `Store::Battery::get()` is something I would like to discourage from using it. At the same time, I'm aware this concept is being used throughout the system(sadly).

Lefucjusz: My bad, it was a change I left there by mistake while working on some other issue, removed it now ;)

swichu91 (approved):

SP2FET (approved):

PR #4336, opened 2023-02-10
[MOS-879] Fixed misleding SMS flow

kkleczkowski:
Fixed misleading SMS flow after sending error.
When modem was rebooting attempt to send SMS was causing Rejected by no sim popup, which was misleading for user.

Make sure that this PR:


swichu91 (approved):
I wonder if there is a procedure for resending SMSes in cases when a user wanted to send some while GSM modem was restarting?

Maciej-Mudita (approved):

PR #4337, opened 2023-02-10
[MOS-357] Broken navigation around contact adding

marziel:
From now on, after cancelling or confirming a new contact creation initiated from the call log, the phone goes back to correct windows.

Make sure that this PR:

### DIFF
--- a/module-apps/application-phonebook/windows/PhonebookNewContact.cpp
+++ b/module-apps/application-phonebook/windows/PhonebookNewContact.cpp
@@ -5,6 +5,7 @@
 
 #include "DialogMetadata.hpp"
 #include "application-phonebook/ApplicationPhonebook.hpp"
+#include <service-appmgr/Controller.hpp>
 
 #include <Dialog.hpp>
 #include <service-db/DBServiceAPI.hpp>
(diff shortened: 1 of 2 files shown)


module-apps/application-phonebook/windows/PhonebookNewContact.cpp
```diff
@@ -119,6 +120,16 @@ namespace gui
 
     auto PhonebookNewContact::onInput(const InputEvent &inputEvent) -> bool
     {
+        auto backToCallLogIfCameFromThere = [this]() {
+            // MOS-357: conditionally (@see ApplicationManagerCommon::handleSwitchBack) return to call log
+            return app::manager::Controller::switchBack(
+                application, std::make_unique<app::manager::SwitchBackRequest>("ApplicationCallLog"));
+        };
```
swichu91, line 127: Do we really want to mark this comment with PR number? 
How about:
```suggestion
        // Conditionally (@see ApplicationManagerCommon::handleSwitchBack) returns to the call log application
        auto backToCallLogIfCameFromThere = [this]() {
            return app::manager::Controller::switchBack(
                application, std::make_unique<app::manager::SwitchBackRequest>("ApplicationCallLog"));
        };
```

swichu91 (approved):

module-apps/application-phonebook/windows/PhonebookNewContact.cpp
marziel: Not a PR number, but a ticket number :) IMO this piece of code may seem weird at first glance, so I wanted to introduce a direct reference for the reader.

awulkiew (approved):

PR #4364, opened 2023-02-23
[MOS-766] Change factory entries initialization routine

Lefucjusz:
Change of the algorithm used to initialize
settings database with values read from EEPROM,
previously first insertion failure resulted
in stopping the whole process, now next values
are inserted even if previous has failed.
Additionally minor cleanup.

Make sure that this PR:

### DIFF
--- a/module-db/Database/Field.hpp
+++ b/module-db/Database/Field.hpp
@@ -1,4 +1,4 @@
-﻿// Copyright (c) 2017-2021, Mudita Sp. z.o.o. All rights reserved.
+﻿// Copyright (c) 2017-2023, Mudita Sp. z.o.o. All rights reserved.
 // For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 #pragma once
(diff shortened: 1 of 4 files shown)


module-db/Database/Field.hpp
```diff
@@ -14,22 +14,16 @@ class Field
...
 
     ~Field()
     {}
 
     const char *getCString() const;
-    std::string getString() const
-    {
-        return mValue;
-    }
+    std::string getString() const;
```
awulkiew, line 26: I'm wondering. What is the lifetime of the returned object WRT the lifetime of the Field?
If the lifetime of the returned object can be longer then we have to copy here. But then `const char *getCString()` is incorrect because it returns pointer to cstring stored in the Field which might be destroyed. However if the lifetime of the `Field` must be longer than the variable returned from the getter then we can return `const&` here to avoid copying.

module-services/service-db/agents/settings/FactorySettings.cpp
```diff
@@ -23,32 +23,33 @@ namespace settings
...
+        if (factoryData->getRowCount() <= 0) {
+            LOG_FATAL("No EEPROM factory data available!");
+            return;
         }
+
+        do {
+            const auto key   = (*factoryData)[0].getString();
+            const auto value = (*factoryData)[1].getCString();
+            const auto path  = settings::factory::entry_key + std::string("/") + key;
+            const settings::EntryPath variablePath{"", "", "", path, settings::SettingsScope::Global};
```
awulkiew, line 37: Would it make sense to create string once?
```suggestion
            const auto variablePath = settings::EntryPath{"", "", "", path, settings::SettingsScope::Global}.to_string();
```

awulkiew (approved):

module-services/service-db/agents/settings/FactorySettings.cpp
Lefucjusz: Done

module-db/Database/Field.hpp
Lefucjusz: Changed the returned value to `const&` as you suggested.

PR #4376, opened 2023-02-28
[BH-1627] Enable brownout detection

dawidwojtas-mudita:
There is the possibility of hanging
the MCU due to a power glitch
if the voltage drops to 2.8V
and the system wants to use some
peripherals like the eink or the backlight.

To prevent this situation brownout
detection has been enabled
which monitors the 1.1V and 2.5V
LDO regulators. If the voltage
will drop to the trigger voltages
then the interrupt is invoked.
The interrupt checks the source
of the interrupt and then reset
the MCU using WDOG_B pin
which is connected to the main power source.

After restarting, the MCU checks
the voltage. If it is too low the system enters SNVS mode.

Make sure that this PR:


SP2FET:
FYI - now commit messages can have line length of 100 :)

awulkiew (approved):

Maciej-Mudita (approved):

PR #4378, opened 2023-02-28
[BH-1624] Fix shutdown procedure

dawidwojtas-mudita:
In some cases, the system wasn't able to turn off
because the GUI service got stuck. The device
was still working in the background. The cause
was an empty queue in DrawCommandQueue which
hang the GUI worker.

The interface was modified and synchronization
the mechanism was removed.
The thread no longer waits in dequeue().

Also changed the worker to close in the right
the way the logger worker.

Make sure that this PR:

### DIFF
--- a/module-sys/Service/Worker.cpp
+++ b/module-sys/Service/Worker.cpp
@@ -206,8 +206,6 @@ namespace sys
     {
         assert(getState() == State::Initiated);
 
-        runnerTask = xTaskGetCurrentTaskHandle();
-
         BaseType_t task_error = 0;
         task_error            = xTaskCreate(
             Worker::taskAdapter, name.c_str(), stackDepth / sizeof(StackType_t), this, priority, &taskHandle);
@@ -222,12 +220,8 @@ namespace sys
 
     bool Worker::stop()
     {
-        if (runnerTask) {
-            assert(xTaskGetCurrentTaskHandle() == runnerTask);
-            assert(getState() == State::Running);
-            return sendControlMessage(ControlMessage::Stop);
-        }
-        return true;
+        assert(getState() == State::Running);
+        return sendControlMessage(ControlMessage::Stop);
     }
 
     bool Worker::sendControlMessage(ControlMessage message)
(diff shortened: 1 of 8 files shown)


awulkiew (approved):

module-sys/Service/Worker.cpp
```diff
@@ -222,12 +220,8 @@ namespace sys
 
     bool Worker::stop()
     {
-        if (runnerTask) {
-            assert(xTaskGetCurrentTaskHandle() == runnerTask);
-            assert(getState() == State::Running);
-            return sendControlMessage(ControlMessage::Stop);
-        }
-        return true;
+        assert(getState() == State::Running);
```
SP2FET, line 223: asserts don't work on release builds - if you meant to keep the check, I'd suggest switching to `if`

dawidwojtas-mudita: It was earlier for testing purposes so I left them except `assert(xTaskGetCurrentTaskHandle() == runnerTask);` that wasn't necessary :)

module-utils/log/Logger.cpp
```diff
@@ -166,19 +166,24 @@ namespace Log
...
-    auto Logger::dumpToFile(std::filesystem::path logPath) -> int
+    auto Logger::dumpToFile(std::filesystem::path logPath, bool isLoggerRunning) -> int
     {
         std::error_code errorCode;
         auto firstDump = !std::filesystem::exists(logPath, errorCode);
 
         if (errorCode) {
-            LOG_ERROR("Failed to check if file %s exists, error: %d", logPath.c_str(), errorCode.value());
+            if (isLoggerRunning) {
+                LOG_ERROR("Failed to check if file %s exists, error: %d", logPath.c_str(), errorCode.value());
```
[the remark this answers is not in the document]
dawidwojtas-mudita: You're right but now the logger has 2 buffers. Before `Logger::dumpToFile` we switch to the second buffer which is not dumped now so I believe here we shouldn't come across any issues :)

SP2FET (approved):

PR #4382, opened 2023-02-28
[BH-1630] Turn off the device for low voltage

dawidwojtas-mudita:
The system closes only if the SoC is 0%, but
it doesn't react if the voltage has a low level.
The new implementation invokes the close
procedure if the voltage is less than 3.4V.
This solution should avoid the possibility to
hang the MCU.

Make sure that this PR:

### DIFF
--- a/module-services/service-evtmgr/battery/BatteryController.cpp
+++ b/module-services/service-evtmgr/battery/BatteryController.cpp
@@ -75,11 +75,26 @@ namespace
     }
 } // namespace
 
-BatteryController::BatteryController(sys::Service *service,
-                                     xQueueHandle notificationChannel,
-                                     BatteryState::Thresholds thresholds)
+BatteryController::BatteryController(sys::Service *service, xQueueHandle notificationChannel, EventManagerParams params)
     : service{service}, charger{hal::battery::AbstractBatteryCharger::Factory::create(notificationChannel)},
-      brownoutDetector(service, *charger),
+      brownoutDetector(
+          service,
+          *charger,
+          params.voltage,
+          [this](BatteryBrownoutDetector::DetectionResult result) {
+              using Brownout = BatteryBrownoutDetector::DetectionResult;
+              switch (result) {
+              case Brownout::eventDetected: {
+                  auto messageBrownout = std::make_shared<sevm::BatteryBrownoutMessage>();
+                  this->service->bus.sendUnicast(std::move(messageBrownout), service::name::system_manager);
+                  LOG_INFO("Battery brownout %s. Sending message...", magic_enum::enum_name(result).data());
+              } break;
+              case Brownout::detectionNegative:
+              default:
+                  LOG_INFO("Battery brownout detection window finish with negative result");
+                  break;
+              }
+          }),
       batteryState{service,
                    [this](const auto state) {
                        Store::Battery::modify().levelState = transformBatteryState(state);
@@ -87,9 +102,9 @@ BatteryController::BatteryController(sys::Service *service,
                        this->service->bus.sendMulticast(std::move(stateChangeMessage),
                                                         sys::BusChannel::ServiceEvtmgrNotifications);
                    },
-                   thresholds}
+                   params.battery}
 {
-    updateSoC();
+    updateSoc();
     Store::Battery::modify().state = transformChargingState(charger->getChargingStatus());
     batteryState.check(transformChargingState(Store::Battery::modify().state), Store::Battery::modify().level);
 
@@ -111,7 +126,7 @@ void sevm::battery::BatteryController::handleNotification(Events evt)
         update();
         break;
     case Events::Brownout:
-        brownoutDetector.startDetection();
+        brownoutDetector.poll();
         break;
     }
 }
@@ -119,6 +134,7 @@ void sevm::battery::BatteryController::handleNotification(Events evt)
 void sevm::battery::BatteryController::poll()
 {
     update();
+    brownoutDetector.poll();
 }
 void sevm::battery::BatteryController::printCurrentState()
 {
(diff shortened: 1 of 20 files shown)


awulkiew (approved):

module-services/service-evtmgr/battery/BatteryController.cpp
```diff
@@ -122,18 +122,18 @@ void sevm::battery::BatteryController::poll()
 }
 void sevm::battery::BatteryController::printCurrentState()
 {
-    LOG_INFO("Charger state:%s Battery SOC %d voltage: %" PRIu32 "mV state: %s",
+    LOG_INFO("Charger state:%s Battery SOC %d voltage: %dmV state: %s",
```
SP2FET, line 125: why did you removed the macro?

dawidwojtas-mudita: Before this change, the voltage wasn't stored anywhere. I added this to `EventStore` and to be consistent now the voltage is `unsigned int` so macro is not necessary at all.

SP2FET: I'd say - better have some little portability than none, this won't hurt the way it's displaying, but when (someday) anyone would like to compile it on Linux or other platform, the compile error can happen :P

dawidwojtas-mudita: In this case, the Linux target wasn't able to compile the firmware so I had to change that. As long as the voltage is `unsigned int` it should be fine :)

module-services/service-evtmgr/battery/BatteryController.cpp
```diff
@@ -89,13 +89,13 @@ BatteryController::BatteryController(sys::Service *service,
...
 {
-    updateSoC();
+    updateSocAndVoltage();
     Store::Battery::modify().state = transformChargingState(charger->getChargingStatus());
     batteryState.check(transformChargingState(Store::Battery::modify().state), Store::Battery::modify().level);
 
     LOG_INFO("Initial charger state:%s", magic_enum::enum_name(Store::Battery::get().state).data());
     LOG_INFO("Initial battery SOC:%d", Store::Battery::get().level);
-    LOG_INFO("Initial battery voltage:%" PRIu32 "mV", charger->getBatteryVoltage());
+    LOG_INFO("Initial battery voltage:%dmV", Store::Battery::get().voltage);
```
swichu91, line 98: Same as below comment

module-services/service-evtmgr/battery/BatteryController.cpp
```diff
@@ -122,18 +122,18 @@ void sevm::battery::BatteryController::poll()
 }
 void sevm::battery::BatteryController::printCurrentState()
 {
-    LOG_INFO("Charger state:%s Battery SOC %d voltage: %" PRIu32 "mV state: %s",
+    LOG_INFO("Charger state:%s Battery SOC %d voltage: %dmV state: %s",
              magic_enum::enum_name(Store::Battery::get().state).data(),
              Store::Battery::get().level,
-             charger->getBatteryVoltage(),
+             Store::Battery::get().voltage,
```
swichu91, line 128: I would leave it as it was before.

products/BellHybrid/services/evtmgr/battery/BatteryController.cpp
```diff
@@ -0,0 +1,39 @@
...
+        }
+    }
+
+    void BatteryController::sendLowVoltageShutdownMessage()
+    {
+        auto message = std::make_shared<sevm::BatteryLowVoltageShutdownMessage>();
+        getService()->bus.sendUnicast(std::move(message), ::service::name::system_manager);
+    }
+
+} // namespace sevm::battery::bell
```
swichu91, line 39: How about extending `BatteryState` class instead? I think it would be a good idea as it already contains battery state logic. Apart from that, PurePhone could benefit from this features too ;)

module-services/service-evtmgr/battery/BatteryController.cpp
dawidwojtas-mudita: As you suggested @SP2FET and @swichu91 now the `EventStore` uses proper units from `Units.hpp` so I left `PRIu32` as it was before. I pushed an additional commit.

module-services/service-evtmgr/battery/BatteryController.cpp
dawidwojtas-mudita: Why? 
I moved the `charger->getBatteryVoltage()` to the `update()` function to save the voltage to `EventStore`. 
This approach was necessary because `CW2015` doesn't have a low voltage detection feature (brownout) 
so I had to save the voltage and then check if the value is critically low.
I wouldn't use `charger->getBatteryVoltage()` two times (for logging and checking low voltage) because it involves I2C to get the data.
If you see any reasons for changing this let me know :)

products/BellHybrid/services/evtmgr/battery/BatteryController.cpp
dawidwojtas-mudita: Thanks for the advice. :) This approach was the fastest because we felt pressure due to the testing schedule.
I agree that should be moved to `BatteryState` so I'll try to rewrite this feature :)

dawidwojtas-mudita: I checked the code and it looks like the `BatteryState` refer to predictable states. Brownout/low voltage is an unpredictable state so maybe it should be somewhere else? The Pure has its own class for brownout `BatteryBrownoutDetector`. This class is fed by `Events::Brownout` when the battery charger sends an interrupt.
What about extending this class and making an interface for Pure and Harmony for detecting brownout/low voltage?

module-services/service-evtmgr/battery/BatteryController.cpp
swichu91: `Store::Battery::get()` is simple a singleton in disguise. I'd suggest limit its usage to bare minimum. It would be nice to remove it altogether, but I know it's not possible for now.

products/BellHybrid/services/evtmgr/battery/BatteryController.cpp
dawidwojtas-mudita: Done. I removed `BatteryController` for the bell and modified `BatteryBrownoutDetector`.

swichu91 (approved):

PR #4384, opened 2023-03-01
[MOS-921] Fixed unexitable screen while having a missed call during t…

SP2FET:
…emplate selection

Created a cloned SMS template window to differenciate it from the templates used in SMS thred - better than creating a bool flag...

Added switchBack on missed call to application messages

Make sure that this PR:

### DIFF
--- a/module-apps/application-messages/include/application-messages/Constants.hpp
+++ b/module-apps/application-messages/include/application-messages/Constants.hpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2021, Mudita Sp. z.o.o. All rights reserved.
+// Copyright (c) 2017-2023, Mudita Sp. z.o.o. All rights reserved.
 // For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 #pragma once
@@ -18,6 +18,7 @@ namespace gui::name::window
     inline constexpr auto thread_options    = "ThreadOptions";
     inline constexpr auto sms_options       = "SMSOptions";
     inline constexpr auto sms_templates     = "SMSTemplates";
+    inline constexpr auto call_sms_templates = "CallSMSTemplates";
     inline constexpr auto thread_view       = "ThreadViewWindow";
 
 } // namespace gui::name::window
--- a/module-apps/application-messages/ApplicationMessages.cpp
+++ b/module-apps/application-messages/ApplicationMessages.cpp
@@ -13,13 +13,8 @@
 #include "ThreadWindowOptions.hpp"
 
 #include <Dialog.hpp>
-#include <DialogMetadata.hpp>
 #include <DialogMetadataMessage.hpp>
-#include <i18n/i18n.hpp>
-#include <memory>
 #include <messages/OptionsWindow.hpp>
-#include <MessageType.hpp>
-#include <module-db/queries/messages/sms/QuerySMSAdd.hpp>
 #include <module-db/queries/messages/sms/QuerySMSRemove.hpp>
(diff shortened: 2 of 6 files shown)


kkleczkowski (approved):

Maciej-Mudita (approved):

module-apps/application-messages/windows/CallSMSTemplatesWindow.cpp
```diff
@@ -0,0 +1,19 @@
...
+
+    CallSMSTemplatesWindow::CallSMSTemplatesWindow(app::ApplicationCommon *app, const std::string &windowName)
+        : SMSTemplatesWindow(app, windowName)
+    {
+        buildInterface();
+    }
+
+    CallSMSTemplatesWindow::~CallSMSTemplatesWindow()
+    {
+        destroyInterface();
```
awulkiew, line 17: This function is already called in the destructor of the base class SMSTemplatesWindow isn't it?

module-apps/application-messages/windows/CallSMSTemplatesWindow.cpp
```diff
@@ -0,0 +1,19 @@
...
+
+#include "CallSMSTemplatesWindow.hpp"
+
+namespace gui
+{
+
+    CallSMSTemplatesWindow::CallSMSTemplatesWindow(app::ApplicationCommon *app, const std::string &windowName)
+        : SMSTemplatesWindow(app, windowName)
+    {
+        buildInterface();
```
awulkiew, line 12: AFAIU this function is called in the constructor of the base class so it's called twice. Or am I missing something?

module-apps/application-messages/windows/CallSMSTemplatesWindow.cpp
SP2FET: I just copied it from the parent window - either way I'll just use two times the same window with different names - this will do the same :)

module-apps/application-messages/windows/CallSMSTemplatesWindow.cpp
SP2FET: I just copied it from the parent window - either way I'll just use two times the same window with different names - this will do the same :)

module-apps/application-messages/windows/CallSMSTemplatesWindow.cpp
SP2FET: resolving as it's no more related - the file has been deleted

module-apps/application-messages/windows/CallSMSTemplatesWindow.cpp
SP2FET: resolving as it's no more related - the file has been deleted

module-apps/application-messages/include/application-messages/Constants.hpp
```diff
@@ -18,6 +18,7 @@ namespace gui::name::window
     inline constexpr auto thread_options    = "ThreadOptions";
     inline constexpr auto sms_options       = "SMSOptions";
     inline constexpr auto sms_templates     = "SMSTemplates";
+    inline constexpr auto call_sms_templates = "CallSMSTemplates";
```
marziel, line 21: It was a good thing to eventually rename it so that it's clear that this thing is about calls :)

marziel (approved):

module-apps/application-messages/include/application-messages/Constants.hpp
SP2FET: tried my best :P

PR #4415, opened 2023-03-16
[MOS-357] Fix backing from new contact window

marziel:
Make sure that this PR:

### DIFF
--- a/module-apps/application-phonebook/windows/PhonebookNewContact.cpp
+++ b/module-apps/application-phonebook/windows/PhonebookNewContact.cpp
@@ -8,8 +8,8 @@
 #include "application-phonebook/ApplicationPhonebook.hpp"
 
 #include <Dialog.hpp>
+#include <service-appmgr/Controller.hpp>
 #include <service-db/DBServiceAPI.hpp>
-#include <messages/DialogMetadataMessage.hpp>
 
 namespace gui
 {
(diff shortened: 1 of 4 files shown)


module-apps/application-phonebook/windows/PhonebookNewContact.cpp
```diff
@@ -132,11 +132,16 @@ namespace gui
...
             verifyAndSave();
-
-            return true;
+        }
+        else if (!inputEvent.isShortRelease(KeyCode::KEY_RF) || !shouldCurrentAppBeIgnoredOnSwitchBack()) {
+            return AppWindow::onInput(inputEvent);
         }
 
-        return ret;
+        return nameOfPreviousApplication.has_value()
```
rrandomsky, line 140: Use `shouldCurrentAppBeIgnoredOnSwitchBack()` instead.
In _OptionContact.cpp : 51_ `ignoreCurrentWindowOnStack` flag is set to true, so is good to check it to.

This flag is in message is pass information "ignore a window" and `nameOfPreviousApplication` is just for help to switc back correctly - it's separate infos. Flag can be changed during "window flow" but name should stay the same.

TL;DR
In this case it's no matters, work fine fore sure, but for other window it can be important to check this flag too. I meet scenario when _Phonebook_ switch to _Messages_ and _Messages_ switch to _Templates_ and all of this windows have flag and nameOfPreviousApp but "window flow" looses this info....  and that's why `saveInfoAboutPreviousAppForProperSwitchBack();` in not on the beginning of `NewMessageWindow::onBeforeShow()` for example.
Using `shouldCurrentAppBeIgnoredOnSwitchBack()` is safe for future copy&paste of this mechanism to other Windows.

marziel: Done

Lefucjusz (approved):

rrandomsky (approved):

PR #4452, opened 2023-04-12
[CP-1942] Re-adding the same file results in an increase in the amoun…

lukasmas:
…t of occupied space

Fixed problem with adding files with the same name resulting in increasing its size instead of replacing the file. USB Stack changes.

Make sure that this PR:

### DIFF
--- a/module-services/service-desktop/endpoints/filesystem/FileContext.cpp
+++ b/module-services/service-desktop/endpoints/filesystem/FileContext.cpp
@@ -35,7 +35,7 @@ FileWriteContext::FileWriteContext(const std::filesystem::path &path,
                                    std::string crc32Digest,
                                    std::size_t offset)
     : FileContext(path, size, chunkSize, offset), crc32Digest(std::move(crc32Digest)),
-      file(path, std::ios::binary | std::ios::app)
+      file(path, std::ios::binary | std::ios::out)
 {}
 
 FileWriteContext::~FileWriteContext()


module-services/service-desktop/endpoints/filesystem/FileContext.cpp
```diff
@@ -35,7 +35,7 @@ FileWriteContext::FileWriteContext(const std::filesystem::path &path,
                                    std::string crc32Digest,
                                    std::size_t offset)
     : FileContext(path, size, chunkSize, offset), crc32Digest(std::move(crc32Digest)),
-      file(path, std::ios::binary | std::ios::app)
+      file(path, std::ios::binary | std::ios::out)
```
Lefucjusz, line 38: I'm not exactly sure how it works currently, as there have been some changes in this part of code recently, but previously the file was opened and closed upon every received chunk - wouldn't this change result in creation of the file containing only the last received chunk?

lukasmas: We changed it and now the file is created and opened as the transfer is started and closed as the transfer is done. 
With the option app, we have a bug which doesn't allow the replacement of the file but increases its size -> you have file X and you want to replace it and suddenly you have two X files combined. And that's why I am changing it to out.

Lefucjusz (approved):

dawidwojtas-mudita (approved):

PR #4464, opened 2023-04-27
[CP-1950] Templates menu navigation loop and pagination on pure gets …

lukasmas:
…blocked

Fixed problem where the Templates window gets unresponsive for user input while changes in templates were done via MC

Make sure that this PR:

### DIFF
--- a/module-apps/application-messages/windows/SMSTemplatesWindow.cpp
+++ b/module-apps/application-messages/windows/SMSTemplatesWindow.cpp
@@ -6,6 +6,7 @@
 #include "SMSTemplateItem.hpp"
 #include "SMSTemplatesWindow.hpp"
 
+#include <service-db/DBNotificationMessage.hpp>
 #include <i18n/i18n.hpp>
 #include <log/log.hpp>
 #include <service-appmgr/Controller.hpp>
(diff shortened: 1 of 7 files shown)


module-apps/application-messages/windows/SMSTemplatesWindow.cpp
```diff
@@ -128,4 +131,18 @@ namespace gui
...
+    {
+        auto msgNotification = dynamic_cast<db::NotificationMessage *>(msgl);
+        if (msgNotification != nullptr) {
+            if (msgNotification->interface == db::Interface::Name::SMSTemplate) {
+                if (msgNotification->dataModified()) {
+                    rebuild();
+                    return true;
+                }
+            }
+        }
```
dawidwojtas-mudita, line 145: Let's add `const` before `auto msgNotification`.
Did you consider checking each condition in one `if()` statement?

dawidwojtas-mudita (approved):

module-apps/application-messages/windows/SMSTemplatesWindow.cpp
Lefucjusz: Or if you're not a fan of such chained `if (condition1 && condition2 && ...)` consider rewriting it with early return pattern:

```
if (msgNotification == nullptr) {
    return false;
}

if (msgNotification->interface != db::Interface::Name::SMSTemplate) {
    return false;
}

etc...
```                

Such heavily nested code is much harder to read IMHO.

lukasmas: I am not a fan of such chained conditions but in this case, I think it looks just better than the early returns pattern :)

Lefucjusz (approved):

PR #4466, opened 2023-04-28
[CP-1937] Amount of space assigned to "Other" changes after uploading…

lukasmas:
… audio file

Fixed by sending information about device memory in floats instead of uints

Make sure that this PR:

### DIFF
--- a/module-services/service-desktop/endpoints/deviceInfo/DeviceInfoEndpointCommon.cpp
+++ b/module-services/service-desktop/endpoints/deviceInfo/DeviceInfoEndpointCommon.cpp
@@ -114,7 +114,7 @@ namespace sdesktop::endpoints
         return entries;
     }
 
-    auto DeviceInfoEndpointCommon::getStorageStats(const std::string &path) -> std::tuple<long, long>
+    auto DeviceInfoEndpointCommon::getStorageStats(const std::string &path) -> std::tuple<float, float>
     {
         constexpr auto bytesInMebibyte = 1024LLU * 1024LLU;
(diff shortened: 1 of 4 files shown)


module-services/service-desktop/endpoints/deviceInfo/DeviceInfoEndpointCommon.cpp
```diff
@@ -124,26 +124,24 @@ namespace sdesktop::endpoints
             return {-1, -1};
         }
 
-        const auto totalMbytes = static_cast<long>(
-            (static_cast<std::uint64_t>(vfstat.f_blocks) * static_cast<std::uint64_t>(vfstat.f_bsize)) /
-            bytesInMebibyte);
-        const auto freeMbytes = static_cast<long>(
-            (static_cast<std::uint64_t>(vfstat.f_bfree) * static_cast<std::uint64_t>(vfstat.f_bsize)) /
-            bytesInMebibyte);
+        const auto totalMbytes = static_cast<float>(
```
Lefucjusz, line 127: Maybe these can just be simplified to `const auto totalMbytes = static_cast<float>(vfstat.f_blocks * vfstat.f_bsize) / bytesInMebibyte` ? `f_blocks` is 64-bit integer, so `f_bsize` will be automatically promoted to 64-bit integer as well for the multiplication - no overflow should occur. Then the result will be casted to float and divided by 64-bit int, what will give float as a result ;)

Lefucjusz (approved):

dawidwojtas-mudita (approved):

PR #4467, opened 2023-04-28
Mos 975 missing new message notification

kkleczkowski:

Make sure that this PR:

### DIFF
--- a/module-apps/application-messages/ApplicationMessages.cpp
+++ b/module-apps/application-messages/ApplicationMessages.cpp
@@ -25,7 +25,7 @@
 #include <service-cellular/CellularMessage.hpp>
 #include <service-db/DBNotificationMessage.hpp>
 #include <service-appmgr/Controller.hpp>
-
+#include <apps-common/locks/data/PhoneLockMessages.hpp>
 namespace app
 {
     static constexpr auto messagesStackDepth = 1024 * 6; // 6Kb stack size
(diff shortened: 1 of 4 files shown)


Maciej-Mudita (approved):

module-apps/application-messages/include/application-messages/ApplicationMessages.hpp
```diff
@@ -69,6 +69,8 @@ namespace app
 
       private:
         ActionResult handleCreateSmsAction(std::unique_ptr<gui::SwitchData> data);
+
+        bool phoneIsLocked = false;
```
lukasmas, line 73: Maybe it will look better if the states will be introduced for example enum LOCKED, UNLOCKED instead of keeping the state in the bool variable

marziel: I think so, too

kkleczkowski: done.

module-apps/application-messages/ApplicationMessages.cpp
```diff
@@ -70,6 +72,16 @@ namespace app
             }
             return sys::MessageNone{};
         });
+
+        connect(typeid(locks::UnlockedPhone), [&](sys::Message *msg) -> sys::MessagePointer {
```
marziel, line 76: This could be done as one type, e.g. `PhoneLockStateMessage`, carrying a `PhoneLockState` enum value.

kkleczkowski: Message is not introduced in this PR and also is used in another services.

PR #4470, opened 2023-05-04
[CP-1966] Deleted file which is played does not stop the playback timer

lukasmas:
Fixed the timer issue when file was deleted during playback.

Make sure that this PR:

### DIFF
--- a/products/BellHybrid/apps/application-bell-relaxation/presenter/RelaxationRunningLoopPresenter.cpp
+++ b/products/BellHybrid/apps/application-bell-relaxation/presenter/RelaxationRunningLoopPresenter.cpp
@@ -57,7 +57,9 @@ namespace app::relaxation
                 return;
             }
         };
-        player.start(song.fileInfo.path, mode, std::move(onStartCallback));
+        auto onErrorCallback = [this]() { getView()->handleDeletedFile(); };
+
+        player.start(song.fileInfo.path, mode, std::move(onStartCallback), std::move(onErrorCallback));
     }
 
     void RelaxationRunningLoopPresenter::stop()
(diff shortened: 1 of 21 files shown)


products/BellHybrid/apps/application-bell-relaxation/presenter/RelaxationRunningLoopPresenter.cpp
```diff
@@ -57,7 +57,9 @@ namespace app::relaxation
                 return;
             }
         };
-        player.start(song.fileInfo.path, mode, std::move(onStartCallback));
+        auto onErrorCallback = [this]() { getView()->handleDeletedFile(); };
+
+        player.start(song.fileInfo.path, mode, std::move(onStartCallback), std::move(onErrorCallback));
```
SP2FET, line 62: As far as I can see, deleted file is not the only one error that can happen, so you should handle other error codes too

lukasmas: When I was testing it, the retCode value in RelaxationPlayer always was "failed" or "success". 
The other values are never used in this case. 
Furthermore, we don't know anything about any other errors that could happen in Relaxation, except unsupported files and files removed while playing. 
I guess that any extraordinary error if it happens will be treated like unsupported files, because of retCode values that are only set with "failed" or "success" in this case.

SP2FET: so, if the unsupported file can be an issue there, there should be a handler for this too :) or at least clean up those error codes, to be clear

dawidwojtas-mudita (approved):
LGTM ;)

Maciej-Mudita (approved):

PR #4472, opened 2023-05-08
[CP-1936] Sync stack forever in progress

lukasmas:
Handilng situtatin when /user/temp directory gets currupted.

Make sure that this PR:


module-services/service-desktop/Sync.cpp
```diff
@@ -57,6 +57,16 @@ Sync::CompletionCode Sync::PrepareSyncPackage(sys::Service *ownerService, std::f
...
+    const auto tempPath = purefs::dir::getTemporaryPath();
+    if (std::filesystem::exists(tempPath) and !std::filesystem::is_directory(tempPath)) {
+        LOG_WARN("%s should be a directory, removing...", tempPath.c_str());
+        if (!std::filesystem::remove(tempPath)) {
+            LOG_ERROR("Failed to remove %s!", tempPath.c_str());
+            return false;
+        }
+        return true;
+    }
+
```
SP2FET, line 69: do you know why the `temp` catalog becomes a file? I think we should find the source of that issue instead of fixing it this way...

swichu91 (changes requested):
@SP2FET's comment is right on point. First, find out why it happened and then implement a fix(if required at all).

module-services/service-desktop/Sync.cpp
lukasmas: We don't know. 
The filesystem was OK. fsck didn't report any errors. The logs also don't contain anything connected to this situation.
So I guess we don't fix it because we don't have any reproduction steps just a phone with such a condition :D

PR #4474, opened 2023-05-09
[MOS-578] Fix misleading SMS notification text

marziel:
Display "New message", translated accordingly - in singular - in case there had been multiple unread
SMS messages and now only one remains unread.

Make sure that this PR:

### DIFF
--- a/module-apps/apps-common/notifications/NotificationsListPresenter.cpp
+++ b/module-apps/apps-common/notifications/NotificationsListPresenter.cpp
@@ -77,7 +77,10 @@ auto NotificationsListPresenter::create(const notifications::NotSeenSMSNotificat
     -> NotificationListItem *
 {
     auto item = new NotificationWithEventCounter(notifications::NotificationType::NotSeenSms, notification->getValue());
-    setNotificationText(item, notification, "app_desktop_unread_messages");
+    setNotificationText(item,
+                        notification,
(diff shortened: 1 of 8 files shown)


module-apps/apps-common/notifications/NotificationsListPresenter.cpp
```diff
@@ -77,7 +77,10 @@ auto NotificationsListPresenter::create(const notifications::NotSeenSMSNotificat
     -> NotificationListItem *
 {
     auto item = new NotificationWithEventCounter(notifications::NotificationType::NotSeenSms, notification->getValue());
-    setNotificationText(item, notification, "app_desktop_unread_messages");
+    setNotificationText(item,
+                        notification,
+                        (notification->getValue() == 1) ? "app_desktop_unread_single_message"
```
SP2FET, line 82: what does this "1" mean?

SP2FET (approved):

Lefucjusz (approved):

module-apps/apps-common/notifications/NotificationsListPresenter.cpp
marziel: This means that there's only 1 notification :D I agree it's confusing, but it's more a matter of the too generic name `getValue()`

PR #4475, opened 2023-05-09
[MOS-686] Fixed MTP availability only after phone unlocked

rrandomsky:
Fixed file access via MTP even when phone is not unlocked. Now access is granted when the phone is unlocked by the user entering a passcode. If the phone is not passcode protected (passcode is nor set) then access to the files is always possible via MTP.

Make sure that this PR:

### DIFF
--- a/module-apps/apps-common/locks/handlers/PhoneLockHandler.cpp
+++ b/module-apps/apps-common/locks/handlers/PhoneLockHandler.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
+// Copyright (c) 2017-2023, Mudita Sp. z.o.o. All rights reserved.
 // For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 #include "PhoneLockHandler.hpp"
@@ -115,9 +115,12 @@ namespace locks
                 std::make_unique<gui::PopupRequestParams>(gui::popup::ID::PhoneLockInfo));
         }
 
-        app::manager::Controller::sendAction(owner,
-                                             app::manager::actions::AbortPopup,
-                                             std::make_unique<gui::PopupRequestParams>(gui::popup::ID::PhoneLockInput));
+        if (lock.inputValue.empty()) {
+            app::manager::Controller::sendAction(
+                owner,
+                app::manager::actions::AbortPopup,
+                std::make_unique<gui::PopupRequestParams>(gui::popup::ID::PhoneLockInput));
+        }
     }
 
     void PhoneLockHandler::phoneUnlockAction()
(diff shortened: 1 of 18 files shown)


module-apps/apps-common/locks/handlers/PhoneLockHandler.cpp
```diff
@@ -127,12 +130,12 @@ namespace locks
         owner->bus.sendMulticast(std::make_shared<locks::UnlockedPhone>(), sys::BusChannel::PhoneLockChanges);
     }
 
-    void PhoneLockHandler::phoneInputRequiredAction()
+    void PhoneLockHandler::phoneInputRequiredAction(bool isSoftRequest = false)
```
lukasmas, line 133: Don't use bool as a function parameter. 
I recommend splitting the function or using an enum instead. 
I will look better and will be easier to understand

rrandomsky: Ok, you right.
I changed bool to enum `PhoneLockHandler::TypeOfRequest`

Maybe code is still not clear enough and I should split functions?

lukasmas: I would consider renaming it "PhoneLockHandler::TypeOfRequest" 
As I noticed it has 2 values Soft and Normal and for example, TypeOfRequest::Soft is not saying much about what it is and what it does
Try something that will explain this -> why unlock for MTP is soft? (UnlockType:: {UnlockPhone, UnlockMTPOnly} 
`
connect(typeid(locks::UnlockPhone), [&](sys::Message *request) -> sys::MessagePointer {
            return phoneLockHandler.handleUnlockRequest(locks::PhoneLockHandler::TypeOfRequest::Normal);
        });
        connect(typeid(locks::UnlockPhoneForMTP), [&](sys::Message *request) -> sys::MessagePointer {
            return phoneLockHandler.handleUnlockRequest(locks::PhoneLockHandler::TypeOfRequest::Soft);
        });
`
if it doesn't help much try to consider the introduction of a second function dedicated only to unlocking MTP

module-apps/apps-common/locks/handlers/PhoneLockHandler.hpp
```diff
@@ -20,6 +20,13 @@ namespace locks
 
     class PhoneLockHandler
     {
+      public:
+        enum class TypeOfRequest
+        {
+            Normal = 0,
+            Soft
```
SP2FET, line 27: why do we have two types of phone lock?

module-apps/apps-common/locks/handlers/PhoneLockHandler.cpp
rrandomsky: Ok then
I changed _enum_ names to:
```cpp
        enum class ReasonForRequest
        {
            Unspecified = 0, // For default behaviors
            UnlockForMTP
        };
 ```
 and added a proper comment where passing of the _new enum_  as arg. actually changes the behavior:
 ```cpp
        // If it's for MTP, don't show the input passcode popup if it's already on the screen
        data->ignoreIfTheSamePopupIsOnTopOfTheStack = (reqReason == ReasonForRequest::UnlockForMTP);
```

module-apps/apps-common/locks/handlers/PhoneLockHandler.hpp
rrandomsky: Applies to the scenario where the user plugs in the USB cable when there is already a pop-up on the screen to enter the password.
This scenario was mentioned by QA ["After connecting Pure to the computer, while entering the passcode,[...]](https://appnroll.atlassian.net/browse/MOS-686?focusedCommentId=67376)

module-apps/apps-common/locks/handlers/PhoneLockHandler.cpp
lukasmas: > Unspecified = 0, // For default behaviors

So maybe default? :D

module-apps/apps-common/locks/handlers/PhoneLockHandler.hpp
```diff
@@ -20,6 +20,13 @@ namespace locks
 
     class PhoneLockHandler
     {
+      public:
+        enum class ReasonForRequest
+        {
+            Default = 0, // For default behaviors
+            UnlockForMTP
```
SP2FET, line 27: I'm wondering if the name is proper - AFAIU this case is when you connect the phone to the PC and MTP wants to unlock the phone - so maybe it should be named `MTPUnlock` or something?

SP2FET (approved):

marziel (approved):

PR #4483, opened 2023-05-19
[BH-1673] Harmony random resets fixes

Lefucjusz:
* Switching VDD_SOC_IN only after PLL2 is turned off.
* Weak LDO stabilization delay.
* Removed switching to 1.275V before frequency change.
* Cleanups in LDO switching.
* Removed switching PeriphClk2Div for fCPU <= 24MHz, using AhbDiv instead.
* Removed log from _exit that caused logger mutex deadlock.

Coauthored with @dawidwojtas-mudita and @Maciej-Mudita.

Make sure that this PR:

### DIFF
--- a/module-bsp/board/rt1051/bsp/lpm/RT1051LPMCommon.cpp
+++ b/module-bsp/board/rt1051/bsp/lpm/RT1051LPMCommon.cpp
@@ -65,7 +65,7 @@ namespace bsp
     CpuFrequencyMHz RT1051LPMCommon::onChangeUp(CpuFrequencyMHz freq, bsp::CpuFrequencyMHz newFrequency)
     {
         if ((freq <= CpuFrequencyMHz::Level_1) && (newFrequency > CpuFrequencyMHz::Level_1)) {
-            // connect internal the load resistor
+            // connect internal load resistor
             ConnectInternalLoadResistor();
             // Switch DCDC to full throttle during oscillator switch
             SetHighestCoreVoltage();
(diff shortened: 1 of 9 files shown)


module-bsp/board/rt1051/bsp/lpm/RT1051LPMCommon.cpp
```diff
@@ -178,27 +178,30 @@ namespace bsp
...
     {
-        // Enable weak 2P5 and turn off regular 2P5
-        PMU->REG_2P5 |= PMU_REG_2P5_ENABLE_WEAK_LINREG_MASK;
-        PMU->REG_2P5 &= ~PMU_REG_2P5_ENABLE_LINREG_MASK;
-        // Enable weak 1P1 and turn off regular 1P1
-        PMU->REG_1P1 |= PMU_REG_1P1_ENABLE_WEAK_LINREG_MASK;
-        PMU->REG_1P1 &= ~PMU_REG_1P1_ENABLE_LINREG_MASK;
+        // Enable weak 2P5 and 1P1 LDO
+        PMU->REG_2P5_SET = PMU_REG_2P5_ENABLE_WEAK_LINREG_MASK;
+        PMU->REG_1P1_SET = PMU_REG_1P1_ENABLE_WEAK_LINREG_MASK;
```
SP2FET, line 198: as you're using the same mask for CLR and SET registers, maybe it's worth removing `ENABLE` from the mask?

dawidwojtas-mudita: Those masks are used by the fsl SDK so we won't to change it (maybe in the future we'll update the fsl SDK) ;)

SP2FET: alright :)

dawidwojtas-mudita (approved):

PR #4487, opened 2023-05-24
[MOS-973] Fix for a ghost call after quick click back key to end a call

rrandomsky:
Fixed a scenario where the user click to make a call to some contact and then quickly click BACK (right function button) to end the call and despite that Pure was still calling without showing any information. This fix prevent to handle key press event as a key longpress even after application lose a focus, or change a state.

Make sure that this PR:

### DIFF
--- a/module-apps/application-call/ApplicationCall.cpp
+++ b/module-apps/application-call/ApplicationCall.cpp
@@ -205,6 +205,12 @@ namespace app
         return ret;
     }
 
+    sys::MessagePointer ApplicationCall::handleAppClose(sys::Message *msgl)
+    {
+        callModel->hangUpCall();
+        return ApplicationCommon::handleAppClose(msgl);
+    }
+
     void ApplicationCall::createUserInterface()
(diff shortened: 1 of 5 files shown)


module-apps/application-call/ApplicationCall.cpp
```diff
@@ -205,6 +205,12 @@ namespace app
         return ret;
     }
 
+    sys::MessagePointer ApplicationCall::handleAppClose(sys::Message *msgl)
+    {
+        callModel->hangUpCall();
```
SP2FET, line 210: does this check in call machine if the call is happening before sending AT+CHUP? (just asking out of curiosity :P )

rrandomsky: No it doesn't do any check. It cause send AT+CHUP in any state of call machine besides Idle.

SP2FET (approved):

Lefucjusz (approved):

PR #4503, opened 2023-06-07
[CP-1968] Disconnecting the device during file upload causes problems…

lukasmas:
… with the USB

Fixed the problem with an endless loop while sending data when USB is already disconnected.

Make sure that this PR:


module-services/service-desktop/WorkerDesktop.cpp
```diff
@@ -179,7 +178,13 @@ bool WorkerDesktop::handleSendQueueMessage(std::shared_ptr<sys::WorkerQueue> &qu
         return false;
     }
 
-    bsp::usbCDCSend(sendMsg);
+    const std::uint32_t dataSent = bsp::usbCDCSend(sendMsg);
+    if (dataSent != sendMsg->length()) {
+        LOG_ERROR("Data not sent! Data sent: %lu, Msg length: %lu",
+                  static_cast<long unsigned int>(dataSent),
```
dawidwojtas-mudita, line 184: You can use `PRIu32` instead of `lu`. In this way, you don't need to use `static_cast`. ;)

dawidwojtas-mudita (approved):

module-services/service-desktop/WorkerDesktop.cpp
Lefucjusz: And by replacing second `%lu` with `%zu` you can remove the second cast too:

`LOG_ERROR("Data not sent! Data sent: %" PRIu32 "B, msg length: %zuB", dataSent, sendMsg->length());`

Lefucjusz (approved):

module-services/service-desktop/WorkerDesktop.cpp
lukasmas: I am not sure about %zu, it had some problems with Linux build, that's why I've replaced it with static_cast

Lefucjusz: I've just tested it locally and it builds on both RT1051 and Linux target with no errors, maybe it's somehow related with different compiler version on CI

lukasmas: never mind, it works, thanks :)

PR #4507, opened 2023-06-13
 [BH-1705] Adjusted power off timer to >=10s

MateuszSz-mudita:
When holding the back button, the device would prompt to power off after ~9-10s. This was not inline with the manual. Now the prompt will come up just after 10s. Previously the timer was set to 9s to account for delays caused be the OS and display, however it is now decided that the only thing that should matter is the time you actually have to hold the button for the prompt to show. This time is more or less exactly the same as the time set in the code.

Make sure that this PR:


Maciej-Mudita:
@MateuszSz-mudita Have you considered a changelog entry?

Maciej-Mudita (approved):

MateuszSz-mudita:
> 

This is only a small fix up of https://github.com/mudita/MuditaOS/commit/1984c7f479314610fd98ad069717f4f4383424a2
The changelog was changed with that commit and is still accurate.

PR #4516, opened 2023-06-21
[MOS-998] Provide additional info in crashdump filename

Lefucjusz:
Added info about product, OS version and
commit hash to crashdump filename to
simplify analysis of the crashdumps
without corresponding logs available.

Make sure that this PR:


dawidwojtas-mudita (approved):

module-utils/CrashdumpMetadataStore/CrashdumpMetadataStore.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+        void setCommitHash(const std::string &hash);
+        const std::string &getCommitHash();
+
+        std::string getMetadataString();
+
+      private:
+        CrashdumpMetadata();
+
+        static constexpr auto separator                 = "_";
```
exeosdev, line 35: Do we use that outisde the .cpp file? If not, we technically don't have to pollute the header and move these to the implementation file.

exeosdev (approved):

module-utils/CrashdumpMetadataStore/CrashdumpMetadataStore.hpp
Lefucjusz: True, I've moved it to `cpp` file ;)

PR #4536, opened 2023-07-10
[MOS-1004] Fix autolock on SMS template call rejection

Lefucjusz:
Fix of the issue that unlocked phone would
automatically lock after user tried to reject
incoming call via SMS and stayed on the
templates window.

Make sure that this PR:

### DIFF
--- a/module-apps/application-messages/data/SMSdata.hpp
+++ b/module-apps/application-messages/data/SMSdata.hpp
@@ -72,18 +72,25 @@ class SMSSendRequest : public SMSRequest
 class SMSSendTemplateRequest : public SMSRequest
 {
   public:
-    explicit SMSSendTemplateRequest(const utils::PhoneNumber::View &phoneNumber, bool preventAutoLock = false)
-        : SMSRequest(phoneNumber), preventAutoLock(preventAutoLock)
+    enum class AutolockBehavior
+    {
+        Allow,
+        Prevent
+    };
+
+    explicit SMSSendTemplateRequest(const utils::PhoneNumber::View &phoneNumber,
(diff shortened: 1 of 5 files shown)


module-apps/application-messages/data/SMSdata.hpp
```diff
@@ -72,18 +72,25 @@ class SMSSendRequest : public SMSRequest
 class SMSSendTemplateRequest : public SMSRequest
 {
   public:
-    explicit SMSSendTemplateRequest(const utils::PhoneNumber::View &phoneNumber, bool preventAutoLock = false)
-        : SMSRequest(phoneNumber), preventAutoLock(preventAutoLock)
+    enum class AutolockBehavior
```
lukasmas, line 75: I think it would be nice to consider in the future to expand this mechanism everywhere not only in this particular message because in the end preventAutolock is still a bool flag

lukasmas (approved):

rrandomsky (approved):

Maciej-Mudita (approved):

module-apps/application-messages/data/SMSdata.hpp
Lefucjusz: You're right, but it'd require a refactor in several places in code that IMO is out of scope of this task. The change I made here is to avoid passing bool flag as a function parameter, as it greatly reduces readability of the function call. It's hard to determine what `true` means in `foo(someObj, someData, true)` at the first glance, unless the function name specifies it explicitly.
Anyway, that's a good remark indeed and we can consider it for some future refactoring ;)

PR #4546, opened 2023-07-20
[MOS-972] Fixed missing contact entries when scrolling phonebook

Lefucjusz:
Fix of the issue that in some cases contact entry
would not be displayed on any page in case it
was on the boundary of two pages.

Make sure that this PR:

### DIFF
--- a/module-apps/application-phonebook/widgets/PhonebookListView.hpp
+++ b/module-apps/application-phonebook/widgets/PhonebookListView.hpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
+// Copyright (c) 2017-2023, Mudita Sp. z.o.o. All rights reserved.
 // For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 #pragma once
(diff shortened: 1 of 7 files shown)


module-apps/application-phonebook/widgets/PhonebookListView.hpp
```diff
@@ -14,16 +14,20 @@ namespace gui
...
+                          std::uint32_t x,
+                          std::uint32_t y,
+                          std::uint32_t w,
+                          std::uint32_t h,
+                          std::shared_ptr<ListItemProvider> prov);
+
+      private:
         void addItemsOnPage() override;
         void addLabelMarker(PhonebookItem *item);
+        unsigned getSlotsLeft();
```
exeosdev, line 28: Maybe `uint32_t` or `size_t` to be more explicit?

module-gui/gui/widgets/ListViewEngine.hpp
```diff
@@ -109,7 +109,7 @@ namespace gui
         [[nodiscard]] unsigned getFocusItemIndex();
 
         /// Total provider elements count
-        unsigned int elementsCount = listview::nPos;
+        unsigned elementsCount = listview::nPos;
```
exeosdev, line 112: `auto` would be even better here 🙂

dawidwojtas-mudita (approved):

module-apps/application-phonebook/widgets/PhonebookListView.hpp
Lefucjusz: Changed to `std::size_t` ;)

module-gui/gui/widgets/ListViewEngine.hpp
Lefucjusz: Unfortunately to declare class member with `auto` it needs to be also `static` - I removed this `int` only to unify unsigned integers declarations across the whole file, as all other declarations here omit it.

exeosdev: Ah, just noticed the context! Too many things at the same time, sorry :)

exeosdev (approved):

PR #4555, opened 2023-07-31
[MOS-1007] Fix for add/update contact with 2 same phone numbers

rrandomsky:
There was possibility to add contact, or edit some contact, with have 2 same phone numbers (exactly the same or witch/without country code). Now this has been fixed and is not possible to create, even by editing, contact with 2 exactly the same numbers or 2 same numbers and one of them with a country code

Make sure that this PR:

### DIFF
--- a/module-db/Interface/ContactRecord.cpp
+++ b/module-db/Interface/ContactRecord.cpp
@@ -35,6 +35,15 @@ ContactRecordInterface::ContactRecordInterface(ContactsDB *db)
 
 auto ContactRecordInterface::Add(ContactRecord &rec) -> bool
 {
+    if (rec.numbers.size() > 2) {
+        LOG_WARN("Contact has more than 2 numbers");
+    }
+
+    if (hasContactRecordSameNumbers(rec)) {
+        LOG_ERROR("New record can not have 2 same numbers");
+        return false;
+    }
+
     bool result = contactDB->contacts.add(ContactsTableRow{Record(DB_ID_NONE), .speedDial = rec.speeddial});
     if (!result) {
         return false;
@@ -125,6 +134,10 @@ auto ContactRecordInterface::RemoveByID(uint32_t id) -> bool
 
 auto ContactRecordInterface::Update(const ContactRecord &rec) -> bool
 {
+    if (rec.numbers.size() > 2) {
+        LOG_WARN("Contact has more than 2 numbers");
+    }
+
     ContactsTableRow contact = contactDB->contacts.getByIdWithTemporary(rec.ID);
     if (!contact.isValid()) {
         return false;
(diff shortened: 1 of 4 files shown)


Maciej-Mudita (approved):

module-db/Interface/ContactRecord.cpp
```diff
@@ -1595,3 +1606,24 @@ auto ContactRecordInterface::changeNumberRecordInPlaceIfCountryCodeIsOnlyDiffere
     }
     return true;
 }
+auto ContactRecordInterface::hasContactRecordSameNumbers(const ContactRecord &rec) -> bool
+{
+    if (rec.numbers.size() >= 2) {
+        if (rec.numbers.size() > 2) {
+            LOG_WARN("Contact record has more tha 2 numbers. Checking similarity for first 2 numbers only");
```
dawidwojtas-mudita, line 1613: Typo here :)

module-db/tests/ContactsRecord_tests.cpp
```diff
@@ -767,7 +767,7 @@ TEST_CASE("Check replacement of number in place in db when only different is hav
 
     // Preparation of DB initial state
     auto records = ContactRecordInterface(&contactsDb.get());
-    ContactRecord testContactRecord;
+    //    ContactRecord testContactRecord;
```
dawidwojtas-mudita, line 770: If it is not necessary just get rid of it.

lukasmas, line 770: remove if not needed

dawidwojtas-mudita (approved):

module-db/Interface/ContactRecord.cpp
lukasmas, line 1613: Spelling error:
"The contact record has more than 2 numbers. Checking similarity for the first 2 numbers only"

BTW. Is it possible to have more than two numbers? Maybe we should handle it somehow and return an error because the user cannot add more than two numbers while creating a new contact.

module-db/Interface/ContactRecord.cpp
```diff
@@ -1595,3 +1606,24 @@ auto ContactRecordInterface::changeNumberRecordInPlaceIfCountryCodeIsOnlyDiffere
...
+        }
+        utils::PhoneNumber newPhoneNumber0(rec.numbers.at(0).number);
```
lukasmas, line 1615: consider rename varaibles:
firstNumber | secondNumber
or
primaryNumber | secondaryNumber

module-db/Interface/ContactRecord.cpp
lukasmas: I would consider doing a precheck in Add and Update functions 
```
if(rec.numbers.size() > 2)
{
LOG_ERROR("Contact can not have more than two numbers");
return false;
}
```

lukasmas (approved):

module-db/Interface/ContactRecord.cpp
rrandomsky: Practically the user are not able to provide more than 2 numbers for Contact. Technically is there a possibility to add more, but only the GUI stands in the way.

I'll add this check, as you wrote, but without return and as warning kind of log, because in code are many "leftovers" which wanted to provide more advanced functions to Contacts, like type of Number (ContactNumberType::HOME/ WORK etc.)
And because are code is Open I wouldn't stay in the way for other implementation of MuditaOS  than ours (for Pure).

PR #4570, opened 2023-08-08
[MOS-1007] Fix for no popup when contact cannot be saved because of sa…

rrandomsky:
…me nubers

Fix for fix. It add popup when user try to save contact witch practically the same numners  (even if having a country code is only difference)

Make sure that this PR:
- [x] Has changelog entry added (**change log from previous MR is enough**)

### DIFF
--- a/image/system_a/data/lang/English.json
+++ b/image/system_a/data/lang/English.json
@@ -372,6 +372,8 @@
     "app_phonebook_new_contact_email": "Email",
     "app_phonebook_new_contact_first_name": "First name",
     "app_phonebook_new_contact_invalid_number": "<text>Cannot save this contact.<br></br>The phone number you entered <br></br>is in an invalid format.</text>",
+    "app_phonebook_new_contact_unable_to_save": "Unable to save contact",
+    "app_phonebook_new_contact_same_numbers": "<text>Despite the country code entered,<br></br>the phone numbers are the same.</text> <br></br><br></br> <text weight='bold' size='27'>Enter another numbers to save <br></br>the contact.</text>",
     "app_phonebook_new_contact_last_name": "Last name",
     "app_phonebook_new_contact_note": "Note",
     "app_phonebook_new_contact_number": "Phone number",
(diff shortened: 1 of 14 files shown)


dawidwojtas-mudita (approved):

module-services/service-db/include/service-db/DBServiceAPI.hpp
```diff
@@ -114,6 +114,8 @@ class DBServiceAPI
 
     static auto IsContactInFavourites(sys::Service *serv, const utils::PhoneNumber::View &numberView) -> bool;
     static auto IsContactInEmergency(sys::Service *serv, const utils::PhoneNumber::View &numberView) -> bool;
+
+    [[deprecated]] static auto hasContactSameNumbers(sys::Service *serv, const ContactRecord &rec) -> bool;
```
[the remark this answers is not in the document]
rrandomsky: Just Copy&Paste mistake. I'll remove this.
Despite this, other methods here like `ContactAdd()` are deprecated, and Idk why. E.g. `CellularServiceAPI` works  in the same way - "send message" way.

image/system_a/data/lang/English.json
```diff
@@ -372,6 +372,8 @@
     "app_phonebook_new_contact_email": "Email",
     "app_phonebook_new_contact_first_name": "First name",
     "app_phonebook_new_contact_invalid_number": "<text>Cannot save this contact.<br></br>The phone number you entered <br></br>is in an invalid format.</text>",
+    "app_phonebook_new_contact_unable_to_save": "Unable to save contact",
+    "app_phonebook_new_contact_same_numbers": "<text>Cannot save this contact.<br></br>The phone numbers you entered <br></br>are practically the same.</text>",
```
Lefucjusz, line 376: IMO this copy doesn't sound right. Personally I'd be very confused seeing such message when trying to add a contact - what does it mean that phone numbers are "practically the same" and what is the value of presenting such message to user? The differences between phone numbers can be defined in a exact, measurable way - the condition that causes this message to appear is that in the phonebook the user already has a number that only differs by (not) having country code. That is the info I'd prefer to see as a user so I can understand why I'm not able to add the contact and what I shall do to make it possible.

lukasmas (approved):

image/system_a/data/lang/English.json
rrandomsky: After discussions the text is changed already.

PR #4575, opened 2023-08-14
[MOS-1018] ServiceAudio power management optimizations

Lefucjusz:
Optimized power management strategy used by
ServiceAudio:
* changed minimum CPU frequency from 528MHz to 264MHz;
* added switching to 24MHz when operation active, but paused;
* minor cleanup in several places.

Make sure that this PR:

### DIFF
--- a/module-bsp/board/rt1051/bsp/lpm/Bandgap.cpp
+++ b/module-bsp/board/rt1051/bsp/lpm/Bandgap.cpp
@@ -1,0 +1,35 @@
+// Copyright (c) 2017-2023, Mudita Sp. z.o.o. All rights reserved.
+// For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
+
+#include "Bandgap.hpp"
+#include <cstdint>
+#include <fsl_common.h>
+
+namespace
+{
+    constexpr std::uint32_t REFTOP_LOWPOWER_MASK = (1U << 2);
+}
+
+namespace bsp::bandgap
+{
+    void SwitchToRegularMode()
+    {
+        /* Enable regular bandgap and wait for it to stabilize */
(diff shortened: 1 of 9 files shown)


Maciej-Mudita (approved):

module-bsp/board/rt1051/bsp/lpm/Bandgap.cpp
```diff
@@ -0,0 +1,35 @@
...
+    constexpr std::uint32_t REFTOP_LOWPOWER_MASK = (1U << 2);
+}
+
+namespace bsp::bandgap
+{
+    void SwitchToRegularMode()
+    {
+        /* Enable regular bandgap and wait for it to stabilize */
+        CCM_ANALOG->MISC0_CLR = CCM_ANALOG_MISC0_REFTOP_PWD_MASK;
+        while ((CCM_ANALOG->MISC0 & CCM_ANALOG_MISC0_REFTOP_VBGUP_MASK) == 0) {}
```
lukasmas, line 19: Are you sure that the execution will never be stuck here?

lukasmas (approved):

module-bsp/board/rt1051/bsp/lpm/Bandgap.cpp
Lefucjusz: This code is just copy-paste from NXP's AN12085 Application Note. I guess it shouldn't loop here forever, but even if it does - this means that some terrible failure has occurred and MCU is unable to turn on its internal bandgap reference. At this point OS is doomed anyway, as it will inevitably crash while executing one of the next instructions - probably switching to PLL2, which will cut off clock from CPU, as PLL2 doesn't work without bandgap.

PR #4589, opened 2023-08-24
[MOS-1024] Fix unlocking phone after locked with tethering popup

Lefucjusz:
Fix of the issue that after phone has locked
on tethering confirmation popup, unplugging
USB cable would result in bypassing phone
lock screen without providing password.
Minor ApplicationCommon cleanup.

Make sure that this PR:

### DIFF
--- a/module-apps/apps-common/ApplicationCommon.cpp
+++ b/module-apps/apps-common/ApplicationCommon.cpp
@@ -78,7 +78,7 @@ namespace app
                                          std::string parent,
                                          StatusIndicators statusIndicators,
                                          StartInBackground startInBackground,
-                                         uint32_t stackDepth,
+                                         std::uint32_t stackDepth,
                                          sys::ServicePriority priority)
         : Service(std::move(name), std::move(parent), stackDepth, priority),
           popupFilter(std::make_unique<gui::popup::Filter>()), windowsStackImpl(std::make_unique<WindowsStack>()),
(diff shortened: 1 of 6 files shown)


rrandomsky (approved):

module-apps/apps-common/ApplicationCommon.cpp
```diff
@@ -914,35 +920,46 @@ namespace app
...
-        if (not windowsFactory.isRegistered(windowName)) {
+        if (!windowsFactory.isRegistered(windowName)) {
             LOG_ERROR("Cannot find window %s windowsFactory in application: %s", windowName.c_str(), GetName().c_str());
             return;
         }
 
         pushWindow(windowName);
         const auto window = getWindow(windowName);
-        if (not window) {
+        if (window == nullptr) {
```
dawidwojtas-mudita, line 957: Usually, you use `!` so maybe also here ;)

dawidwojtas-mudita (approved):

module-apps/apps-common/ApplicationCommon.cpp
Lefucjusz: I try to explicitly compare pointers against `nullptr` value, so that it is easily seen that we're dealing with pointer. Using `!` creates an ambiguity - it can be used to check either null pointer, empty `std::optional` or `bool` with `false` value. But that's just my personal preference, so if you think `!` fits better here I'll change it ;)

PR #4591, opened 2023-08-24
[MOS-1026] Add extended logging to Pure's charger driver

Lefucjusz:
* Added extended logging to Pure's charger driver.
* Removed redundant handling of INOKB pin interrupt - charger is configured to provide
the same interrupt via INTB pin.
* Minor code cleanup.

Make sure that this PR:

### DIFF
--- a/module-bsp/board/rt1051/puretx/bsp/battery_charger/MAX77818.hpp
+++ b/module-bsp/board/rt1051/puretx/bsp/battery_charger/MAX77818.hpp
@@ -1,13 +1,12 @@
-// Copyright (c) 2017-2021, Mudita Sp. z.o.o. All rights reserved.
+// Copyright (c) 2017-2023, Mudita Sp. z.o.o. All rights reserved.
 // For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 #pragma once
 
 namespace bsp::battery_charger
 {
-
-    constexpr inline auto BATTERY_CHARGER_I2C_ADDR = 0xD2 >> 1;
-    constexpr inline auto FUEL_GAUGE_I2C_ADDR      = 0x6C >> 1;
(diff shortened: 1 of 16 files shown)


lukasmas (approved):

module-bsp/board/rt1051/puretx/bsp/battery_charger/MAX77818.hpp
```diff
@@ -181,6 +187,20 @@ namespace bsp::battery_charger
         lim1600mA = 0x20, /// 1C of battery
     };
 
+    // CHG_CNFG_06
+    enum class ChargerSettingsProtection
+    {
+        Enable  = 0,
+        Disable = (0x03 << 2)
```
dawidwojtas-mudita, line 194: This shift looks a bit weird here. I would instead use `0x0C` ;)

dawidwojtas-mudita (approved):

module-bsp/board/rt1051/puretx/bsp/battery_charger/MAX77818.hpp
Lefucjusz: I wrote it this way as it is shown like this in RM of MAX77818: https://www.analog.com/media/en/technical-documentation/data-sheets/MAX77818.pdf, see p. 71, `CHG_CNFG_06 Register Bit Description (0xBD)` - it states that `0x03 = write capability is unlocked`, and the value is set on bits 2 and 3, so I just left-shifted the value copied from the docs :P If you prefer I'll change this ;)

dawidwojtas-mudita: You don't need to change this then ;) Firstly I thought that the explicit value is 0x0C.

PR #4596, opened 2023-08-29
[BH-1761] Add a log when manual reset is performed

dawidwojtas-mudita:
If the user performs a manual reset there is no information in the logs. It can be treated as software resets.
So the best solution is to add a log about incoming manual reset and dump all logs to the file.

Make sure that this PR:

### DIFF
--- a/products/BellHybrid/services/evtmgr/EventManager.cpp
+++ b/products/BellHybrid/services/evtmgr/EventManager.cpp
@@ -173,8 +173,8 @@ void EventManager::buildKeySequences()
 
     auto resetSeq      = std::make_unique<ResetSequence>(*this);
     resetSeq->onAction = [this]() {
-        app::manager::Controller::sendAction(
-            this, app::manager::actions::ShowPopup, std::make_unique<RebootPopupRequestParams>());
+        LOG_INFO("Incoming manual reset caused by user!");
+        EventManagerCommon::dumpLogsToFile();
     };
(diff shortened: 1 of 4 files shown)


products/BellHybrid/services/evtmgr/EventManager.cpp
```diff
@@ -173,8 +173,8 @@ void EventManager::buildKeySequences()
 
     auto resetSeq      = std::make_unique<ResetSequence>(*this);
     resetSeq->onAction = [this]() {
-        app::manager::Controller::sendAction(
```
Lefucjusz, line 176: I'm not sure what's the current status of this popup (whether it is shown at all), but isn't this needed?

Lefucjusz (approved):

products/BellHybrid/services/evtmgr/EventManager.cpp
dawidwojtas-mudita: This popup is not used yet because the timeout for this is ~10s. However, manual reset takes 6,7s. Probably the ecoboot will be responsible for displaying this popup, but there is no final decision so I decided to leave this popup in the codebase.

Lefucjusz: Alright, thanks for clarification!

rrandomsky (approved):

MateuszSz-mudita (approved):

PR #4606, opened 2023-09-05
[MOS-552] Fix alarm clock input behavior

Lefucjusz:
Fix of multiple issues that manifested in
input fields of alarm clock, i.e. value
not erasing automatically, redundant
double-zero filling of hours field,
lack of zero-filling of minutes field
when navigating to AM/PM field in
12h mode.

Make sure that this PR:

### DIFF
--- a/module-apps/apps-common/widgets/DateAndTimeStyle.hpp
+++ b/module-apps/apps-common/widgets/DateAndTimeStyle.hpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2021, Mudita Sp. z.o.o. All rights reserved.
+// Copyright (c) 2017-2023, Mudita Sp. z.o.o. All rights reserved.
 // For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 #pragma once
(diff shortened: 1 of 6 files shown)


rrandomsky (approved):

module-apps/apps-common/widgets/DateAndTimeStyle.hpp
```diff
@@ -11,7 +11,7 @@ namespace style::window::date_and_time
     inline constexpr auto separator        = 30;
     inline constexpr auto time_input_12h_w = 120;
     inline constexpr auto time_input_24h_w = 195;
-    inline constexpr auto hBox_h           = height - 1.25 * topMargin;
+    inline constexpr auto hBox_h           = static_cast<int>(height - 1.25 * topMargin);
```
dawidwojtas-mudita, line 14: What about add `f` to `1.25f`?

dawidwojtas-mudita (approved):

module-apps/apps-common/widgets/DateAndTimeStyle.hpp
Lefucjusz: It probably won't change runtime computational complexity coming from float vs. double operations as all of the values involved in this equation are `constexpr`'s, so the result shall be computed at compilation time, but I've changed it to 1.25f as you suggested ;)

PR #4619, opened 2023-09-18
[MOS-195] Phone number input style unification

rrandomsky:
Unification of input style for phone number. Now every time when the user will provide a phone number in gui the right function key (RF key) will be work as CLEAR. This means that if any number is in the phone number field, the RF key will remove the last character. Otherwise it'll work as typical BACK button. This changes also provide ability to add * and # sign to number.

Make sure that this PR:


module-apps/application-messages/windows/NewMessage.cpp
```diff
@@ -46,14 +46,40 @@ namespace gui
...
 
-      private:
+      protected:
         std::unique_ptr<gui::TextBackup> state;
     };
 
+    class NewMessageWindow::RecipientMemento : public NewMessageWindow::MessageMemento
+    {
+      public:
+        void setState(const gui::Text *_state, const std::shared_ptr<ContactRecord> _contact)
```
lukasmas, line 56: I don't think we use code such naming convention which starts with _. Consider rename the variables to state and contact

lukasmas: Additionally, I think that with such a convention members are named with _  and arguments without it

module-apps/application-messages/windows/NewMessage.cpp
```diff
@@ -46,14 +46,40 @@ namespace gui
...
+        {
+            if (_contact != nullptr && !_contact->numbers.empty()) {
+                phoneNumberView = std::make_unique<utils::PhoneNumber::View>(_contact->numbers.at(0).number);
+            }
+
+            MessageMemento::setState(_state);
+        }
+
+        void restoreState(gui::Text *_state, utils::PhoneNumber::View *_phoneNumberView)
```
lukasmas, line 65: same as above
what's the difference between _phoneNumberView and phoneNumberView
I would consider naming it as newPhoneNumber and currentPhoneNumber. If this is what this is about

module-apps/application-messages/windows/NewMessage.cpp
```diff
@@ -312,7 +344,8 @@ namespace gui
         };
         body->addWidget(message);
         body->addWidget(new Span(Axis::Y, body->getHeight()));
-        message->setText(""); // to set actual size of Text
+        recipient->setText(""); // to set actual size of Text
```
lukasmas, line 347: Is this comment sth like TO DO marker or the text size is 0?

module-apps/application-phonebook/widgets/InputLinesWithLabelWidget.hpp
```diff
@@ -18,13 +18,14 @@ namespace gui
...
-            const std::function<void(Text *text)> &inputOptions                                = nullptr,
-            unsigned int lines                                                                 = 1);
+            const std::function<void(const UTF8 &text, bool emptyOthers)> &navBarTemporaryMode            = nullptr,
+            const std::function<void()> &navBarRestoreFromTemporaryMode                                   = nullptr,
+            const std::function<void(const UTF8 &text, bool isRFKeyFunctionAsClear)> &navBarSetRFKeyLabel = nullptr,
+            const std::function<void(const UTF8 &text, bool active)> &navBarSetOptionsLabel               = nullptr,
+            const std::function<void()> &selectSpecialCharacter                                           = nullptr,
+            const std::function<void(std::function<void()> restoreFunction)> &restoreInputMode            = nullptr,
+            const std::function<void(Text *text)> &inputOptions                                           = nullptr,
+            unsigned int lines                                                                            = 1);
```
lukasmas, line 28: std::uint32_t

lukasmas (approved):

module-apps/application-phonebook/widgets/InputLinesWithLabelWidget.cpp
```diff
@@ -303,5 +309,14 @@ namespace gui
             return contact->note != inputText->getText();
         };
     }
+    void InputLinesWithLabelWidget::setRightFunctionKayLabel()
+    {
+        if (inputText != nullptr && inputText->isInputMode(InputMode::phone) && this->navBarSetRFKeyLabel &&
+            !inputText->isEmpty()) {
+            this->navBarSetRFKeyLabel(utils::translate(utils::translate("app_call_clear")), true);
+            return;
```
dawidwojtas-mudita, line 317: It'd be better to use `else` statement for `navBarSetRFKeyLabel` instead of using `return`. I suppose that you don't need to use `this` after all ;)

module-gui/gui/widgets/text/Text.cpp
```diff
@@ -360,6 +360,19 @@ namespace gui
         this->inputMode = mode;
     }
 
+    bool Text::isInputMode(InputMode::Mode mode)
+    {
+        if (inputMode == nullptr) {
```
dawidwojtas-mudita, line 365: I believe you can simplify this function eg.
`        if (inputMode != nullptr && inputMode->is(mode)) {
            return true;
        }
        return false;`

dawidwojtas-mudita (approved):

module-apps/application-messages/windows/NewMessage.cpp
rrandomsky: This is a comment of the code I found here. I will change it to a more appropriate one

module-gui/gui/widgets/text/Text.cpp
rrandomsky: Yes, but now the structure of this method looks similar to the methods around this one
(and we can log situation when `inputMode` is nullptr)

dawidwojtas-mudita: If it is important to catch situations when the `inputMode` is `nullptr` that's ok. You can leave it as it is. ;)

PR #4621, opened 2023-09-19
[MOS-1035] Fixes in new alarm window UI flow

Lefucjusz:
* Fixed disappearing left functional button label
when popup appeared while focused on sound 
selection field or AM/PM field
* Fixed erasing selections on custom alarm 
edit window when popup appeared.
* Added preview playback stopping after manually 
blocking the phone.

Make sure that this PR:

### DIFF
--- a/module-apps/application-alarm-clock/models/NewEditAlarmModel.hpp
+++ b/module-apps/application-alarm-clock/models/NewEditAlarmModel.hpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2021, Mudita Sp. z.o.o. All rights reserved.
+// Copyright (c) 2017-2023, Mudita Sp. z.o.o. All rights reserved.
 // For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 #pragma once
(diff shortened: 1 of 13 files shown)


module-apps/application-alarm-clock/models/NewEditAlarmModel.hpp
```diff
@@ -48,8 +48,8 @@ namespace app::alarmClock
         void clearData() override;
         void createData();
 
-        [[nodiscard]] unsigned int getMinimalItemSpaceRequired() const override;
-        [[nodiscard]] unsigned int requestRecordsCount() override;
+        [[nodiscard]] unsigned getMinimalItemSpaceRequired() const override;
+        [[nodiscard]] unsigned requestRecordsCount() override;
         [[nodiscard]] gui::ListItem *getItem(gui::Order order) override;
         void requestRecords(uint32_t offset, uint32_t limit) override;
```
lukasmas, line 54: std::uint32_t

lukasmas (approved):

rrandomsky (approved):

module-apps/application-alarm-clock/models/NewEditAlarmModel.hpp
Lefucjusz: It cannot simply be changed to other type as these methods are derived from `ListItemProvider` interface class, so to avoid function signature mismatch I'd need to change it there too, what would imply changing it in every `ListItemProvider` implementation across the whole system.

PR #4634, opened 2023-10-11
[MOS-669] Add too hot battery notification when charging

Lefucjusz:
* Added notification showing on the main screen if the battery has reached too
high temperature to continue charging.
* Code cleanups.

Make sure that this PR:


module-services/service-evtmgr/EventManager.cpp
```diff
@@ -144,6 +145,7 @@ sys::ReturnCodes EventManagerCommon::InitHandler()
             if (!targetApplication.empty()) {
                 bus.sendUnicast(std::make_shared<sevm::BatteryStatusChangeMessage>(), targetApplication);
             }
+            bus.sendUnicast(std::make_shared<sevm::BatteryStatusChangeMessage>(), service::name::appmgr);
```
Maciej-Mudita, line 148: I wonder if it's possible to send it via multicast instead of informing each service separately?

Maciej-Mudita (approved):

module-services/service-evtmgr/EventManager.cpp
Lefucjusz: Good question ;) Because of some reason `EventManager` works as a dispatcher for this message, i.e. it receives unicast sent from `BatteryController` and then forwards it to other services. I haven't seen such strategy in other places of our OS, so I assumed it's done that way on purpose and didn't change it to prevent introducing potential bugs/regressions, which could appear in both products, as this part is common.

It would be worth analyzing it, though, as these unicasts here are the reason why logs such as "ServiceCellular does not exist" show up on Harmony.

Lefucjusz: Done, reported under MOS-1045 ticket!

module-apps/apps-common/notifications/NotificationProvider.cpp
```diff
@@ -120,6 +120,25 @@ void NotificationProvider::handleSnooze(unsigned snoozeCount)
     send();
 }
 
+void NotificationProvider::handleBatteryTooHot(bool tooHot, bool allowUpdate)
```
rrandomsky, line 123: I suggest change name `tooHot` to `isBatteryTooHot`
`tooHot` could mean that some other temperature in device is too high and we assume that battery is hot too.
`is<sth>` is more clear name convention when using bool.

Lefucjusz: Fixed!

rrandomsky (approved):

PR #4648, opened 2023-10-24
[MOS-644] Displaying number entered on the keypad during a call

ghost:
Now, on the active call screen, the user can see the numbers provided during the call, which is useful when, for example, the user calls a call center.

Make sure that this PR:

### DIFF
--- a/module-apps/application-call/data/CallAppStyle.hpp
+++ b/module-apps/application-call/data/CallAppStyle.hpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
+// Copyright (c) 2017-2023, Mudita Sp. z.o.o. All rights reserved.
 // For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 #pragma once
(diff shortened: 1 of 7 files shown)


module-apps/application-call/data/CallAppStyle.hpp
```diff
@@ -84,6 +84,13 @@ namespace callAppStyle
             inline constexpr auto w = 240U;
             inline constexpr auto h = 20U;
         } // namespace durationLabel
+        namespace providedNumbersDuringTheCallLabel
```
Lefucjusz, line 87: I'd avoid creating such long names with, especially with particles in them. Maybe something more like `enteredDigitsField` or `enteredDigitsLabel`?

module-apps/application-call/windows/CallWindow.cpp
```diff
@@ -154,6 +168,23 @@ namespace gui
         }
     }
 
+    void CallWindow::updateEnteredNumber(const char &text, bool isVisible)
```
Lefucjusz, line 171: This entire logic should IMO be put in presenter instead of view, now it breaks MVP model rules - business logic is present in view part.

module-apps/application-call/presenter/CallPresenter.cpp
```diff
@@ -204,5 +211,10 @@ namespace app::call
     {
         model->clear();
     }
+    bool CallWindowContract::Presenter::isProvidedCharValid(const char digitChar)
```
dawidwojtas-mudita, line 214: `const` is not necessary if you pass a value ;)

dawidwojtas-mudita (approved):

lukasmas (approved):

PR #4656, opened 2023-11-02
[BH-1787]  Reboot to MSC endpoint

MateuszSz-mudita:
To allow the automation of gathering logs and swapping out OS images during tests, a MSC reboot endpoint was added. Together with a change to ecoboot, this allows the sending of a "reboot to MSC" command. The device will then reboot and enter MSC mode.

Make sure that this PR:


module-services/service-desktop/endpoints/msc/MscEndpoint.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+namespace sdesktop::endpoints
+{
+    auto MscEndpoint::handle(Context &context) -> void
+    {
+        switch (context.getMethod()) {
+        case http::Method::post: {
+            const auto &body = context.getBody();
+
+            auto msg = body["msg"].string_value();
```
lukasmas, line 18: how about creating a key called "msc" holding bool value instead of checking if the key "msg" is in JSON and then if the message in it contains proper string just to check
if(body[json::msc].bool_value()){// code here}
And add "msc" to a variable in the proper namespace as it is done in other endpoints

lukasmas: Another proposition to make it more general is rename this endpoint to RebootEndpint / shutdownEndpoint and create an enum with different reasons
and then check like this:
if(body[json::rebootReason].int_value() == ReabootReason::MSC){// code here}
else{ //TO DO 
}

lukasmas: It will give us the possibility to implement reboot devices without touching them if we need it in the future

module-services/service-desktop/endpoints/reboot/RebootEndpoint.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include <endpoints/message/Sender.hpp>
+
+namespace sdesktop::endpoints
+{
+    auto RebootEndpoint::handle(Context &context) -> void
+    {
+        switch (context.getMethod()) {
+        case http::Method::post: {
+
+            const auto &body = context.getBody();
```
lukasmas, line 17: You can create a file with helper functions and move it there.
You can check how it is done in other helpers
function name: handlePost(context)

MateuszSz-mudita:
> There is old version of master branch for `third-party/usb_stack` submodule.

Thanks for noticing, removed this from the commit :D

module-services/service-desktop/endpoints/reboot/RebootHelper.cpp
```diff
@@ -0,0 +1,37 @@
...
+#include <json11.hpp>
+
+namespace sdesktop::endpoints
+{
+    using sender::putToSendQueue;
+    auto RebootHelper::processPost(Context &context) -> ProcessResult
+    {
+        const auto rebootType = context.getBody()[json::rebootTypes::rebootType].string_value();
+
+        if (rebootType == json::rebootTypes::MSC) {
```
lukasmas, line 20: Maybe use switch() instead of if :D

module-services/service-desktop/endpoints/include/endpoints/JsonKeyNames.hpp
```diff
@@ -125,4 +114,16 @@ namespace sdesktop::endpoints::json
         inline constexpr auto timeLeftToNextAttempt = "timeLeftToNextAttempt";
     } // namespace usb
 
+    namespace reboot
+    {
+        inline constexpr auto rebootType = "rebootType";
+        enum
+        {
+            MSC = 1, // JSON11 evaluates empty message to 0
```
lukasmas, line 122: "msc" instead of "MSC"

module-services/service-desktop/endpoints/reboot/RebootHelper.cpp
```diff
@@ -0,0 +1,35 @@
...
+        case json::reboot::MSC:
+            sys::SystemManagerCommon::RebootMSC(owner);
+            return {sent::no, ResponseContext{.status = http::Code::Accepted}};
+        case json::reboot::reboot:
+            sys::SystemManagerCommon::Reboot(owner);
+            return {sent::no, ResponseContext{.status = http::Code::Accepted}};
+        case json::reboot::shutDown:
+            sys::SystemManagerCommon::RegularPowerDown(owner);
+            return {sent::no, ResponseContext{.status = http::Code::Accepted}};
+        }
```
lukasmas, line 30: I would consider adding "default" in the end to make the code look better.

lukasmas (approved):

module-services/service-desktop/endpoints/reboot/RebootHelper.cpp
MateuszSz-mudita: I considered it, but I found this to be the most consistent with our codebase.

Lefucjusz (approved):
Please make sure that this change doesn't enable Pure to reboot to MSC mode too.

PR #4661, opened 2023-11-08
[MOS-220] Changed signal strength refresh mechanism

Lefucjusz:
Modified signal strength update mechanism
to avoid unnecessary screen refreshes in
case new signal strength value is equal
to the previous one.
Minor cleanups.

Make sure that this PR:

### DIFF
--- a/module-services/service-cellular/src/CSQHandler.cpp
+++ b/module-services/service-cellular/src/CSQHandler.cpp
@@ -8,56 +8,64 @@
 #include <ticks.hpp>
 #include <chrono>
 
+namespace
+{
+    constexpr auto urcThreshold     = 4;
+    constexpr auto pollModeDuration = std::chrono::minutes{60};
+
+    bool isRssiValid(std::uint32_t csq)
+    {
+        constexpr auto invalidRssiLow  = 99;
+        constexpr auto invalidRssiHigh = 199;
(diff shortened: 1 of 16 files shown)


module-services/service-cellular/src/CSQHandler.cpp
```diff
@@ -8,70 +8,78 @@
...
+namespace
 {
+    constexpr auto urcThreshold     = 4;
+    constexpr auto pollModeDuration = std::chrono::minutes{60};
+
+    bool isRssiValid(std::uint32_t csq)
+    {
+        constexpr auto invalidRssiLow  = 99;
+        constexpr auto invalidRssiHigh = 199;
+        return ((csq != invalidRssiLow) && (csq != invalidRssiHigh));
```
lukasmas, line 20: csq can't has these 2 particular value, and all other are ok?
or should it be more like this:
return ((csq > invalidRssiLow) && (csq < invalidRssiHigh))
?

lukasmas (approved):

module-services/service-cellular/src/CSQHandler.cpp
Lefucjusz: TBH I have no idea - I just copied this code from another place and put in separate function to improve readability

dawidwojtas-mudita (approved):

PR #4662, opened 2023-11-09
[BH-1814] Fix EInk errors in logs

dawidwojtas-mudita:
According to the newest fsl library, the LPSPI_Reset function shouldn't be invoked in the LPSPI_MasterInit function.
As a result, the LPSPI module can't work after calling this API.

Fix the system crash when an EInk error occurs. If the display returns the error the driver could invoke the EinkPowerDown function recursively which causes a crash.

Make sure that this PR:

### DIFF
--- a/module-bsp/board/rt1051/bsp/eink/bsp_eink.cpp
+++ b/module-bsp/board/rt1051/bsp/eink/bsp_eink.cpp
@@ -29,6 +29,8 @@
 #define BSP_EINK_TRANSFER_WRITE_CLOCK 18000000
 #define BSP_EINK_TRANSFER_READ_CLOCK  3000000
 
+using namespace drivers;
+
 typedef enum
 {
     EventWaitRegistered,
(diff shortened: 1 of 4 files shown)


MateuszSz-mudita (approved):

lukasmas (approved):

Maciej-Mudita (approved):

module-bsp/board/rt1051/bsp/eink/bsp_eink.cpp
```diff
@@ -79,7 +79,7 @@ typedef struct _bsp_eink_driver
 namespace
 {
     using namespace drivers;
-    static lpspi_master_config_t s_eink_lpspi_master_config;
+    static lpspi_master_config_t spi_config;
     static std::shared_ptr<drivers::DriverGPIO> gpio;
```
Lefucjusz, line 83: `static` seems to be redundant here, as all of these objects are declared in anonymous namespace, also `using namespace drivers` is not needed if I understand correctly, as references to classes are done with `drivers::` anyway

module-bsp/board/rt1051/bsp/eink/bsp_eink.cpp
```diff
@@ -188,25 +186,24 @@ status_t BSP_EinkInit()
...
+    LPSPI_MasterGetDefaultConfig(&spi_config);
 
-    s_eink_lpspi_master_config.whichPcs           = kLPSPI_Pcs0;
-    s_eink_lpspi_master_config.pcsActiveHighOrLow = kLPSPI_PcsActiveLow;
+    spi_config.baudRate                      = BSP_EINK_TRANSFER_WRITE_CLOCK;
+    spi_config.bitsPerFrame                  = 8;
+    spi_config.cpol                          = kLPSPI_ClockPolarityActiveHigh;
+    spi_config.cpha                          = kLPSPI_ClockPhaseFirstEdge;
+    spi_config.direction                     = kLPSPI_MsbFirst;
+    spi_config.pcsToSckDelayInNanoSec        = 1000000000 / spi_config.baudRate;
```
Lefucjusz, line 196: Maybe some constexpr for that value? I guess it's number of nanoseconds in second :smile:

Lefucjusz (approved):

module-bsp/board/rt1051/bsp/eink/bsp_eink.cpp
dawidwojtas-mudita: Thanks, done ;)

module-bsp/board/rt1051/bsp/eink/bsp_eink.cpp
dawidwojtas-mudita: It is just copied from the fsl library, but I agree. Added constexpr value ;)

PR #4673, opened 2023-11-20
[MOS-1059] Erase only sensitive data from logs

ghost:

Make sure that this PR:

### DIFF
--- a/module-cellular/modem/ATCommon.cpp
+++ b/module-cellular/modem/ATCommon.cpp
@@ -11,6 +11,7 @@
 #include "ATStream.hpp"
 #include <at/ATFactory.hpp>
 #include <gsl/util>
+#include <Anonymize.hpp>
 
 using namespace at;
 using namespace std::chrono_literals;
@@ -41,16 +42,20 @@ void Channel::cmdLog(std::string cmd, const Result &result, std::chrono::millise
     } break;
     case Result::Code::ERROR: {
         LOG_ERROR("AT response: error");
-        LOG_SENSITIVE(
-            LOGERROR, "[AT]: >%s<, >%s<", cmd.c_str(), result.response.size() ? result.response.back().c_str() : "");
+        LOG_SENSITIVE(LOGERROR,
+                      "[AT]: >%s<, >%s<",
+                      utils::anonymize::anonymizeCellularIfNecessary(cmd).c_str(),
+                      result.response.size() ? result.response.back().c_str() : "");
     } break;
     default:
-        LOG_SENSITIVE(
-            LOGDEBUG, "[AT]: >%s<, >%s<", cmd.c_str(), result.response.size() ? result.response.back().c_str() : "");
+        LOG_SENSITIVE(LOGDEBUG,
+                      "[AT]: >%s<, >%s<",
+                      utils::anonymize::anonymizeCellularIfNecessary(cmd).c_str(),
+                      result.response.size() ? result.response.back().c_str() : "");
         break;
     }
     for ([[maybe_unused]] const auto &s : result.response) {
-        LOG_SENSITIVE(LOGINFO, "[AT] > %s", s.c_str());
+        LOG_SENSITIVE(LOGINFO, "[AT] > %s", utils::anonymize::anonymizeCellularIfNecessary(s).c_str());
     }
 }
 
@@ -86,7 +91,7 @@ Result Channel::cmd(const std::string &cmd, std::chrono::milliseconds timeout, s
 
     cmdInit();
     std::string cmdFixed = formatCommand(cmd);
-    LOG_DEBUG("Start of %s", cmdFixed.c_str());
+    LOG_DEBUG("Start of %s", utils::anonymize::anonymizeCellularIfNecessary(cmdFixed).c_str());
     cmdSend(cmdFixed);
 
     auto startTime = std::chrono::steady_clock::now();
--- a/module-utils/utility/Anonymize.hpp
+++ b/module-utils/utility/Anonymize.hpp
@@ -1,0 +1,25 @@
+// Copyright (c) 2017-2023, Mudita Sp. z.o.o. All rights reserved.
+// For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
+
+#pragma once
+
+#include <string>
+#include <algorithm>
+#include <functional>
+#include <map>
+
+namespace utils::anonymize
+{
+    // Determine how many characters to leave at the end of anonymized substring
+    constexpr auto SIGNS_TO_LEAVE_FOR_PIN_AND_PUK       = 0;
+    constexpr auto SIGNS_TO_LEAVE_FOR_PHONE_NUMBERS     = 2;
+    constexpr auto SIGNS_TO_LEAVE_FOR_NET_PROVIDER_NAME = 1;
+
+    // Basic function to anonymize provided string
+    std::string anonymizeInQuotationMarks(const std::string &textToAnonymize, std::size_t singsToLeaveAtEnd = 0);
+    std::string anonymizeNumbers(const std::string &textToAnonymize, std::size_t singsToLeaveAtEnd = 0);
+
+    // To anonymize logs from cellular
+    std::string anonymizeCellularIfNecessary(const std::string &text);
+
+} // namespace utils::anonymize
--- a/module-bluetooth/Bluetooth/interface/profiles/HFP/HFP.cpp
+++ b/module-bluetooth/Bluetooth/interface/profiles/HFP/HFP.cpp
@@ -16,6 +16,7 @@
 
 #include <BluetoothWorker.hpp>
 #include "SCO/ScoUtils.hpp"
+#include <Anonymize.hpp>
 
 #define ARRAY_LENGTH(x) (sizeof(x) / sizeof((x)[0]))
 
@@ -145,7 +146,7 @@ namespace bluetooth
     std::uint8_t HFP::HFPImpl::serviceBuffer[serviceBufferSize];
 
     std::unique_ptr<SCO> HFP::HFPImpl::sco;
-    SCOCodec HFP::HFPImpl::codec                       = SCOCodec::CVSD;
+    SCOCodec HFP::HFPImpl::codec = SCOCodec::CVSD;
     std::shared_ptr<CVSDAudioDevice> HFP::HFPImpl::audioDevice;
 
     std::unique_ptr<CellularInterface> HFP::HFPImpl::cellularInterface;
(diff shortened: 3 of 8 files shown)


module-cellular/modem/ATCommon.cpp
```diff
@@ -156,3 +162,108 @@ Result Channel::checkResult(bsp::cellular::CellularResultCode cellularResult)
 
     return result;
 }
+
+std::string Channel::anonymizeIfNecessary(std::string text)
```
Lefucjusz, line 166: IMO this shouldn't be here. Why `Channel` class should be responsible for data anonymization? This should be logger's responsibility - that way not only will it be much clearer from the code logic point of view, but also easier to implement any data anonymization outside `Channel` class (inside the class too, as now every time when you want to add some sensitive logs you need to remember to use these functions instead of just writing the log).

module-cellular/modem/ATCommon.cpp
```diff
@@ -156,3 +162,108 @@ Result Channel::checkResult(bsp::cellular::CellularResultCode cellularResult)
...
+{
+    std::string substringCPIN = "CPIN";
```
Lefucjusz, line 168: These can be `const`

module-cellular/modem/ATCommon.cpp
```diff
@@ -156,3 +162,108 @@ Result Channel::checkResult(bsp::cellular::CellularResultCode cellularResult)
...
+    }
+    else if (text.find(substringCSPN) != std::string::npos) {
+        text = anonymizeInQuotationMarks(text, 2);
+    }
+
+    return text;
+}
+std::string Channel::anonymizeInQuotationMarks(std::string textToAnonymize, size_t singsToLeft)
+{
+    size_t startPosition = 0;
```
Lefucjusz, line 194: `std::size_t`

module-cellular/modem/ATCommon.cpp
```diff
@@ -156,3 +162,108 @@ Result Channel::checkResult(bsp::cellular::CellularResultCode cellularResult)
...
+    else if (text.find(substringCPBR) != std::string::npos) {
+        text = anonymizeInQuotationMarks(text, 1);
+    }
+    else if (text.find(substringCSPN) != std::string::npos) {
+        text = anonymizeInQuotationMarks(text, 2);
+    }
+
+    return text;
+}
+std::string Channel::anonymizeInQuotationMarks(std::string textToAnonymize, size_t singsToLeft)
```
Lefucjusz, line 192: `std::size_t signsToLeave`

module-cellular/modem/ATCommon.cpp
```diff
@@ -156,3 +162,108 @@ Result Channel::checkResult(bsp::cellular::CellularResultCode cellularResult)
...
+    size_t endPosition   = 0;
+
+    // Find the position of the substring within the input string
+    while (startPosition < textToAnonymize.length()) {
+        startPosition = textToAnonymize.find('\"', startPosition);
+        endPosition   = textToAnonymize.find('\"', startPosition + 1);
+
+        if (startPosition != std::string::npos && endPosition != std::string::npos) {
+            // Extract the substring between the quotes
+            std::string extractedSubstring = textToAnonymize.substr(startPosition + 1, endPosition - startPosition - 1);
```
Lefucjusz, line 204: `const auto &`

module-cellular/modem/ATCommon.cpp
```diff
@@ -156,3 +162,108 @@ Result Channel::checkResult(bsp::cellular::CellularResultCode cellularResult)
...
+
+            // Determine the length of the extracted substring
+            size_t extractedLength = extractedSubstring.length();
```
Lefucjusz, line 207: `const auto`

module-cellular/modem/ATCommon.cpp
```diff
@@ -156,3 +162,108 @@ Result Channel::checkResult(bsp::cellular::CellularResultCode cellularResult)
...
+    return text;
+}
+std::string Channel::anonymizeInQuotationMarks(std::string textToAnonymize, size_t singsToLeft)
+{
+    size_t startPosition = 0;
+    size_t endPosition   = 0;
+
+    // Find the position of the substring within the input string
+    while (startPosition < textToAnonymize.length()) {
+        startPosition = textToAnonymize.find('\"', startPosition);
```
Lefucjusz, line 199: I'd move this token to some `constexpr`, as it might not be obvious what it represents

module-cellular/modem/ATCommon.cpp
```diff
@@ -156,3 +162,108 @@ Result Channel::checkResult(bsp::cellular::CellularResultCode cellularResult)
...
+
+            // Determine the length of the extracted substring
+            size_t extractedLength = extractedSubstring.length();
+
+            // Amount of signs to left can't be higher than extracted substring
+            singsToLeft = std::min(extractedLength, singsToLeft);
+
+            // Replace the substring with the appropriate number of '*' characters
+            textToAnonymize.replace(
+                startPosition + 1, extractedLength - singsToLeft, extractedLength - singsToLeft, '*');
```
Lefucjusz, line 214: I'd also define this `'*'` as a `constexpr` - it will be easier to read and to change the anonymization character in case it's needed in the future

module-cellular/modem/ATCommon.cpp
lukasmas, line 192: const std::string&

module-cellular/modem/ATCommon.cpp
```diff
@@ -156,3 +162,108 @@ Result Channel::checkResult(bsp::cellular::CellularResultCode cellularResult)
...
+            startPosition = endPosition + 1;
+        }
+        else {
+            break;
+        }
+    }
+    return textToAnonymize;
+}
+std::string Channel::anonymizeNumbers(std::string textToAnonymize, size_t singsToLeft)
```
lukasmas, line 223: const std::string&

module-cellular/modem/ATCommon.cpp
lukasmas, line 166: const std::string&

module-cellular/modem/ATCommon.cpp
```diff
@@ -156,3 +162,108 @@ Result Channel::checkResult(bsp::cellular::CellularResultCode cellularResult)
...
+std::string Channel::anonymizeIfNecessary(std::string text)
+{
+    std::string substringCPIN = "CPIN";
+    std::string substringCLCC = "CLCC";
+    std::string substringATD  = "ATD";
+    std::string substringCPBR = "CPBR";
+    std::string substringCSPN = "CSPN";
+
+    if (text.find(substringCPIN) != std::string::npos) {
+        text = anonymizeInQuotationMarks(text);
```
lukasmas, line 175: You are using else if
then you can just return when calling the function
if(){
return func():
}
else if()
return func1():
}

module-cellular/modem/ATCommon.cpp
```diff
@@ -156,3 +162,108 @@ Result Channel::checkResult(bsp::cellular::CellularResultCode cellularResult)
...
+                singsToLeft = std::min(extractedLength, singsToLeft);
+
+                // Replace the substring with the appropriate number of '*' characters
+                textToAnonymize.replace(
+                    startPosition, extractedLength - singsToLeft, extractedLength - singsToLeft, '*');
+
+                startPosition = endPosition + 1;
+            }
+            else {
+                LOG_ERROR("startPosition or/and endPosition not valid!");
```
lukasmas, line 260: Is this log relevant?

module-cellular/modem/ATCommon.cpp
```diff
@@ -28,6 +29,7 @@ const std::string Channel::CMS_ERROR  = "+CMS ERROR:";
 
 void Channel::cmdLog(std::string cmd, const Result &result, std::chrono::milliseconds timeout)
 {
+    LOG_WARN("--------- cmd:%s", cmd.c_str()); // for debug only
```
lukasmas, line 32: delete this

module-cellular/modem/ATCommon.cpp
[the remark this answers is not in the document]
lukasmas: Also if you created a new file for this function it can be extracted to anonymous namespace

module-cellular/modem/ATCommon.cpp
```diff
@@ -156,3 +162,108 @@ Result Channel::checkResult(bsp::cellular::CellularResultCode cellularResult)
...
+        text = anonymizeNumbers(text, 2);
+    }
+    else if (text.find(substringCPBR) != std::string::npos) {
+        text = anonymizeInQuotationMarks(text, 1);
+    }
+    else if (text.find(substringCSPN) != std::string::npos) {
+        text = anonymizeInQuotationMarks(text, 2);
+    }
+
+    return text;
```
lukasmas, line 190: I have also another proposition for this function, IMO it looks better than this else if block :) 

```std::string anonymizeIfNecessary(const std::string& text){
    std::map <std::string, std::function<std::string()>> anonymizeOption
    {
    {substringCPIN,[&text](){return anonymizeInQuotationMarks (text);}},
    {substringCLCC,[&text](){return anonymizeInQuotationMarks (text, 2);}},
    {substringATD,[&text](){return anonymizeNumbers (text, 2);}},
    {substringCPBR,[&text](){return anonymizeInQuotationMarks (text, 1);}},
    {substringCSPN,[&text](){return anonymizeInQuotationMarks (text, 2);}}
    };
    
    for(const auto& [key, anonymizeFunction] : anonymizeOption){
        if(text.find(key) != std::string::npos){
            return anonymizeFunction();
        }
    }
    return text;
    
}```

File: module-cellular/modem/ATCommon.cpp (hunk 1 above)
Comment (lukasmas) [reply module-cellular/modem/ATCommon.cpp]: Or we can just use LOG_SENSITIVE and adjust it.
AFAIK the LOG_SENSITIVE is doing nothing special, the flag is always ON and if logs exactly the way we pass in the argument.

File: module-utils/utility/Anonymize.hpp [hunk 14]
```diff
@@ -0,0 +1,122 @@
+// Copyright (c) 2017-2023, Mudita Sp. z.o.o. All rights reserved.
```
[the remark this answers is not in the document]
lukasmas: You can also consider moving it to module-utils/log

module-utils/utility/Anonymize.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+#include <string>
+#include <algorithm>
+#include <functional>
+#include <map>
+
+namespace utils::anonymize
+{
+    // Anonymized text will replace by this character
+    constexpr auto anonymizationCharacter = '*';
```
lukasmas, line 14: Move all constexpr variables which are used **only** in Anonymize to the anonymous namespace in .cpp
They don't need to be available everywhere in the code if they are only required in those functions

lukasmas (approved):

Maciej-Mudita (approved):

PR #4693, opened 2023-12-13
[BH-1836] Increase range of volume adjustment

lukasmas:
Extended volume scale from 10 to 15 point scale
Modified volume approximation function for better user experience

Make sure that this PR:

### DIFF
--- a/module-audio/Audio/Audio.cpp
+++ b/module-audio/Audio/Audio.cpp
@@ -34,15 +34,7 @@ namespace audio
 
     audio::RetCode Audio::SetOutputVolume(float vol)
     {
-        auto volToSet = vol;
-        if (vol > maxVolume) {
-            volToSet = maxVolume;
-        }
-        if (vol < minVolume) {
-            volToSet = minVolume;
-        }
-
-        return currentOperation->SetOutputVolume(volToSet);
+        return currentOperation->SetOutputVolume(static_cast<float>(vol));
     }
 
     audio::RetCode Audio::SetInputGain(Gain gain)
--- a/module-audio/board/rt1051/bellpx/BellPxAudioCodec.hpp
+++ b/module-audio/board/rt1051/bellpx/BellPxAudioCodec.hpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
+// Copyright (c) 2017-2023, Mudita Sp. z.o.o. All rights reserved.
 // For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 #pragma once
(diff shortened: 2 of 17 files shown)


module-audio/Audio/Audio.cpp
```diff
@@ -34,15 +34,7 @@ namespace audio
 
     audio::RetCode Audio::SetOutputVolume(Volume vol)
     {
-        auto volToSet = vol;
-        if (vol > maxVolume) {
-            volToSet = maxVolume;
-        }
-        if (vol < minVolume) {
-            volToSet = minVolume;
-        }
```
Maciej-Mudita, line 43: imo additional security won't be a problem - consider using std::clamp

Lefucjusz: This creates an issue after changing number of volume levels for Harmony - this code is placed in common part and it's not so simple to determine which maximum value should be used. We've checked that every call to this function is already guarded with `std::clamp`

module-audio/board/rt1051/bellpx/BellPxAudioCodec.hpp
```diff
@@ -29,6 +29,8 @@ namespace audio
 
     class BellPxAudioCodec : public SAIAudioDevice
     {
+        static constexpr auto maxVolume = 15.0f;
+        static constexpr auto minVolume = 0.0f;
```
Maciej-Mudita, line 33: is this called somewhere outside the BellPxAudioCodec.cpp module? If not, please move it to an anim space inside BellPxAudioCodec.cpp. This will speed up compilation if someone modifies these values.
By the way, have you considered reducing the volume constants? Currently, there are several of them scattered in different modules

module-bsp/board/rt1051/bellpx/bsp/audio/CodecAW8898.cpp
```diff
@@ -21,21 +21,22 @@ using namespace bsp::audio;
...
-    constexpr float transformVolumeLvl(float value)
+    /// Higher layers operate using 0-15 range. Here we are transforming it into more usable one.
+    /// Anything above 13.5 is going to distort the speaker and values below 5 are too quiet.
+
+    float transformVolumeLvl(float value)
     {
-        constexpr auto maxOutVolume = 9;
-        constexpr auto minOutVolume = 5;
+        constexpr auto maxOutVolume = 13.5f;
+        constexpr auto minOutVolume = 7.5f;
```
Maciej-Mudita, line 35: description does not match the set value

Maciej-Mudita (approved):

module-audio/board/rt1051/bellpx/BellPxAudioCodec.hpp
lukasmas: I've reduced some of them in harmony apps. All others are used in different parts of code; some are a Pure related.

lukasmas: max/min Volume are now only in two files. Both are common and Pure used them, so we need to leave the value 10 there.

Lefucjusz (approved):

PR #4713, opened 2024-01-15
[BH-1857] Fix improper PWM module clock frequency computation

Lefucjusz:
* Fix of the issue that source clock for PWM 
module was improperly assumed to be 
derived from AHB_CLK, while in reality it 
is derived from IPG_CLK, what resulted 
in module generating signal with 4 times
lower frequency than the configured one.
* Cleanups.

Make sure that this PR:


module-bsp/board/linux/torch/torch.cpp
```diff
@@ -1,65 +1,62 @@
...
+        return true;
+    }
 
-        ColourTemperature getColorTemp()
-        {
-            return currentColourTemp;
-        }
+    State getState()
+    {
+        return state_simulated;
```
dawidwojtas-mudita, line 45: What about camel case convention instead of snake here ;)

dawidwojtas-mudita (approved):

module-bsp/board/linux/torch/torch.cpp
Lefucjusz: Done :wink:

PR #4724, opened 2024-01-30
[BH-1863] Fix deleted file popup showing in Relaxation

Lefucjusz:
* Fix of the issue that 'File has been deleted' 
popup would show in Relaxation app at the 
end of playback if the playback was paused 
at least once, even though the file wasn't 
actually deleted.
* Added very basic audio decoder error 
handling and propagation mechanism.
* Minor refactor around several 
audio-related parts.

Make sure that this PR:


module-audio/Audio/decoder/Decoder.hpp
```diff
@@ -16,19 +16,19 @@ namespace audio
 {
     namespace channel
     {
-        constexpr inline auto monoSound   = 1;
-        constexpr inline auto stereoSound = 2;
+        inline constexpr auto monoSound   = 1;
+        inline constexpr auto stereoSound = 2;
```
dawidwojtas-mudita, line 20: Let's consider using `enum` here ;)

dawidwojtas-mudita (approved):

Maciej-Mudita (approved):
good job!

module-audio/Audio/decoder/Decoder.hpp
Lefucjusz: I thought about it, but these values are used as an absolute numbers in `pushAudioData`:

```
const auto readScale     = (channelMode == ChannelMode::ForceStereo) ? channel::stereoSound : channel::monoSound;
```
 and I would have to `static_cast` them anyway - IMO this will look more complicated.

dawidwojtas-mudita: Thanks for the explanation ;) It makes sense

PR #4729, opened 2024-02-05
[BH-1873] Update FreeRTOS ticks before re-enabling interrupts

Lefucjusz:
Fix of the issue that when leaving WFI mode
interrupts were being enabled before catching
up FreeRTOS ticks, what resulted in IRQ
handlers working on out-of-date system tick
counter.

Make sure that this PR:


MateuszSz-mudita (approved):

dawidwojtas-mudita (approved):

module-sys/SystemManager/PowerManager.cpp
```diff
@@ -243,18 +243,37 @@ namespace sys
...
+            /* Increase the frequency immediately after exiting WFI so that the xTaskCatchUpTicks
+             * procedure has time to execute and does not block the button press detection
+             * mechanism. A critical section is required so that any exit from any critical section
+             * inside any of the functions called here does not re-enable interrupts - unfortunately
+             * FreeRTOS critical section implementation for RT1051 completely ignores initial PRIMASK
+             * state and will re-enable interrupts even if they were disabled initially. We don't
+             * want to do that until the tick counter has been updated. */
+            cpp_freertos::CriticalSection::Enter();
+            {
+                SetCpuFrequency(bsp::CpuFrequencyMHz::Level_3);
```
Maciej-Mudita, line 267: why are we changing now to `CpuFrequencyMHz::Level3` instead of `CpuFrequencyMHz::Level4`?

Maciej-Mudita (approved):

module-sys/SystemManager/PowerManager.cpp
Lefucjusz: After leaving critical section power management algorithms would often increase CPU frequency to 528MHz if we changed the frequency to `CpuFrequencyMHz::Level4` here. I'm not sure why that happened, but lowering the level here helped without impacting initial responsiveness and now the frequency is increased to at most 132MHz

Maciej-Mudita: This may also affect power management because redrawing eInk at 132 MHz will take longer. I once did some research and it turned out that it is better to redraw at the highest frequency. We need to do the research again because power management is now significantly changed.

Maciej-Mudita: @Lefucjusz take a look https://appnroll.atlassian.net/browse/MOS-117

Lefucjusz: Before the fix I've implemented the frequency after leaving WFI would almost always switch to 132MHz, only sometimes to 528MHz, that's why I changed the CPU frequency level - to match the behavior to the current one. It's worth checking the entire power management mechanism and optimize it as it has been significantly modified indeed, but IMO it's better to do it under a separate task.

Maciej-Mudita: I agree

PR #4755, opened 2024-02-27
[BH-1887] Speed up update process

Lefucjusz:
* Added stream buffering for reading data 
from tar archive with update.
* Increased size of the buffer used for 
unpacking files from tar archive.
* Changed buffers data type from vectors 
to raw heap-allocated arrays.

Make sure that this PR:

### DIFF
--- a/module-utils/tar/tar.cpp
+++ b/module-utils/tar/tar.cpp
@@ -1,9 +1,8 @@
-// Copyright (c) 2017-2022, Mudita Sp. z.o.o. All rights reserved.
+// Copyright (c) 2017-2024, Mudita Sp. z.o.o. All rights reserved.
 // For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
 
 #include "tar/tar.hpp"
 
-#include <fstream>
 #include <vector>
 #include <iterator>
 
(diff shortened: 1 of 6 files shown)


module-utils/tar/tar.cpp
```diff
@@ -14,23 +13,25 @@ namespace
         return entry.name() != "." and entry.name() != ".." and entry.name() != "...";
     }
 
-    void write_to_file(const tar::entry &entry, const std::filesystem::path &path)
+    void writeToFile(const tar::entry &entry, const std::filesystem::path &path)
     {
-        constexpr std::size_t chunk_size = 1024 * 128;
-        std::ofstream out_file{path, std::ios::binary};
-        if (not out_file.is_open()) {
+        auto fd = std::fopen(path.c_str(), "wb");
```
dawidwojtas-mudita, line 18: Why do you use `std::fopen` instead of `std::ofstream`? The `ofstream` uses RAII so you don't need to take care of freeing resources if something goes wrong.

dawidwojtas-mudita (approved):
Before merge check the usage of the heap ;)

module-utils/tar/tar.cpp
Lefucjusz: I was experimenting with adding `setvbuf` to this file stream too (but it didn't speed up the process at all) and just left it that way :wink: 
I can change it back to `std::ofstream` if you prefer, but IMO this change shouldn't create any issues in case of failure - if the file descriptor is successfully created, it will be closed at the end of this method, if opening fails, we just get a `nullptr` and throw an exception, there's nothing to be closed. Additionally, after the change all file accesses during update process are performed using the same API :wink:

dawidwojtas-mudita: Ok, I get it. So leave `std::fopen` ;)

PR #4767, opened 2024-03-12
[BH-1913] Fix debug mode

dawidwojtas-mudita:
Add init eMMC pins to make sure that the pins are configured in debug mode.
If we run a debug mode the 1st bootloader is not run so the pins may have incorrect config. 
Added config also for Pure.

Make sure that this PR:


harmony_changelog.md
```diff
@@ -5,6 +5,7 @@
 ### Fixed
 * Fixed BIWIN eMMC memory errors
 * Fixed UI freeze after changing any volume intensively
+* Fixed debug mode
```
Lefucjusz, line 8: I'm not sure whether this changelog entry should be added - after all it's a non-released bug with no impact on users whatsoever.

Lefucjusz (approved):

harmony_changelog.md
dawidwojtas-mudita: Thanks, done ;)

PR #4773, opened 2024-03-18
[BH-1896] Add opening the vertical lists on the currently set sound

Maciej-Mudita:
After starting applications that use vertical lists of music, a list will open on the appropriate page with the previously selected song.

Make sure that this PR:

### DIFF
--- a/module-db/Tables/MultimediaFilesTable.cpp
+++ b/module-db/Tables/MultimediaFilesTable.cpp
@@ -386,4 +386,28 @@ namespace db::multimedia_files
 
         return (*retQuery)[0].getUInt32();
     }
+
+    auto MultimediaFilesTable::getOffsetOfSortedRecordByPath(const std::string &folderPath,
+                                                             const std::string &recordPath,
+                                                             SortingBy sorting) -> SortedRecord
+    {
+        const std::string query = "SELECT * FROM ("
+                                  "      SELECT"
+                                  "          ROW_NUMBER () OVER ("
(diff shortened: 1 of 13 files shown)


dawidwojtas-mudita (approved):

module-db/Tables/MultimediaFilesTable.cpp
```diff
@@ -386,4 +386,28 @@ namespace db::multimedia_files
...
+                                  "          files"
+                                  "      WHERE path LIKE '" +
+                                  folderPath + "%%'" +
+                                  "  )"
+                                  "  WHERE"
+                                  "      path='" +
+                                  recordPath + "'";
+
+        std::unique_ptr<QueryResult> retQuery = db->query(query.c_str());
+        return SortedRecord{.path = (*retQuery)[1].getString(), .offset = (*retQuery)[0].getUInt32()};
```
Lefucjusz, line 411: You can omit the type declaration here and just use `{}`, but if you prefer it with explicit type, feel free to resolve :wink:

```suggestion
        return {.path = (*retQuery)[1].getString(), .offset = (*retQuery)[0].getUInt32()};
```

Lefucjusz (approved):

module-db/Tables/MultimediaFilesTable.cpp
Maciej-Mudita: IMO it is slightly more readable with the exact type, so I'll leave it as it is :)

PR #4809, opened 2024-05-10
[BH-1944] Changing the rounding method for minutes in relaxation

Maciej-Mudita:
Only at startup, when the number of seconds is less than 30, the minutes are rounded down, otherwise they are rounded up

Make sure that this PR:

### DIFF
--- a/module-apps/apps-common/widgets/TimeMinuteSecondWidget.cpp
+++ b/module-apps/apps-common/widgets/TimeMinuteSecondWidget.cpp
@@ -17,6 +17,21 @@ namespace
         return digits;
     }
 
+    std::uint32_t getRoundedMinutes(std::uint32_t currentSeconds, std::uint32_t totalSeconds)
+    {
+        constexpr auto halfMinute{utils::time::secondsInMinute / 2};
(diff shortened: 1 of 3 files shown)


dawidwojtas-mudita (approved):

module-apps/apps-common/widgets/TimeMinuteSecondWidget.cpp
```diff
@@ -100,9 +115,12 @@ namespace gui
 
     void TimeMinuteSecondWidget::updateTime(std::uint32_t seconds)
     {
+        if (!totalSeconds) {
```
Lefucjusz, line 118: I'd explicitly compare it against zero so that it's clear that it's a numeric value, not a boolean type, pointer, `std::optional` etc.

Lefucjusz (approved):

module-apps/apps-common/widgets/TimeMinuteSecondWidget.cpp
Maciej-Mudita: done

PR #4815, opened 2024-05-24
[MOS-1068] Fix A2DP stream not restarting after song changes

Lefucjusz:
* Workaround for the issue that A2DP stream
would sometimes not restart when music
player changes song to the next one.
* Code cleanup in various places of
BTStack integration.

Make sure that this PR:


module-bsp/board/rt1051/bluetooth/BlueKitchen.cpp
```diff
@@ -17,62 +15,57 @@ using namespace bsp;
...
+    static std::unique_ptr<BlueKitchen> instance;
+    if (instance == nullptr) {
+        instance = std::make_unique<BlueKitchen>();
     }
-    return k;
+    return *instance;
 }
 
-BTdev::Error BlueKitchen::read(uint8_t *buf, size_t nbytes)
+BTDevice::Error BlueKitchen::read(uint8_t *buf, size_t nbytes)
```
dawidwojtas-mudita: To be consistent with previous changes. I'd add `std::` to `uint8_t` and `size_t`. I'd changed `nbytes` to `size`. However, feel free to resolve it if it's not important ;)

dawidwojtas-mudita (approved):

module-bsp/board/rt1051/bluetooth/BlueKitchen.cpp
Lefucjusz: You're right, I tried to change it everywhere but I must've missed that one, thanks!

PR #4835, opened 2024-06-27
[BH-2019] Fix crash after deleting WAV alarm sound

Lefucjusz:
Fix of the issue that the device would crash
when entering alarm sound selection list
after deleting selected file during alarm
ringing, but only when file's format was
WAV.

Make sure that this PR:


dawidwojtas-mudita (approved):
What about other formats? Do they also need this protection?

Lefucjusz:
Other formats already have this protection implemented, because of some unknown (at least for me) reason WAV was the only one that was missing it :wink:

Maciej-Mudita (approved):

PR #4839, opened 2024-07-04
[BH-2026] Add mechanism for getting previously OS version

dawidwojtas-mudita:
This task relates to the What’s new feature that needs to know which version of MuditaOS was installed before the update. Normally What’s new uses a database to get the previous  version, but after the first run this field will be empty. To bypass this issue the PureRecovery needs to save the previous version in recovery_status.json that is generated after the update process.

Make sure that this PR:

### DIFF
--- a/module-utils/utility/CMakeLists.txt
+++ b/module-utils/utility/CMakeLists.txt
@@ -4,6 +4,7 @@ target_sources(utility
     PRIVATE
         Utils.cpp
         Anonymize.cpp
+        Version.cpp
 
     PUBLIC
         integer.hpp
@@ -14,6 +15,7 @@ target_sources(utility
         Utils.hpp
         Units.hpp
         Anonymize.hpp
+        Version.hpp
 )
 
 target_include_directories(utility
@@ -28,6 +30,8 @@ target_link_libraries(utility
 
     PRIVATE
         hash-library::hash-library
+        json::json
+        purefs-paths
 )
 
 if (${ENABLE_TESTS})
--- a/module-utils/utility/Version.cpp
+++ b/module-utils/utility/Version.cpp
@@ -1,0 +1,23 @@
+// Copyright (c) 2017-2024, Mudita Sp. z.o.o. All rights reserved.
+// For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
+
+#include "Version.hpp"
+
+#include <module-utils/i18n/JSONLoader.hpp>
+#include <log/log.hpp>
+
+namespace utils
+{
+
+    std::optional<std::string> getPreviousVersionFromFile(const std::filesystem::path &path)
+    {
+        const auto &load = utils::JSONLoader(path);
+        if (not load.has_value()) {
+            LOG_ERROR("Cannot parse %s to json", path.c_str());
+            return {};
+        }
+        const auto &json = load.value();
+        return json["prevOsVersion"].string_value();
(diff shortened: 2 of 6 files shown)


module-utils/utility/Version.cpp
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) 2017-2024, Mudita Sp. z.o.o. All rights reserved.
+// For licensing, see https://github.com/mudita/MuditaOS/LICENSE.md
+
+#include "Version.hpp"
+
+#include <service-db/agents/settings/SystemSettings.hpp>
+#include <service-desktop/Constants.hpp>
```
Maciej-Mudita, line 7: is this necessary?

dawidwojtas-mudita: `#include <service-db/agents/settings/SystemSettings.hpp>` is needed by `settings::SystemProperties::osCurrentVersion`
but `#include <service-desktop/Constants.hpp>` isn't necessary here. Thanks ;)

module-utils/utility/CMakeLists.txt
```diff
@@ -28,6 +30,8 @@ target_link_libraries(utility
 
     PRIVATE
         hash-library::hash-library
+        purefs-paths
+        service-appmgr
```
Maciej-Mudita, line 34: where is used `appmgr` in `utils`?

Maciej-Mudita (approved):

module-utils/utility/CMakeLists.txt
dawidwojtas-mudita: I had to add `service-appmgr` to access `service-db/agents/settings/SystemSettings.hpp` because adding `serice-db` didn't resolve all dependencies.

Maciej-Mudita: maybe it would be better to pass a string with the saved version number from the settings rather than the settings themselves? Please consider this

module-utils/utility/Version.cpp
dawidwojtas-mudita: It's not important now, I removed those includes.

module-utils/utility/CMakeLists.txt
dawidwojtas-mudita: I redesigned the implementation. If the db returns the default version then the `getPreviousVersionFromFile` function is invoked to get the version from the file. So `service-db/agents/settings/SystemSettings.hpp` isn't longer needed.

module-utils/utility/Version.cpp
```diff
@@ -0,0 +1,23 @@
...
+
+    std::optional<std::string> getPreviousVersionFromFile(const std::filesystem::path &path)
+    {
+        const auto &load = utils::JSONLoader(path);
+        if (not load.has_value()) {
+            LOG_ERROR("Cannot parse %s to json", path.c_str());
+            return {};
+        }
+        const auto &json = load.value();
+        return json["preOsVersion"].string_value();
```
Lefucjusz, line 20: I'd name this key `prevOsVersion`, as with current form it sounds to me more like as "version that was there before OS" rather than "version of previous OS" :wink:

Lefucjusz (approved):

module-utils/utility/Version.cpp
dawidwojtas-mudita: I'm fine with your proposition ;) Done

PR #4840, opened 2024-07-09
[BH-1932] What's New application

Lefucjusz:
Added What's New application, which starts after
an update has been completed to display features
implemented between the previous OS version
and the one the user has just updated to.

Make sure that this PR:


Maciej-Mudita (approved):

patrycja-paczkowska (approved):

products/BellHybrid/apps/application-bell-whats-new/models/WhatsNewFeaturesModel.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+namespace app::whatsnew::models
+{
+    WhatsNewFeaturesModel::WhatsNewFeaturesModel(app::ApplicationCommon *app, settings::Settings *settings)
+        : settings{settings}
+    {
+        auto previousVersion =
+            settings->getValue(settings::SystemProperties::osCurrentVersion, settings::SettingsScope::Global);
+        if (previousVersion == utils::defaultPreviousVersion) {
+            previousVersion = utils::getPreviousVersionFromFile(purefs::dir::getTemporaryPath() /
```
exeosdev, line 60: If it's either a valid string or `utils::defaultPreviousVersion` why not return `utils::defaultPreviousVersion` from `getPreviousVersionFromFile` as a fallback value instead of an empty optional? The name suggests it should be done this way 🙂 Not a blocker, though!

exeosdev (approved):

products/BellHybrid/apps/application-bell-whats-new/models/WhatsNewFeaturesModel.cpp
Lefucjusz: Good point! IMO the implementation returning `std::optional` is more flexible, as it allows to distinguish between failure of retrieving version data from the file and retrieving the value that happens to be the same that the fallback one. In this case it doesn't matter at all, but if this function is used in some other context in the future this additional info might be useful...and if it's not, then at least such implementation doesn't introduce much overhead :wink:

PR #4871, opened 2024-09-04
[BH-2062] Fix alarm volume fade in duration

Maciej-Mudita:
The alarm volume should gradually increase by 0.1 every 300 ms, which gives a time of 45 seconds for a full 15-point scale

Make sure that this PR:

### DIFF
--- a/products/BellHybrid/services/audio/VolumeFade.cpp
+++ b/products/BellHybrid/services/audio/VolumeFade.cpp
@@ -10,24 +10,14 @@ namespace audio
     {
         using namespace std::chrono_literals;
         constexpr auto timerName{"volumeFadeTimer"};
-        constexpr auto fadeInterval{33ms};
+        constexpr auto minFadeInterval{33ms};
         constexpr auto fadeStep{0.1f};
-
-        std::chrono::milliseconds calculateTimerPeriod(std::chrono::seconds elapsed,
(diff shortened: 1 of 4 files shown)


dawidwojtas-mudita (approved):

products/BellHybrid/services/audio/VolumeFade.cpp
```diff
@@ -10,24 +10,39 @@ namespace audio
...
         using namespace std::chrono_literals;
         constexpr auto timerName{"volumeFadeTimer"};
-        constexpr auto fadeInterval{33ms};
+        constexpr auto minFadeInterval{33ms};
         constexpr auto fadeStep{0.1f};
 
-        std::chrono::milliseconds calculateTimerPeriod(std::chrono::seconds elapsed,
-                                                       std::chrono::seconds duration,
-                                                       float currentVolume)
+        std::chrono::milliseconds calculateFadeInterval(std::chrono::seconds maxDuration,
```
Lefucjusz, line 16: What about moving both of these functions inside `VolumeFade` class and making them private? As far as I see the arguments they take are mostly the fields of the class (apart from `timeElapsed`, which is calculated based on current time). That will greatly reduce number of arguments passed to the functions (in case of `calculateFadeInterval` to zero)

Lefucjusz (approved):

PR #4895, opened 2024-11-08
[BH-2082] Add support for Harmony 2 Pro

Lefucjusz:
* Added support for Harmony 2 Pro features
requiring different configuration.
* Added smoother brightness transitions
for Harmony 1/2 when brightness fades
in/out.

Make sure that this PR:


products/BellHybrid/alarms/src/actions/FrontlightAction.cpp
```diff
@@ -3,30 +3,33 @@
...
+        constexpr auto initialZeroBrightnessValue = 5.0f;
+#else
+        constexpr auto initialZeroBrightnessValue = 10.0f;
+#endif
 
-        void validateBrightness(std::string &brightness)
+        inline void validateBrightness(std::string &brightness)
         {
-            constexpr auto defaultBrightness = std::string_view{"50.0"};
+            constexpr auto defaultBrightness = std::string_view{"5"};
```
Maciej-Mudita, line 32: why are we changing the default value?

Lefucjusz: I changed the way brightness setting is stored in the database, previously PWM percentage was stored, now it's 1-10 scale used by the OS. 50% PWM is the same as 5 in this scale

Maciej-Mudita (approved):

PR #4903, opened 2025-01-09
[BH-2091] Remove timer from loop mode in Relaxation

Lefucjusz:
* Removed timer logic that would refresh the 
entire loop mode screen in Relaxation.
The contents of the screen doesn't change
unless there's a user activity or time
update, so the timer is not needed.
This should improve battery life in
Relaxation loop mode.
* Minor cleanups.

Make sure that this PR:

### DIFF
--- a/products/BellHybrid/apps/application-bell-relaxation/widgets/RelaxationPlayer.cpp
+++ b/products/BellHybrid/apps/application-bell-relaxation/widgets/RelaxationPlayer.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2024, Mudita Sp. z.o.o. All rights reserved.
+// Copyright (c) 2017-2025, Mudita Sp. z.o.o. All rights reserved.
 // For licensing, see https://github.com/mudita/MuditaOS/blob/master/LICENSE.md
 
 #include "RelaxationPlayer.hpp"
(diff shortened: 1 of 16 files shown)


Maciej-Mudita:
@Lefucjusz please add label

products/BellHybrid/apps/application-bell-relaxation/widgets/RelaxationPlayer.cpp
```diff
@@ -38,8 +38,7 @@ namespace app::relaxation
 
         auto fadeParams = audio::FadeParams{.mode = getFadeMode(), .playbackDuration = playbackDuration};
         audioModel.setPlaybackFinishedCb(std::move(onPlayerFinished));
-        audioModel.play(
-            filePath, Type::Multimedia, playbackMode, std::move(stateChangeCallback), std::move(fadeParams));
+        audioModel.play(filePath, Type::Multimedia, playbackMode, std::move(stateChangeCallback), fadeParams);
```
Maciej-Mudita, line 41: why was `std::move` removed for `fadeParams`?

Maciej-Mudita (approved):

products/BellHybrid/apps/application-bell-relaxation/widgets/RelaxationPlayer.cpp
Lefucjusz: This is a stack-allocated POD struct, so it cannot be moved, because there's no concept of ownership for such objects (the object is itself the data, it doesn't hold any pointers to heap-allocated data that could be moved). The only way to transfer such object is to just copy it, unless it has an explicitly defined move constructor...but then it's not an POD object anymore :wink:

It's very minor change, but I thought I'd make it, especially since `clang-tidy` complained about it too

PR #4913, opened 2025-02-11
[BH-2095] Add quote interval setting

Maciej-Mudita:
Modifying the time after which new quotes are displayed on the Harmony screen.

Make sure that this PR:

### DIFF
--- a/products/BellHybrid/services/db/agents/ShuffleQuoteModel.cpp
+++ b/products/BellHybrid/services/db/agents/ShuffleQuoteModel.cpp
@@ -7,6 +7,7 @@
 #include <log/log.hpp>
 #include <bsp/trng/trng.hpp>
 #include <service-db/agents/settings/SystemSettings.hpp>
+#include <time/time_constants.hpp>
 
 #include <memory>
 #include <ctime>
@@ -27,6 +28,12 @@ namespace
             return true;
         }
         return (std::abs(current - last) > dayInSec); // If the day is the same but different month/year
+    }
+
+    bool isTimeElapsed(std::time_t current, std::time_t last, std::uint32_t intervalInMinutes)
+    {
+        const auto elapsedMinutes = std::abs(current - last) / utils::time::secondsInMinute;
+        return elapsedMinutes >= intervalInMinutes;
     }
 } // namespace
 
@@ -102,6 +109,7 @@ namespace Quotes
         std::time_t lastTimestamp;
         const auto lastTimestampString =
             settings.getValue(settings::Quotes::randomQuoteIDUpdateTime, settings::SettingsScope::Global);
+        const auto interval = settings.getValue(settings::Quotes::selectedInterval, settings::SettingsScope::Global);
 
         try {
             lastTimestamp = std::stoll(lastTimestampString);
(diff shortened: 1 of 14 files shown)


products/BellHybrid/services/db/agents/ShuffleQuoteModel.cpp
```diff
@@ -28,6 +28,14 @@ namespace
         }
         return (std::abs(current - last) > dayInSec); // If the day is the same but different month/year
     }
+
+    bool isTimeElapsed(std::time_t current, std::time_t last, std::chrono::minutes minutes)
+    {
+        const auto currentTime = system_clock::from_time_t(current);
```
Lefucjusz, line 34: Do we need to convert timestamps to that intermediate type and `duration_cast` to minutes? If I get the idea of this method right IMO it will be easier to just do `std::abs(current - last) / utils::time::secondsInMinute`

products/BellHybrid/services/db/agents/ShuffleQuoteModel.cpp
```diff
@@ -112,12 +121,24 @@ namespace Quotes
...
 
         const auto currentTimestamp = std::time(nullptr);
-        if (hasCrossedMidnight(currentTimestamp, lastTimestamp)) {
-            settings.setValue(settings::Quotes::randomQuoteIDUpdateTime,
-                              utils::to_string(currentTimestamp),
-                              settings::SettingsScope::Global);
-            return true;
+
+        if (utils::is_number(interval)) {
+            if (isTimeElapsed(currentTimestamp, lastTimestamp, std::chrono::minutes(utils::toNumeric(interval)))) {
```
Lefucjusz, line 126: AFAIU if you apply my previous suggestion it should be able to directly pass `utils::toNumeric(interval)` here and avoid the division by `utils::time::secondsInMinute` (if `interval` is expressed in seconds)

products/BellHybrid/services/db/agents/ShuffleQuoteModel.hpp
```diff
@@ -12,6 +12,7 @@ namespace Quotes
 {
     inline constexpr auto customGroup     = "Custom";
     inline constexpr auto predefinedGroup = "Predefined";
+    inline constexpr auto atMidnight      = "at midnight";
```
Lefucjusz, line 15: I'd avoid using whitespace characters in such strings, what about `AtMidnight`?

products/BellHybrid/services/db/agents/ShuffleQuoteModel.cpp
Maciej-Mudita: You're right, it will be easier. Done.

products/BellHybrid/services/db/agents/ShuffleQuoteModel.cpp
Maciej-Mudita: The interval will be in minutes because the algorithm is written so that the quote duration is checked when the time is updated, i.e. every 1 minute.

products/BellHybrid/services/db/agents/ShuffleQuoteModel.hpp
Maciej-Mudita: done

products/BellHybrid/services/db/agents/ShuffleQuoteModel.cpp
Lefucjusz: Is it needed to use `std::chrono::minutes` then? It should be possible to pass the value of `utils::toNumeric(interval)` directly

products/BellHybrid/services/db/agents/ShuffleQuoteModel.hpp
Lefucjusz: Please update it also in `up.sql`

products/BellHybrid/services/db/agents/ShuffleQuoteModel.cpp
Maciej-Mudita: You are right. Done.

products/BellHybrid/services/db/agents/ShuffleQuoteModel.hpp
Maciej-Mudita: done

Lefucjusz (approved):

